PR #4, opened 2016-05-18
Renaming: Use JS for ethereumjs-abi.

chriseth:
@axic the "inconsistent" naming is actually on purpose: Modules that are supposed to be used together with `new` use uppercase letters, others use lowercase letters.


axic:
I thought that is the reason. Why not keep all of them uppercase then? Or use lint hints?

chriseth:
@axic sorry, I don't understand. I explained why not all of them are kept uppercase.

axic:
It is a lint requirement for those used as constructors to start with an uppercase letters. You can still choose to make the rest uppercase too. It is only a styling question.

chriseth:
@axic merging for now. We should add some automatic style checking for all of this code at some future point.

PR #6, opened 2016-05-18
Support block header parameters in VM mode (timestamp for now)

axic:
Fixes https://github.com/chriseth/browser-solidity/issues/90

*let's try this new travis build system*


axic:
This works now 👌

PR #10, opened 2016-05-18
Fix undo problems.

### LINKED ISSUE #5: ctrl+z
As per:
https://github.com/chriseth/browser-solidity/issues/162

chriseth:
Fixes https://github.com/chriseth/browser-solidity/issues/162
Fixes #5

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -184,6 +184,7 @@ var run = function() {
 
 	function newEditorSession(filekey) {
 		var s = new ace.EditSession(window.localStorage[filekey], "ace/mode/javascript")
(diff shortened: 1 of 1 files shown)


axic:
Did a local build and it works with cmd+z on Mac.

I'm not sure if this is a bug or unwanted: it reverts a full text block and not a single character.

e.g.

Initial text:
```
Hello

World
```

Adding two chunks of text:
```
Hello
chunk1
World
chunk2
```

Two subsequent cmd+z will first remove chunk2, then chunk1.

Let's leave this message here as a future reference if it comes up as an issue.

PR #12, opened 2016-05-19
Require ethereumjs-abi 0.6.1 to fix fixed-length array decoding

axic:
Fixes https://github.com/chriseth/browser-solidity/issues/156


chriseth:
Ahhh, these things are so much nicer now!

PR #13, opened 2016-05-19
Copy files over to another instance.

chriseth:
This PR provides a way to migrate content from one browser-solidity instance to another. It should work across origins and potentially even from `http://` to `file://`, but the latter might vary from browser to browser.

The way it works is that it opens an iframe with the destination and sends the files using `postMessage`.

The danger here lies in the fact that any other website could inject files in that way, origin checking is not done on purpose, to maintain maximal flexibility.

### DIFF
--- a/assets/css/browser-solidity.css
+++ b/assets/css/browser-solidity.css
@@ -236,10 +236,11 @@ body {
     padding-left: 6em;
 }
 
-#header #gist {
+#publishView button {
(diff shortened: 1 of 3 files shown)


axic:
Will this overwrite the destination's storage or append to it?

chriseth:
It is just like loading from a gist, so files with the same name but different contents will not be overwritten.

PR #19, opened 2016-05-20
Require ethereumjs-vm 1.4.0 to have 'consistent' gas costs

axic:
Don't merge yet - pending ethereumjs-vm release.

The VM is basically always starts in a state similar to *private blockchains*. Precompiled contracts are never transacted with, therefore the first interaction with them will trigger the `callNewAccount` gas fee of 25000.

This patch "activates" all precompiles so that fee won't be triggered by contracts.

Fixes https://github.com/chriseth/browser-solidity/issues/137.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "devDependencies": {
     "ethereumjs-block": "^1.2.2",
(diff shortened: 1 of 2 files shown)


axic:
It can be merged 👍

PR #20, opened 2016-05-20
Display transaction and inter-contract call costs separately

axic:
```
Result: "0x0000000000000000000000000000000000000000000000000000000000000001"
Transaction cost: 21486 gas. 
Inter-contract call cost: 214 gas.
Decoded: 
uint256: 1
```

Not sure how to display/explain this properly.  Perhaps the numbers need some tweaking too, right now:
- *inter-contract call cost* equals to the gas used during execution (perhaps this could be increased by the cost of the CALL/DELEGATECALL to show the final figure or maybe a message is sufficient?)
- *transaction cost* equals to the above cost + base fee + data fee

Fixes (partially) https://github.com/chriseth/browser-solidity/issues/137.

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -313,12 +313,18 @@ UniversalDApp.prototype.getCallButton = function(args) {
         return $('<div class="' + returnCls + '">').html('<strong>' + returnName + ':</strong> ' + JSON.stringify( result, null, 2 ) );
     };
 
-    var getGasUsedOutput = function (result) {
+    var getGasUsedOutput = function (result, vmResult) {
         var $gasUsed = $('<div class="gasUsed">');
         var caveat = lookupOnly ? '<em>(<span class="caveat" title="Cost only applies when called by a contract">caveat</span>)</em>' : '';
         if (result.gasUsed) {
             var gas = result.gasUsed.toString(10);
-            $gasUsed.html('<strong>Cost:</strong> ' + gas + ' gas. ' + caveat );
+            $gasUsed.html('<strong>Transaction cost:</strong> ' + gas + ' gas. ' + caveat );
+        }
+        if (vmResult.gasUsed) {
+            var $callGasUsed = $('<div class="gasUsed">');
+            var gas = vmResult.gasUsed.toString(10);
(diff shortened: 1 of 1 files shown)


src/universal-dapp.js
```diff
@@ -290,12 +290,18 @@ UniversalDApp.prototype.getCallButton = function(args) {
...
         var caveat = lookupOnly ? '<em>(<span class="caveat" title="Cost only applies when called by a contract">caveat</span>)</em>' : '';
         if (result.gasUsed) {
             var gas = result.gasUsed.toString(10)
-            $gasUsed.html('<strong>Cost:</strong> ' + gas + ' gas. ' + caveat )
+            $gasUsed.html('<strong>Transaction cost:</strong> ' + gas + ' gas. ' + caveat )
+        }
+        if (vmResult.gasUsed) {
+            var $callGasUsed = $('<div class="gasUsed">')
+            var gas = vmResult.gasUsed.toString(10)
+            $callGasUsed.append('<strong>Inter-contract call cost:</strong> ' + gas + ' gas.' )
```
chriseth: `Gas used by execution`?

axic: I have no good suggestion regarding this. It should be the cost of calling it from another contract (ie. `Execution cost` - without the transaction fee)

chriseth: The gas estimator in Solidity uses something like: `gas used: x + y (execution + transaction)`.

axic: One thing to keep in mind that we don't have that granular data when connecting to a real node, so we have to different kinds of output.

chriseth: But then even more so an argument for something like `gas used: x + y (execution + transaction)` in the vm mode and `gas used: x` in the web3 mode.

chriseth:
Rebase plase.

axic:
@chriseth rebased

axic:
With this version the output is like:

```
Result: "0x0000000000000000000000000000000000000000000000000000000000000001"
Transaction cost: 21486 gas. 
Execution cost: 214 gas.
Decoded: 
  uint256: 1
```

PR #23, opened 2016-05-21
Support account creation on both VM & RPC

axic:
Don't merge yet.

Moved from https://github.com/chriseth/browser-solidity/pull/154

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "semver": "^5.3.0",
     "solc": "https://github.com/ethereum/solc-js",
     "tape": "^4.5.1",
-    "web3": "^0.16.0",
+    "web3": "https://github.com/ethereum/web3.js#35bcace7173fda742ddb1bba84c8a70f0fa849e0",
     "webworkify": "^1.2.1",
     "yo-yo": "^1.2.2",
     "yo-yoify": "^3.3.0",
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -7,6 +7,7 @@ var ethJSABI = require('ethereumjs-abi');
 var EthJSBlock = require('ethereumjs-block');
 var BN = ethJSUtil.BN;
 var EventManager = require('./lib/eventManager');
+var crypto = require('crypto');
 
 /*
   trigger debugRequested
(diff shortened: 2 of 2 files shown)


chriseth:
Please rebase

axic:
I think I'll split out the changes for using the "personal API" of web3.  I think that also requires changes to runTx. I'd postpone this PR until that is sorted.

axic:
This is untested with web3 due to the personal account support - do not merge yet.

axic:
Implementing this PR in a clean way depends on the outcome of https://github.com/ethereum/EIPs/issues/113

chriseth:
ethereum/EIPs#113 was resolved

axic:
@chriseth how should this work? Go through `personal` if it is available in the node (recent geth?), othwerise fall back to `eth` and disable account creation? Is it worth it? :)

axic:
@chriseth also turns out we'll need to wait for web3.js 0.17.0 for the actual implementation

axic:
This needs some testing, but I'm more confident that the latest version is suitable for merging soon.

It doesn't actually turn on `personal` support, it has to be explicitly required via the configuration option.

package.json
```diff
@@ -30,7 +30,7 @@
     "semistandard": "^7.0.0",
     "solc": "https://github.com/ethereum/solc-js",
     "tape": "^4.5.1",
-    "web3": "^0.16.0",
+    "web3": "https://github.com/ethereum/web3.js#35bcace7173fda742ddb1bba84c8a70f0fa849e0",
```
chriseth: Is that really needed? How will we keep track?

axic: That hash is the current version of `master`. Asked the web3.js team when a 0.17.0 release is coming, but no response yet.

chriseth:
Please rebase

axic:
I wonder why this wasn't merged in July. After the rework which happened since it is quite a big change to rebase to.

axic:
@yann300 @chriseth can you test this and merge? In its current form it should have no adverse effect on anything.

chriseth:
Tested and seems to have no effect on anything. Did not enable the switch. Did I see correctly that this does not have any UI components yet?

axic:
@chriseth yes. There was no point complicating this PR any further. I think as a first step we could add a personal vs. non-personal mode switch to the UI.

PR #27, opened 2016-05-24
Improving imports

redsquirrel:
Currently the `importRegex` doesn't match unless there is exactly one space between `import` and `"file.sol"`. If someone put two+ spaces (or zero spaces) between `import` and `"file.sol"`, then the files have to be compiled twice.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -723,6 +723,9 @@ var run = function() {
 				if (getFiles().indexOf(fileKey(m)) !== -1) {
 					files[m] = window.localStorage[fileKey(m)];
 					reloop = true;
+				} else if (m.startsWith('./') && getFiles().indexOf(fileKey(m.slice(2))) !== -1) {
+					files[m] = window.localStorage[fileKey(m.slice(2))];
+					reloop = true;
 				} else if (m in cachedRemoteFiles) {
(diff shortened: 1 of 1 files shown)


src/app.js
```diff
@@ -708,7 +708,7 @@ var run = function() {
 			cb(files[fileNameFromKey(SOL_CACHE_FILE)]);
 			return;
 		}
-		var importRegex = /import\s*[\'\"]([^\'\"]+)[\'\"];/g;
+		var importRegex = /^\s*import\s*[\'\"]([^\'\"]+)[\'\"];/g;
```
redsquirrel: Currently, if someone comments out an import, browser-solidity will still parse them (and return an error message if they're missing). This update only matches on import statements that have nothing (other than spaces) in front of them.

axic: Haha! I had the exact same issue an hour ago 😃 

See #28. I like your change better as it is more flexible, will remove that part from mine.

redsquirrel:
Cherry-picked in @axic's https://github.com/ethereum/browser-solidity/pull/28/commits/a0f7885e224a902e8840b1a0efedb90f412de816 from #28.

PR #33, opened 2016-05-26
Setting web3 provider when web3 context is toggled

redsquirrel:
When a user chooses the "Web3 Provider" option, it doesn't actually set the provider unless the endpoint value is changed.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -243,22 +243,28 @@ var run = function() {
 	$vmToggle.on('change', executionContextChange );
 	$web3Toggle.on('change', executionContextChange );
 	$web3endpoint.on('change', function() {
-		var endpoint = $web3endpoint.val();
-		if (endpoint == 'ipc')
-			web3.setProvider(new web3.providers.IpcProvider());
-		else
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -243,22 +243,28 @@ var run = function() {
...
 	$web3Toggle.on('change', executionContextChange );
 	$web3endpoint.on('change', function() {
-		var endpoint = $web3endpoint.val();
-		if (endpoint == 'ipc')
-			web3.setProvider(new web3.providers.IpcProvider());
-		else
-			web3.setProvider(new web3.providers.HttpProvider(endpoint));
-		compile();
+		setProviderFromEndpoint();
+		if (executionContext == 'web3') compile();
```
chriseth: Why only in this case?

redsquirrel: I'm making two assumptions here:
* the only possible values for `executionContext` are `web3` and `vm`
* the value of `endpoint` is meaningless while in `vm` context

chriseth: Oh now I see. We really need to refactor that code...

redsquirrel: challenge accepted! 🚑

PR #34, opened 2016-05-26
Use better JS source formatting [WIP]

axic:


chriseth:
We decided on "standard" (i.e. no semicolons) for remix and also ethereumjs uses "standard". Would it be too much work to also change to standard here?

axic:
Eventually it'll end up with "standard", but there were less missing semicolons than semicolons in total. I'm going in a small step by step fashion (there are more commits locally) to be able to verify it properly. There's a lot of other fixes needed to be done, as a last step before indent I'll can all the semicolons.

chriseth:
Can't we just use an auto-formatter?

PR #36, opened 2016-05-27
Looking up libraryName instead of contractName

redsquirrel:
The code appears to be trying to detect a missing library, but it's actually looking up the name of the contract that's using the library.

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -499,7 +499,7 @@ UniversalDApp.prototype.linkBytecode = function(contractName, cb) {
     if (!m)
         return cb("Invalid bytecode format.");
     var libraryName = m[1];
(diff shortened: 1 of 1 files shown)


chriseth:
Why did that work before?

redsquirrel:
It didn't throw an immediate error because `contractName` was in scope: https://github.com/ethereum/browser-solidity/blob/master/src/universal-dapp.js#L494

chriseth:
Sure, but linking actually worked.

redsquirrel:
The only time it wouldn't work is if the contract's bytecode referred to a non-existent library. The only thing this PR fixed was that instead of getting a runtime error due to invalid bytecode, the user gets a link-time error of `Library <libraryName> not found.`

PR #39, opened 2016-05-27
Initial refactoring effort: extract high-level sections into modules

redsquirrel:
Based on [this comment](https://github.com/ethereum/browser-solidity/pull/33#discussion_r64923904) by @chriseth, I'm attempting to refactor app.js. This is an initial attempt, mainly to see if the approach makes sense. The approach is to extract methods and concepts into cohesive modules in order to make dependencies clearer, and to eventually pare down app.js to something that can more easily fit into someone's head.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -4,6 +4,10 @@ var web3 = require('./web3-adapter.js');
 var ace = require('brace');
 require('./mode-solidity.js');
 
+var queryParams = require('./app/query-params');
+var gistHandler = require('./app/gist-handler');
+var StorageHandler = require('./app/storage-handler');
+
 // The event listener needs to be registered as early as possible, because the
 // parent will send the message upon the "load" event.
 var filesToLoad = null;
@@ -20,55 +24,15 @@ var run = function() {
 
 	// ------------------ query params (hash) ----------------
 
-	function getQueryParams() {
-
-		var qs = window.location.hash.substr(1);
-
-		if (window.location.search.length > 0) {
-			// use legacy query params instead of hash
-			window.location.hash = window.location.search.substr(1);
-			window.location.search = "";
-		}
-
-		var params = {};
-		var parts = qs.split("&");
-		for (var x in parts) {
-			var keyValue = parts[x].split("=");
-			if (keyValue[0] !== "") params[keyValue[0]] = keyValue[1];
-		}
-		return params;
-	}
-
-	function updateQueryParams(params) {
-		var currentParams = getQueryParams();
-		var keys = Object.keys(params);
-		for (var x in keys) {
-			currentParams[keys[x]] = params[keys[x]];
-		}
-		var queryString = "#";
-		var updatedKeys = Object.keys(currentParams);
-		for( var y in updatedKeys) {
-			queryString += updatedKeys[y] + "=" + currentParams[updatedKeys[y]] + "&";
-		}
-		window.location.hash = queryString.slice(0, -1);
-	}
-
-
 	function syncQueryParams() {
-		$('#optimize').attr( 'checked', (getQueryParams().optimize == "true") );
+	  $('#optimize').attr( 'checked', (queryParams.get().optimize == "true") );
 	}
-
 
 	window.onhashchange = syncQueryParams;
 	syncQueryParams();
 
 	// ------------------ gist load ----------------
 
-	function getGistId(str) {
-		var idr = /[0-9A-Fa-f]{8,}/;
-		var match = idr.exec(str);
-		return match ? match[0] : null;
-	}
 	function loadFiles(files) {
 		for (var f in files) {
 			var key = fileKey(f);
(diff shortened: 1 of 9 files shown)


src/app.js
```diff
@@ -81,134 +33,59 @@ var run = function() {
 			}
 			window.localStorage[key] = content;
 		}
-		SOL_CACHE_FILE = fileKey(Object.keys(files)[0]);
+		editor.setCacheFile(Object.keys(files)[0]);
```
chriseth: That function should be named differently, but can / should be done in a subsequent refactoring step.

redsquirrel: Agreed, there's a ton more to clean up, rename, and clarify. For now, I tried to keep the names as similar as possible.

src/app.js
```diff
@@ -81,134 +33,58 @@ var run = function() {
...
-		if(loadingFromGist) return;
-		// Backwards-compatibility
-		while (window.localStorage[SOL_CACHE_UNTITLED + untitledCount])
-			untitledCount = (untitledCount - 0) + 1;
-		SOL_CACHE_FILE = SOL_CACHE_UNTITLED + untitledCount;
-		window.localStorage[SOL_CACHE_FILE] = window.localStorage['sol-cache'] || BALLOT_EXAMPLE;
-		window.localStorage.removeItem('sol-cache');
-	}
+	var loadingFromGist = gistHandler.handleLoad(function(gistId) {
+		$.ajax({
```
chriseth: You mixed tabs and spaces here, but we should probably just run a formatter over this...

src/app.js
```diff
@@ -81,134 +33,58 @@ var run = function() {
...
-	SOL_CACHE_FILE = getFiles()[0];
 
+	// ----------------- storage --------------------
 
-	var files = getFiles();
-	for (var x in files) {
-		sessions[files[x]] = newEditorSession(files[x]);
-	}
+	var storageHandler = new StorageHandler(updateFiles);
+	window.syncStorage = storageHandler.sync;
```
chriseth: I know it was like that before, but why is that needed?

src/app.js
```diff
@@ -373,75 +194,53 @@ var run = function() {
 	$filesEl.on('click', '.file .remove', function(ev) {
 		ev.preventDefault();
 		var name = $(this).parent().find('.name').text();
-		var index = getFiles().indexOf( fileKey(name) );
 
 		if (confirm("Are you sure you want to remove: " + name + " from local storage?")) {
-			window.localStorage.removeItem( fileKey( name ) );
-			SOL_CACHE_FILE = getFiles()[ Math.max(0, index - 1)];
+			window.localStorage.removeItem( utils.fileKey( name ) );
+			editor.setNextFile(utils.fileKey(name));
```
chriseth: Isn't this rather `setCacheFileByName`?

src/app.js
```diff
@@ -373,75 +194,53 @@ var run = function() {
...
 		ev.preventDefault();
-		SOL_CACHE_FILE = fileKey( $(this).find('.name').text() );
+		editor.setCacheFile($(this).find('.name').text());
 		updateFiles();
 		return false;
 	}
 
-	function fileTabFromKey(key) {
-		var name = fileNameFromKey(key);
+	function fileTabFromKey() {
```
chriseth: If you remove the parameter I think you should also rename it: `activeFileTab`?

src/app/compiler.js
```diff
@@ -0,0 +1,208 @@
...
+
+  var cachedRemoteFiles = {};
+  var worker = null;
+
+  var compileTimeout = null;
+
+  function onChange() {
+    var input = editor.getValue();
+    if (input === "") {
+      window.localStorage.setItem(editor.getRawCacheFile(), '');
```
chriseth: That should probably move to the editor.

src/universal-dapp.js
```diff
@@ -192,7 +192,6 @@ UniversalDApp.prototype.getInstanceInterface = function (contract, address, $tar
                 return;
 
             $event = $('<div class="event" />');
-
```
chriseth: Not sure if this commit was correctly applied...

src/app.js
redsquirrel: Do you want me to handle that now?

src/app.js
redsquirrel: I have no clue what `window.syncStorage` does. I'm just trying to preserve behavior since this is a refactoring.

src/app.js
redsquirrel: No, because it's using the name to find the *next* file. (Since the named file has been removed.) There are a flurry of subsequent minor refactorings to do once we get this major one out of the way.

src/universal-dapp.js
redsquirrel: Every line that @axic changed in app.js was extracted to the new modules. In the process, I made the same formatting updates. (Except for jsonInterface vs interface, which is about to be committed.)

redsquirrel:
⚠️  DO NOT MERGE YET. ⚠️ 

I just found a bug. 🐜

chriseth:
Looks good to me up to this point.

redsquirrel:
OK, false alarm on the bug. Ready for merge.

PR #43, opened 2016-05-31
Deploying and linking libraries in Web3 context

### LINKED ISSUE #37: Contract needs to be linked to a library, this is only supported in the JavaScript VM for now.
Context: https://github.com/ethereum/browser-solidity/blob/master/src/universal-dapp.js#L397

Is there any reason why this can't or shouldn't be supported when in __Web3 Provider__ mode? I'm working on library support and could investigate this.

redsquirrel:
This is an attempt to resolve #37.

__Note:__ Even though it deploys and links the library, the bytecode in the UI does not update. (It still has the unlinked version.)

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -418,18 +418,14 @@ UniversalDApp.prototype.getCallButton = function (args) {
         if (isConstructor) {
             if (args.bytecode.indexOf('_') >= 0) {
                  replaceOutput($result, $('<span>Deploying and linking required libraries...</span>'));
-                 if (self.options.vm) {
-                     self.linkBytecode(args.contractName, function (err, bytecode) {
-                         if (err) {
-                             replaceOutput($result, $('<span/>').text('Error deploying required libraries: ' + err));
-                         } else {
-                             args.bytecode = bytecode;
-                             handleCallButtonClick(ev, $result);
-                         }
-                     });
-                 } else {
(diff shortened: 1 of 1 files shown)


src/universal-dapp.js
```diff
@@ -562,24 +556,7 @@ UniversalDApp.prototype.runTx = function( data, args, cb) {
         }
     }
 
-    if (!this.vm) {
```
axic: Can you please have this part of the change as a separate commit? It doesn't seem to be related, but rather cosmetical.

redsquirrel: @axic Yes, good call. I just forced pushed a simpler commit.

axic:
I think #63 is the right way to handle `tryTillResponse()`. What do you think?

redsquirrel:
Re: #63, that seems like a great update for `tryTillFinished`. The purpose of this PR is to remedy #37.

axic:
Absolutely. You did change where `tryTrillFinished` is used, hence my comment. If it gets merged, you'll need to update this PR though.

redsquirrel:
The squirrel with no name is ready.

chriseth:
Did not work 100% for me, but probably still better than nothing. Also it would be good to get same feedback about how many libraries still need to be deployed.

axic:
@chriseth @redsquirrel I think merging this is breaking after web3 was moved to the context. `tryTillResponse` references an invalid `this` now?

redsquirrel:
@axic you're right, `this` should be `self`.

PR #48, opened 2016-06-02
Define undefined/redefined variables

axic:


axic:
Rebased against @redsquirrel's changes.

PR #49, opened 2016-06-02
Removing an unused method in UniversalDApp

redsquirrel:
Neither of these methods are used in this project. Improve readability by deleting them?

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -534,10 +534,6 @@ UniversalDApp.prototype.deployLibrary = function(contractName, cb) {
     }
 };
 
-UniversalDApp.prototype.clickNewContract = function ( self, $contract, contract ) {
-    $contract.append( self.getInstanceInterface(contract) );
-};
-
 UniversalDApp.prototype.clickContractAt = function ( self, $output, contract ) {
     var address = prompt( "What Address is this contract at in the Blockchain? ie: '0xdeadbeaf...'" );
     self.getInstanceInterface(contract, address, $output );


src/universal-dapp.js
```diff
@@ -53,30 +53,6 @@ UniversalDApp.prototype.getAccounts = function (cb) {
     }
 };
 
-UniversalDApp.prototype.getBalance = function (address, cb) {
```
axic: This was specifically added for a new feature - to display account balances after each transaction.

redsquirrel: Cool. Is anyone working on that feature?

axic: I won't have time. The account system on the UI side must be reworked. Currently we have a dropdown of accounts available and the one selected is used to make the next transaction. My idea was to somehow make that nicer and display a balance next to each address.

Either update after every transaction or have an *update balance* button.

axic: And #23 could be part of the *reworking the account system* task 😃

src/universal-dapp.js
```diff
@@ -534,10 +510,6 @@ UniversalDApp.prototype.deployLibrary = function(contractName, cb) {
     }
 };
 
-UniversalDApp.prototype.clickNewContract = function ( self, $contract, contract ) {
```
redsquirrel: @axic Is this also intended to be used in the future?

axic: No idea about that one. According to *blame* it is there since the beginning: https://github.com/ethereum/browser-solidity/commit/bf0acf274b4726902f8e47efc429bae56d78595d

redsquirrel:
I updated this commit to only remove `clickNewContract` which is dead with no plans for use.

PR #22, opened 2016-05-21
Use ethereumjs-abi for encoding requests [WI]

axic:
Moved from https://github.com/chriseth/browser-solidity/pull/163

Do not merge yet.

PR #50, opened 2016-06-02
Web3 cleanup

axic:
I'm sure that the first 3 commits are useful, but not sure if @redsquirrel plans to do something with web3 in the upper layers, hence the last commit might or might not be useful.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,5 +1,4 @@
 var $ = require('jquery');
-var web3 = require('./web3-adapter.js');
 
 var utils = require('./app/utils');
 var queryParams = require('./app/query-params');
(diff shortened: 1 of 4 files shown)


redsquirrel:
Don't worry about what I'm planning. I'm still figuring things out. This is a nice refactoring. 👍

chriseth:
Looks good!

axic:
@chriseth rebased

axic:
@chriseth I think I've managed to break it with this one. web3 is needed for ABI encoding.

kumavis:
wasnt abi broken out into a separate module somewhere?

axic:
It only uses ethereumjs-abi for decoding the results. For encoding there's a PR #22 to change from web3.js to ethereumjs-abi.

kumavis:
sweet

PR #51, opened 2016-06-02
Use formatting according to 'istandard'

axic:
We still need to apply indentation changes and fix some of the bugs found, such as `if (widthOfList() + getLeftPosi() > + widthOfVisible()) {`

cc @redsquirrel @chriseth

PR #54, opened 2016-06-03
Cleanup compiler (webworkify)

axic:


redsquirrel:
Great to see the continued simplification of app.js 😄

src/app/compiler.js
```diff
@@ -128,21 +128,23 @@ function Compiler(web3, editor, handleGithubCall, outputField, hidingRHP, update
   }
 
   this.loadVersion = function (usingWorker, version, setVersionText) {
+    var url = 'https://ethereum.github.io/solc-bin/bin/' + version;
```
axic: This could also be moved to the caller.

axic:
My end goal is to use `solc-js` instead of the duplicate (twice!) wrapper around soljson.

There's a lot of improvements needed to get there, but the above changes make it way easier.

axic:
@chriseth apply #51 before this as rebasing would be more of a pain otherwise

axic:
Rebased.

@chriseth and @redsquirrel please review these changes

redsquirrel:
Reviewed, and smoke-tested. 👍

PR #60, opened 2016-06-06
Use solc.js compiler [WIP]

axic:
This is a follow up of #54 and needs that PR to be merged first (has those commits here also, but will rebase).  It also depends on https://github.com/ethereum/solc-js/pull/21.

It is incomplete and work in progress. Looking for comments, @chriseth @redsquirrel

### DIFF
--- a/src/app/compiler-worker.js
+++ b/src/app/compiler-worker.js
@@ -1,4 +1,5 @@
-var version = function () { return '(loading)'; };
+var solc = require('solc/wrapper');
+
 var compileJSON = function () { return ''; };
 var missingInputs = [];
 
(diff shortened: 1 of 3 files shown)


src/app/compiler-worker.js
```diff
@@ -12,25 +13,22 @@ module.exports = function (self) {
...
-				{
-					compileJSONInternal = Module.cwrap("compileJSONCallback", "string", ["string", "number", "number"]);
-					var missingInputCallback = Module.Runtime.addFunction(function(path) {
-						missingInputs.push(Module.Pointer_stringify(path));
-					});
-					compileJSON = function(input, optimize) {
-						return compileJSONInternal(input, optimize, missingInputCallback);
-					};
+
+				var compiler = solc(Module);
```
redsquirrel: Where does `Module` come from?

axic: Loading an emscripten compiled binary in the global context will be assigned to the `Module` variable.

`importScripts(data.data);` above will load the url and put it into the global context

chriseth:
Tests fail?

axic:
@chriseth it needs a newer solc release of package.json needs to refer to the solc github repo

axic:
Would be nice to find a resolution to https://github.com/ethereum/solc-js/issues/24 before merging this.

axic:
> Would be nice to find a resolution to ethereum/solc-js#24 before merging this.

The reason for that is to avoid double JSON encoding/decoding in the case of the worker.

axic:
@chriseth apart from the previous possible improvement, this worked last time I've checked

PR #62, opened 2016-06-06
Supporting an injected web3

### LINKED ISSUE #59: Add "publish" button for publishing to the blockchain.
*When* browsers who have `web3` available visit your page, it would make sense to provide a `Publish` button that allows publishing the compiled bytecode to the blockchain.

If you need a web3 browser to practice with, I would welcome you to join the MetaMask private beta.

redsquirrel:
Although `web3-adapter.js` supports web3 injection (from Mist, etc), the application code does not. This is an attempt to resolve #59.


src/app/renderer.js
```diff
@@ -8,17 +8,22 @@ var utils = require('./utils');
 function Renderer (editor, compiler, updateFiles) {
 
   var detailsOpen = {};
-  var executionContext = 'vm';
+  var executionContext = web3.injected ? 'injected' : 'vm';
```
axic: I think this will fail in the udapp code. It expects `vm` or `web3`.

redsquirrel: It doesn't fail. From udapp's perspective, it's either using the vm object or a web3 object. If the `executationContext` isn't `vm`, then it uses the a web3 object. That object is either setup manually with a provider, or gets the provider injected in `web3-adapter.js`.

axic:
```
    if (options.mode === 'vm') {
      ...
    } else if (options.mode !== 'web3') {
        throw new Error('Either VM or Web3 mode must be selected');
    }
```

redsquirrel:
```
    var dapp = new UniversalDApp(udappContracts, {
      mode: executionContext === 'vm' ? 'vm' : 'web3',
```

axic: Cool, it is sorted then! :)

axic:
Wouldn't it make sense merging in `web3-adapter.js` into `renderer.js` at this stage? Might simplify it a bit.

redsquirrel:
I just gave that a try, but then backed it out for now. I'd prefer to `rm web3-adapter.js` in some future rework.

redsquirrel:
Do not merge this. After some additional testing, I've found a bug.

redsquirrel:
The problem is right [here](https://github.com/ethereum/browser-solidity/blob/master/src/web3-adapter.js#L6) in web3-adapter.js. When that code is executed, web3 is defined, and its value is the web3 radio button right [here](https://github.com/ethereum/browser-solidity/blob/master/index.html#L108). This is happening because this is apparently the way HTML works: http://www.2ality.com/2012/08/ids-are-global.html. Any element with an `id` attribute automatically gets exported to a global variable by the same name.

The result is that `web3.currentProvider` returns `undefined`. Since vm mode is the default and the provider is set when the web3 mode is selected, this bug goes unnoticed.

So I'm renaming the id of the radio button to `_web3` and leaving a note to explain it. 😭

axic:
@redsquirrel oh I had the same issue when trying to sort out the mess of/by #50. Should have mentioned somewhere.

redsquirrel:
OK, this is now ready for review/merge.

src/app/execution-context.js
```diff
@@ -0,0 +1,81 @@
...
+var injectedProvider;
+
+if (typeof web3 !== 'undefined') {
+  injectedProvider = web3.currentProvider;
+  web3 = new Web3(injectedProvider);
+} else {
+  web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:8545'));
+}
+
+function ExecutionContext (compiler) {
```
axic: I do not think this is the "execution context". I think this should stay in renderer.js (or something which handles the UI).  By execution context I've meant the core part of universal-dapp.js, which handles:
- account/balance mangement
- transaction processing
- and hides the differences between web3 and VM

PR #63, opened 2016-06-06
Ensure runTx always returns details about the transaction in both VM and Web3 mode

axic:
Basically this merges `tryTillResponse()` into `runTx()` as I think it belongs there.

More testing and fine-tuning might be required.

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -245,7 +245,7 @@ UniversalDApp.prototype.getInstanceInterface = function (contract, address, $tar
         }
       });
     } else {
-      var eventFilter = this.web3.eth.contract(abi).at(address).allEvents();
+      var eventFilter = self.web3.eth.contract(abi).at(address).allEvents();
       eventFilter.watch(parseLogs);
     }
     $instance.append($title);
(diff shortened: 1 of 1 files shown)


src/universal-dapp.js
```diff
@@ -481,31 +483,8 @@ UniversalDApp.prototype.getCallButton = function (args) {
...
-                    function testResult (err, address) {
-                        if (!err && !address) {
-                            setTimeout(function () { tryTillResponse(txhash, done); }, 500);
-                        } else {
-                            done(err, address);
-                        }
-                    }
-
-                }
-                tryTillResponse(result, function (err, result) {
```
redsquirrel: Why did you remove this call to `tryTillResponse`? It seems like that would change the behavior (in a bad way).

axic: `tryTillResponse` is merged into `runTx`

redsquirrel: Aha, nice. I haven't tested it, but that seems like the right approach.

axic:
Rebased (makes #43 much cleaner also)

chriseth:
Change looks good, is it tested (also taking the recent merges into account)?

axic:
Merge #79 first.

This was working yesterday, but wasn't tested with the #43 change.

axic:
@redsquirrel if you have a chance, please try this.

It worked for me for both regular transactions and library deployments. It also decodes the response in web3 mode when calling constants (those which are calls and not transactions).

redsquirrel:
@axic I tested this via web3 with a standalone contract, a standalone library, and a contract with a link to a library. It all deployed fine to morden. 👍

Issue #65, opened 2016-06-06
Settle on and enforce a coding style


axic:
To conform with `remix` we should be using `standard`.

Tasks:
- [x] fix bugs found during `standard`
- [x] apply indentation changes required by `standard`
- [ ] apply semicolon changes required by `standard`
- [x] include test for lint (`npm run lint`)
- [x] include lint test in travis and require it before a PR is merged

axic:
I have a patchset changing it to `standard` (removing semicolons).

We should merge it after #262, #269 and #273 is merged. cc @yann300 @chriseth

PR #68, opened 2016-06-06
Codestyle formatting (Part 1)

axic:
Don't merge it yet. It might have to change to `standard` depending on #65.

This one ignores `mode-solidity.js` because that is generated by ACE.

Remaining warnings:
```
  src/app.js:30:13: "otherKey" is defined but never used
  src/app.js:135:9: "iframe" is defined but never used
  src/app.js:251:12: "widthOfHidden" is defined but never used
  src/app.js:270:41: Unexpected space after unary operator "+".
  src/app.js:417:7: Module is a read-only native object.
  src/app.js:417:7: "Module" is read only.
  src/app/compiler.js:179:16: Expected a conditional expression and instead saw an assignment.
  src/app/compiler.js:197:20: Expected a conditional expression and instead saw an assignment.
  src/app/storage-handler.js:14:9: "dont" is defined but never used
  src/universal-dapp.js:486:9: Move function declaration to function body root.
  src/web3-adapter.js:9:3: web3 is a read-only native object.
  src/web3-adapter.js:9:3: "web3" is read only.
  src/web3-adapter.js:11:3: web3 is a read-only native object.
  src/web3-adapter.js:11:3: "web3" is read only.
  worker.js:8:18: Parsing error: Delete of an unqualified identifier in strict mode.
```

I would rather rebase this after merging some of the pending PRs which move files around, such as #54.


axic:
Removed the unstable commits so we can merge this faster. 

The whitespace change can be verified via `git show -w <hash>`.

PR #70, opened 2016-06-07
Safer handling of imports with leading ./

redsquirrel:
The code that allows us to prefix imported files with "./" breaks when you're importing a library. This is because it's being defined twice, which is allowed for contracts. This prevents the double definition.

### DIFF
--- a/src/app/compiler.js
+++ b/src/app/compiler.js
@@ -203,7 +203,11 @@ function Compiler (editor, handleGithubCall, outputField, hidingRHP, updateFiles
       for (var fileName in files) {
         var match;
         while ((match = importRegex.exec(files[fileName]))) {
-          importHints.push(match[1]);
+          var importFilePath = match[1];
+          if (importFilePath.startsWith('./')) {
+            importFilePath = importFilePath.slice(2);
+          }
+          importHints.push(importFilePath);
         }
       }
       while (importHints.length > 0) {


chriseth:
Hm, I don't know if this is the correct fix. This should probably be handled at the compiler level.
Doing `import "github.com/a/b/c.sol";` is very different from `import "./github.com/a/b/c.sol";`.

redsquirrel:
If the leading ./ isn't sliced off before 
```
         if (m in files) {
            continue;
          }
```
Then the file gets compiled twice. I'm not sure why, but compiler only complains if the file has a library in it: `pong.sol:3:1: Error: Library "pong" declared twice (will create ambiguities during linking).`

src/app/compiler.js
```diff
@@ -180,15 +180,15 @@ function Compiler (editor, handleGithubCall, outputField, hidingRHP, updateFiles
       }
       while (importHints.length > 0) {
         var m = importHints.pop();
+        if (m.startsWith('./')) {
```
chriseth: Ah right, now I see it. Please move that to after the regex match.
Actually we should not do regex searches if we have a compiler that supports the import callback.

src/app/compiler.js
```diff
@@ -203,7 +203,11 @@ function Compiler (editor, handleGithubCall, outputField, hidingRHP, updateFiles
       for (var fileName in files) {
         var match;
         while ((match = importRegex.exec(files[fileName]))) {
-          importHints.push(match[1]);
```
redsquirrel: @chriseth I moved the code to under the regex. This solves an immediate problem, so I suggest merging it before we look into how to leverage the compiler to handle the imports (instead of a nasty regex).

Issue #64, opened 2016-06-06
Group storage related methods


axic:
Currently the storage related methods are scattered through different files:
- [x] rename `storage-handler.js` to `storage.js`
- [x] move file loading into it
- [ ] move caching into it
- [x] create a global instance (needed for effective caching) and use it in all sources
- [x] alternatively create the storage instance on the top level and pass it down to every module

chriseth:
In addition to that, we should also encapsulate the file-handling and separate it from the generic "persistency" provider.

redsquirrel:
@axic Can you review #72 (just merged) and see where this issue stands?

axic:
@redsquirrel I see that caching is part of the `editor` instance? Is that correct?

redsquirrel:
@axic What kind of caching are you looking for?

chriseth:
@axic can you elaborate on the caching part here, please, so we can close this issue?

axic:
We have a couple of random things name *cached*:
- `getEditorSize` / `setEditorSize` in app.js + app/storage.js
- `utils.isCachedFile`

I think storage.js:
- shouldn't contain these `EditorSize` methods
- should use a proven localStorage abstraction (e.g https://www.npmjs.com/package/store)

chriseth:
Ah yes, the "cached" part of some variable names should go completely, there is nothing "cached" about it.

axic:
I think editor should have the following:
- createEmptyFile(name)
- getFileContent(name)
- setFileContent(name, content)
- hasFile(name)
- renameFile(oldName, newName)
- switchToFile(name)
- getFileList() -> array of file names

chriseth:
I agree with the methods, but why should they be part of the editor? I think all of them except `switchToFile` should be methods of the storage abstraction, where the editor registers for change events (`fileChanged`, `fileRenamed`, `fileCreated`, `fileDeleted`).

Furthermore, please already keep in mind that `name` will be a path in the future, but I think we can still keep it a simple string with `/` as directory separator.

chriseth:
If this is just meant as a first step into that direction, then "big thumbs up" :-)

axic:
I'm currently inclined to look at "editor" as the abstraction maintaining the files available, including the one currently being changed.

It may be better to actually have 3 different modules:
- "files"
- editor
- storage

Basically I would like to simplify how `app.js` handles files.

chriseth:
My suggestion would be:
[...]

PR #72, opened 2016-06-07
Pulling all storage-related methods into storage.js

redsquirrel:
Responding to #64

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -8,7 +8,7 @@ var queryParams = new QueryParams();
 var GistHandler = require('./app/gist-handler');
 var gistHandler = new GistHandler();
 
-var StorageHandler = require('./app/storage-handler');
+var Storage = require('./app/storage');
 var Editor = require('./app/editor');
 var Renderer = require('./app/renderer');
 var Compiler = require('./app/compiler');
@@ -25,16 +25,13 @@ window.addEventListener('message', function (ev) {
 }, false);
 
 var run = function () {
+  var storage = new Storage(updateFiles);
+
   function loadFiles (files) {
     for (var f in files) {
       var key = utils.fileKey(f);
       var content = files[f].content;
-      if (key in window.localStorage && window.localStorage[key] !== content) {
-        var count = '';
-        while ((key + count) in window.localStorage) count = count - 1;
-        window.localStorage[key + count] = window.localStorage[key];
-      }
-      window.localStorage[key] = content;
+      storage.loadFile(key, content);
     }
     editor.setCacheFile(utils.fileKey(Object.keys(files)[0]));
     updateFiles();
@@ -82,15 +79,14 @@ var run = function () {
     });
   });
 
-  // ----------------- storage --------------------
+  // ----------------- storage sync --------------------
 
-  var storageHandler = new StorageHandler(updateFiles);
-  window.syncStorage = storageHandler.sync;
-  storageHandler.sync();
+  window.syncStorage = storage.sync;
+  storage.sync();
 
   // ----------------- editor ----------------------
 
-  var editor = new Editor(loadingFromGist);
+  var editor = new Editor(loadingFromGist, storage);
 
   // ----------------- tabbed menu -------------------
 
@@ -166,7 +162,7 @@ var run = function () {
     var fileList = $('input.inputFile')[0].files;
     for (var i = 0; i < fileList.length; i++) {
       var name = fileList[i].name;
-      if (!window.localStorage[utils.fileKey(name)] || confirm('The file ' + name + ' already exists! Would you like to overwrite it?')) {
+      if (!storage.exists(utils.fileKey(name)) || confirm('The file ' + name + ' already exists! Would you like to overwrite it?')) {
         editor.uploadFile(fileList[i], function () {
           updateFiles();
         });
@@ -200,9 +196,7 @@ var run = function () {
       $fileNameInputEl.off('keyup');
 
       if (newName !== originalName && confirm('Are you sure you want to rename: ' + originalName + ' to ' + newName + '?')) {
-        var content = window.localStorage.getItem(utils.fileKey(originalName));
-        window.localStorage[utils.fileKey(newName)] = content;
-        window.localStorage.removeItem(utils.fileKey(originalName));
+        storage.rename(utils.fileKey(originalName), utils.fileKey(newName));
         editor.setCacheFile(utils.fileKey(newName));
       }
 
@@ -218,7 +212,7 @@ var run = function () {
     var name = $(this).parent().find('.name').text();
 
     if (confirm('Are you sure you want to remove: ' + name + ' from local storage?')) {
-      window.localStorage.removeItem(utils.fileKey(name));
+      storage.remove(utils.fileKey(name));
       editor.setNextFile(utils.fileKey(name));
       updateFiles();
     }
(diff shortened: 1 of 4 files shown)


src/app/storage.js
```diff
@@ -0,0 +1,97 @@
...
+
+    var obj = {};
+    var done = false;
+    var count = 0
+    var dont = 0;
+
+    function check (key) {
+      chrome.storage.sync.get(key, function (resp) {
+        console.log('comparing to cloud', key, resp);
+        if (typeof resp[key] !== 'undefined' && obj[key] !== resp[key] && confirm('Overwrite "' + fileNameFromKey(key) + '"? Click Ok to overwrite local file with file from cloud. Cancel will push your local file to the cloud.')) {
```
axic: The bug fixed by https://github.com/ethereum/browser-solidity/pull/68/commits/0b44735763adfbb0ea84a1c5642ccff6d5c68804 still applies here.

src/app/storage.js
```diff
@@ -0,0 +1,97 @@
...
+    if (this.exists(filename) && this.get(filename) !== content) {
+      var count = '';
+      var otherKey = filename + count;
+      while (this.exists(filename + count)) count = count - 1;
+      this.rename(filename, filename + count);
+    }
+    this.set(filename, content);
+  };
+
+  this.sync = function () {
```
axic: This is unrelated to this PR: is this `sync()` really about syncing with the "cloud"? Shouldn't it be named appropriately? Perhaps if its is Chrome only: `syncChromeCloud()`?

src/app/storage.js
```diff
@@ -0,0 +1,97 @@
...
+    var content = this.get(originalName);
+    this.set(newName, content);
+    this.remove(originalName);
+  };
+
+  this.remove = function (name) {
+    window.localStorage.removeItem(name);
+  };
+
+  this.setEditorSize = function (size) {
```
redsquirrel: This encapsulates the storage/retrieval of the editor size from localStorage. In the future, we could create wrapper methods in editor.js if that makes more sense.

src/app/storage.js
redsquirrel: I haven't dug into what this is actually doing. I've just been trying to preserve behavior. I agree that its intent should be clarified.

src/app/storage.js
redsquirrel: Good point. Fixing...

src/app/storage.js
axic: `storage.get()` and `storage.set()` are public so it could be safely called from within the editor?

What it is used for? I don't see any references - though github diffs lately are weird and don't show everything.

redsquirrel: I am encapsulating the `EDITOR_SIZE_CACHE_KEY` constant in storage.js, which is where it belongs since it's only used for localStorage.

`storage.setEditorSize` and `storage.getEditorSize` are both extracted (and now called) from app.js.

chriseth:
Please rebas.

redsquirrel:
Rebased

chriseth:
Please rebase.

redsquirrel:
I will rebase this once #77 is merged because I believe they conflict.

redsquirrel:
Rebased

src/app/storage.js
```diff
@@ -2,7 +2,56 @@
...
+
+  this.get = function (key) {
+    return window.localStorage.getItem(key);
+  };
+
+  this.keys = function () {
+    return Object.keys(window.localStorage);
+  };
+
+  this.loadFile = function (filename, content) {
```
chriseth: It seems that `loadfile` has to be called with an already "encoded" filename, so its parameter should rather be called `key` and actually the function name should not refer to the concept of a "file".

I think it would be better to not expose the plain `get` and `set` methods here and only give access to file names (where this class would handle the encoding via `utils.fileKey`). If you say that this would be a subsequent refactoring step, I'm fine with that.

src/app/storage.js
```diff
@@ -2,7 +2,56 @@
...
+function Storage (updateFiles) {
+  var EDITOR_SIZE_CACHE_KEY = 'editor-size-cache';
+
+  this.rename = function (originalName, newName) {
+    var content = this.get(originalName);
+    this.set(newName, content);
+    this.remove(originalName);
+  };
+
+  this.remove = function (name) {
```
chriseth: Why not use prototypes?

src/app.js
```diff
@@ -25,16 +25,13 @@ window.addEventListener('message', function (ev) {
 }, false);
 
 var run = function () {
+  var storage = new Storage(updateFiles);
+
   function loadFiles (files) {
     for (var f in files) {
       var key = utils.fileKey(f);
```
chriseth: No need for these two local variables.

src/app/compiler.js
```diff
@@ -210,10 +210,7 @@ function Compiler (editor, renderer, queryParams, handleGithubCall, outputField,
           continue;
         }
         if (editor.hasFile(m)) {
-          files[m] = window.localStorage[utils.fileKey(m)];
-          reloop = true;
-        } else if (m.startsWith('./') && editor.hasFile(m.slice(2))) {
```
chriseth: Good that this is not needed anymore - what was the exact reason?

src/app/editor.js
```diff
@@ -147,16 +151,15 @@ function Editor (loadingFromGist) {
 
   function setupStuff (files) {
     var untitledCount = '';
-    if (!files.length || window.localStorage['sol-cache']) {
+    if (files.length === 0) {
       if (loadingFromGist) return;
       // Backwards-compatibility
```
chriseth: Actually we might not need that anymore - nobody used the old version at the new location.

chriseth:
All comments are minor, can be merged as is, but please answer the questions.

src/app/storage.js
redsquirrel: Will be cleaned up in subsequent refactoring.

src/app/storage.js
redsquirrel: No particular reason. This approach works, and it's the approach I've used in all of my recent refactorings. If you'd like to rework those to use prototypes, create an issue an I'll investigate.

src/app.js
redsquirrel: Are you referring to something that changed in this Pull Request? Or are you referring to `f` and `key`?

src/app/compiler.js
redsquirrel: I'm lumping this into this Pull Request because I realized this code is now dead thanks to c2047b3b9482411446303bb8882d9f4d19421a29, which strips off leading `./`.

src/app/editor.js
redsquirrel: Right, so I'm getting rid of all of the `sol-cache` stuff, which I assume is from an early version of browser-solidity. Is there more that we can remove?

src/app.js
chriseth: Yes, `key` and `content`

src/app/editor.js
chriseth: I think that's the only one for now.

src/app.js
redsquirrel: Ah, I see. OK. I've committed a change that inlines those variables.

redsquirrel: Oops, you already merged. 😄  The "subsequent refactoring" has begun!

Issue #71, opened 2016-06-07
Delete Button Broken


Denton-L:
# Expected Behaviour
After pressing the delete button, the file disappears and the next file is shown, or a blank screen if there are no other files.

# Current Behaviour
The file's tab is removed but the content still remains on screen. I am unable to edit anything or switch tabs. Opening up the Developer Console shows that every mouse movement, click or keystroke generates a JavaScript error.

# Reproduction
1. Create a new file.
2. Enter some text.
3. Press the X on the file to delete it.
4. Everything is now broken. Open the Developer Console for more information.

# Other Information
@redsquirrel, I did a `git bisect` and it appears that 0a37e6a3fbcd8be1e09affa6b860bb6d03143941 was the commit that broke it.

redsquirrel:
Yes, same as #53. It's on my todo list.

Denton-L:
Alright, thanks!

Issue #53, opened 2016-06-03
Uncaught TypeError: Cannot read property 'on' of undefined


redsquirrel:
When you close a file tab, the above JavaScript error occurs. I'm seeing this on Mac OS X 10.10.5, with Chrome 50.0.2661.102 (64-bit).

Denton-L:
It appears that `editor.getSession()` and `editor.on()` are both not defined, even though it's been used it earlier code.

chriseth:
Is this fixed?

Denton-L:
Yes, should be fixed now.

PR #73, opened 2016-06-07
Fixing the Delete button bug

redsquirrel:
Fixing #53 and #71 by reverting 0a37e6a

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -32,7 +32,7 @@ var run = function () {
       }
       window.localStorage[key] = content;
     }
-    editor.setCacheFile(Object.keys(files)[0]);
+    editor.setCacheFile(utils.fileKey(Object.keys(files)[0]));
     updateFiles();
   }
 
@@ -198,7 +198,7 @@ var run = function () {
         var content = window.localStorage.getItem(utils.fileKey(originalName));
         window.localStorage[utils.fileKey(newName)] = content;
         window.localStorage.removeItem(utils.fileKey(originalName));
-        editor.setCacheFile(newName);
+        editor.setCacheFile(utils.fileKey(newName));
       }
 
       updateFiles();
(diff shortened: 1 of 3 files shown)


redsquirrel:
Thanks to @Denton-L for finding the guilty commit!

axic:
Not sure it is relevant, but I've found this typo: https://github.com/ethereum/browser-solidity/pull/68/commits/0b44735763adfbb0ea84a1c5642ccff6d5c68804

redsquirrel:
@axic Nice catch. But I don't think it relates to this issue.

Denton-L:
I've been digging into it and your commit seems to only expose the bug, rather than create it. For some reason, `editor.on` and `editor.getSession` are both undefined. Even in earlier "good" versions, resizing the window will cause a JavaScript error because it's attempting to make a call to something undefined.

I guess this can serve as a band-aid fix for now but at some point, we should figure out what's wrong.

redsquirrel:
The first priority is to fix the bug in the user experience. Next step will be to retry the refactoring and get to the bottom of why it failed.

chriseth:
This seems to be stalled on travise. Can you force-push to retrigger the build?

Denton-L:
If this gets merged in its current state, the tree is going to get dirtied with double commits.

Can you `git reset --hard 0a37e6a3fbcd8be1e09affa6b860bb6d03143941` and then force push to get rid of the double commits?

redsquirrel:
Rebased, with clean commit history

redsquirrel:
@chriseth This fixes a fairly urgent bug (report at #71). Merge soon?

Denton-L:
@redsquirrel, are you going to be resubmitting a fixed version of this commit in the future? I think I discovered the source of the bug.

redsquirrel:
@Denton-L Yes, I had planned to. Want to share the source of the bug? 😄

Denton-L:
Yeah, sorry it's written down somewhere on the whiteboard at the office. I'll send it to you when I get to work.

Denton-L:
In `editor.js`, `this.setNextFile` calls `this.setCacheFile`. Either `setCacheFile` has to lose its call to `utils.fileKey` or `setNextFile` has to wrap its call with `utils.fileNameFromKey`.

redsquirrel:
Got it. Thanks @Denton-L, I'll take a look at that.

PR #74, opened 2016-06-08
Instrumenting project for unit testing

redsquirrel:
Addressing #8.

### DIFF
--- a/src/app/compiler.js
+++ b/src/app/compiler.js
@@ -1,19 +1,13 @@
 var webworkify = require('webworkify');
-var QueryParams = require('./query-params');
 var utils = require('./utils');
-var Renderer = require('./renderer');
 
 var Base64 = require('js-base64').Base64;
 
-function Compiler (editor, handleGithubCall, outputField, hidingRHP, updateFiles) {
-  var renderer = new Renderer(editor, this, updateFiles);
-  var queryParams = new QueryParams();
-
+function Compiler (editor, renderer, queryParams, handleGithubCall, outputField, hidingRHP, updateFiles) {
   var compileJSON;
   var compilerAcceptsMultipleFiles;
 
   var previousInput = '';
-  var sourceAnnotations = [];
 
   var cachedRemoteFiles = {};
   var worker = null;
@@ -40,7 +34,6 @@ function Compiler (editor, handleGithubCall, outputField, hidingRHP, updateFiles
 
   var compile = function (missingInputs) {
     editor.clearAnnotations();
-    sourceAnnotations = [];
     outputField.empty();
     var input = editor.getValue();
     editor.setCacheFileContent(input);
@@ -59,10 +52,10 @@ function Compiler (editor, handleGithubCall, outputField, hidingRHP, updateFiles
   };
   this.compile = compile;
 
-  this.addAnnotation = function (annotation) {
-    sourceAnnotations[sourceAnnotations.length] = annotation;
-    editor.setAnnotations(sourceAnnotations);
-  };
+  function setCompileJSON (_compileJSON) {
+    compileJSON = _compileJSON;
+  }
+  this.setCompileJSON = setCompileJSON; // this is exposed for testing
 
   function onCompilerLoaded (setVersionText, version) {
     setVersionText(version);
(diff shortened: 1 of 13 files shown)


redsquirrel:
This is just a small commit to get feedback from @axic and @chriseth on whether it makes sense to continue this approach to unit testing the app modules.

chriseth:
Not sure what you mean here. Yes, we should have unit tests. Although I think that this unit test in particular is not a very good example. The actual name used in storage should be internal information.

redsquirrel:
I mean, are you ok with using mocha and the way that this is setup? (Agreed that the test is weak.)

Any ideas for how to hook it into travis?

redsquirrel:
Pulled out the utils-test example. Committed an example of "testing" compiler.js. It's really just a smoke test for now, and more importantly, a way to expose dependencies and help decoupling efforts.

axic:
I personally like mocha, but since `solc-js` uses tape, perhaps we should follow?

chriseth:
I used tape for solc-js because ethereumjs seems to use it and there is a general trend towards lightweight stuff in npm-world. We have a much stronge requirement for the testing framework to be able to work with karma. If that works equally well with tape and there are no other objections, I would propose to use tape.

chriseth:
As far as travis is concerned: Just copy the `.travis.yml` file from solc-js and it will work.

axic:
`ethereumjs` is undergoing a move to tape and I think one of them has karma with tape (others have karma with mocha)

redsquirrel:
@chriseth: 
> Just copy the .travis.yml file from solc-js and it will work.

I'm not sure what you mean. Currently, travis runs the build, right? I assume we want to continue running that build, and then also run the tests. I need some guidance on how to hook the tests into current travis process.

chriseth:
Oh I am sorry, I somehow looked at the wrong repository. `.travis.yml` currently has

    script: npm run lint && npm run build

and I think changing that to

    script: npm run lint && npm run test && npm run build

will do the trick.

redsquirrel:
Wow, trying to get these modules under test via node.js is proving difficult due to tight coupling to the browser. I'm going to spend some time seeing if I can decouple, but it might be more practical to start with browser-based tests.

redsquirrel:
@axic @chriseth Feedback requested. You can see the test output in the travis results. I tested query-params.js by allowing the tests to provide a fake window object.

chriseth:
Looks good, although I agree - it is probably better to start with a browser test first, just as a quick check if some basic functions work of if a tiny comma error broke everything.

redsquirrel:
For now, I think I'll keep going down this unit testing road, since it does help expose dependencies between modules. Let me know if you think I should revert and focus on browser-testing.

chriseth:
We should do both unit and integration tests, I think

src/app/compiler.js
```diff
@@ -1,18 +1,13 @@
...
-var queryParams = require('./query-params');
 var utils = require('./utils');
-var Renderer = require('./renderer');
 
 var Base64 = require('js-base64').Base64;
 
-function Compiler (editor, handleGithubCall, outputField, hidingRHP, updateFiles) {
-  var renderer = new Renderer(editor, this, updateFiles);
-
+function Compiler (editor, renderer, queryParams, handleGithubCall, outputField, hidingRHP, updateFiles) {
```
chriseth: Uff, these dependencies need to be slimmed down at some point providing just a set of callbacks I would say.

src/app/editor.js
```diff
@@ -100,9 +110,15 @@ function Editor (loadingFromGist) {
...
     editor.getSession().clearAnnotations();
   };
 
+  this.addAnnotation = function (annotation) {
+    sourceAnnotations[sourceAnnotations.length] = annotation;
+    this.setAnnotations(sourceAnnotations);
+  };
+
   this.setAnnotations = function (sourceAnnotations) {
     editor.getSession().setAnnotations(sourceAnnotations);
```
chriseth: You should also update the array here. But I think it would be better to go through a tiny interface for all of these functions.

redsquirrel:
@chriseth Understood re: two kinds of testing. For this PR, I'm focused on unit testing.

src/app/editor.js
redsquirrel: Sorry, can you clarify? Which array should be updated? Are you suggesting extracting annotations behind its own interface?

chriseth: Oh I'm sorry. Of course the array does not have to be updated here. But nevertheless, I think it is good practice to have only one entry point into updating the `sourceAnnotations` array which then also calls `editor.getSennio().setAnnotations`.

chriseth:
Please rebase.

redsquirrel:
@chriseth rebased

PR #82, opened 2016-06-08
Codestyle formatting (Part 2)

axic:
Follow up of #68 (and that must be applied first). Adds `npm run lint` test to packages.json and has the more delicate commits regarding globals (Module and web3).

Todo: include the test in Travis so that future PRs are validated.

We could switch over to `standard` from `semistandard` either in this or a subsequent PR to keep diffs digestable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "Minimalistic browser-based Solidity IDE",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "build": "mkdir -p build; browserify src/index.js -o build/app.js"
```
chriseth: `,`

chriseth:
Nice job!

Issue #83, opened 2016-06-09
File tab font, double clicks, upload tool tip


redsquirrel:
It seems like since the file upload icon was put in place, a couple of things have changed in the UI:
* The font in the file tabs has changed to something different (was this intentional?)
* If you double-click anywhere in the UI, the file upload dialog appears
* When you first focus your mouse anywhere in the UI, the "Upload" tool tip appears.

Denton-L:
I think we can close this now.

PR #85, opened 2016-06-09
Fix all problems with the upload button

### LINKED ISSUE #80: File upload is broken
File upload in firefox is broken. File selection works fine and also upload (file is visible after reload), but the error `editor is undefined` appears here:
```
 this.onChangeSetup = function (onChange) {
    editor.getSession().on('change', onChange);
    editor.on('changeSession', function () {
      editor.getSession().on('change', onChange);
      /// ^ error here
      onChange();
    });
  };
```

Denton-L:
This should fix #80.

### DIFF
--- a/index.html
+++ b/index.html
@@ -45,7 +45,7 @@
 
 		<div id="editor">
 			<span class="newFile" title="New File"><i class="fa fa-file-code-o"></i></span>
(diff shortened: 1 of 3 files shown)


Denton-L:
This is embarrassing... I fixed #83 and discovered that the problem was a missing `>` in the html...

Denton-L:
My fix broke multiple file upload. Working on that right now.

chriseth:
I think this is still better to merge than not.

PR #86, opened 2016-06-09
Fixed multiple file upload

Denton-L:
Very sorry for the multiple badly formed pull requests...

### DIFF
--- a/src/app/editor.js
+++ b/src/app/editor.js
@@ -18,12 +18,12 @@ function Editor (loadingFromGist) {
 
   this.uploadFile = function (file, callback) {
     var fileReader = new FileReader();
-
     var cacheName = utils.fileKey(file.name);
+
     fileReader.onload = function (e) {
+      window.localStorage[cacheName] = e.target.result;
+      sessions[cacheName] = null;
       SOL_CACHE_FILE = cacheName;
-      window.localStorage[SOL_CACHE_FILE] = e.target.result;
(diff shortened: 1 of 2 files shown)


src/app/editor.js
```diff
@@ -19,8 +19,9 @@ function Editor (loadingFromGist) {
   this.uploadFile = function (file, callback) {
     var fileReader = new FileReader();
 
-    SOL_CACHE_FILE = utils.fileKey(file.name);
+    var cacheName = utils.fileKey(file.name);
     fileReader.onload = function (e) {
+      SOL_CACHE_FILE = cacheName;
```
chriseth: Is the browser stalled while the file loads? Is this safe if the user switches the current file in the meantime?

Denton-L: The browser shouldn't be stalled while the file loads because the `FileReader` is asynchronous. The other thing shouldn't be a problem but I'll push up a change to be sure.

Denton-L:
@chriseth, the most recent commit should be safer in case the user switches the file while another is uploading. In reality, though, this shouldn't be a problem because the files upload too quickly for anything to be able to happen.

chriseth:
I would like to merge #72 first, this has to be changed accordingly.

Denton-L:
Alright, I can rebase on top of #72.

chriseth:
Please rebase this one or rebase the other one properly.

Denton-L:
Alright, will rebase the other one.

PR #94, opened 2016-06-13
Initial browser testing

redsquirrel:
Addressing #8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 language: node_js
 node_js:
-- stable
-script: npm run lint && npm run test && npm run build
+  - stable
+before_script:
+  - npm run serve &
+script:
+  - npm run lint && npm run test && npm run build && nightwatch --env chrome
+  - pkill node
 addons:
   sauce_connect:
     username: "chriseth"
@@ -11,7 +15,7 @@ deploy:
   script: ci/deploy_from_travis.sh
   skip_cleanup: true
   on:
-        branch: master
+    branch: master
 env:
   global:
   - ENCRYPTION_LABEL="b5c2730599da"
@@ -19,5 +23,3 @@ env:
   - COMMIT_AUTHOR="Travis CI"
   - PUSH_REPO="git@github.com:ethereum/browser-solidity.git"
   - FILES_TO_PACKAGE="assets background.js build icon.png index.html manifest.json README.md"
-
-
--- a/test-browser/new-file-test.js
+++ b/test-browser/new-file-test.js
@@ -1,0 +1,10 @@
+module.exports = {
+  'New file test': function (browser) {
+    browser
+      .url('http://127.0.0.1:8080')
+      .waitForElementVisible('.newFile', 5000)
+      .click('.newFile')
+      .assert.containsText('.active', 'Untitled')
+      .end();
+  }
+};
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .vscode
 build
 node_modules
+reports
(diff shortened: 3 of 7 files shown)


.travis.yml
```diff
@@ -1,7 +1,12 @@
 language: node_js
 node_js:
-- stable
-script: npm run lint && npm run build
+  - stable
+before_script:
+  - npm run serve &
```
chriseth: Wouldn't it be better to run that as part of the script?

.travis.yml
```diff
@@ -1,7 +1,12 @@
...
+script:
+  - npm run lint && npm run build
+  - nightwatch --env chrome
```
chriseth: As far as I know, if you use this list syntax, it will run the next item even if the previous one failed. -> We should not run the tests if build failed.

.travis.yml
```diff
@@ -11,7 +16,7 @@ deploy:
   script: ci/deploy_from_travis.sh
   skip_cleanup: true
   on:
-        branch: master
+  branch: master
```
chriseth: Is that correct?

.travis.yml
redsquirrel: I'm following this tutorial to set it up: https://medium.com/@mikaelberg/zero-to-hero-with-end-to-end-tests-using-nightwatch-saucelabs-and-travis-e932c8deb695

.travis.yml
redsquirrel: OK, fixing.

redsquirrel:
@chriseth I believe this PR is configured and coded correctly, but the results of the Selenium tests have been inconsistent. Their elapsed times are significantly variable, and there are occasional false negatives. I assume this is due to instability in the connection between Travis & Sauce.

I get consistent/fast results when running the tests on my laptop with a local Selenium Server. I get consistent/slow results when running the tests on my laptop via the Sauce connection. I get inconsistent results when the tests are run via Travis. You can see this here: https://travis-ci.org/ethereum/browser-solidity/pull_requests

axic:
Looks cool! Could you think about a way to make it more concise (i.e. not to have every single test in a separate file)? Not sure how to group tests though.

test-browser/new-file-test.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  'New file test': function (browser) {
+    browser
+      .url('http://127.0.0.1:8080')
+      .waitForElementVisible('.newFile', 10000)
```
axic: Are you sure 10000 is enough given the speed of Travis?

redsquirrel:
@axic You can learn more about writing tests at: http://nightwatchjs.org/guide#usage. It appears that each file represents a browser session. You can break up the tests in the file, but they really represent "steps" in a sequence. Can you do me a favor and try running the tests locally?

test-browser/new-file-test.js
redsquirrel: I'm not sure. I'm still figuring this out.

redsquirrel:
OMG I think I figured out how to get it working consistently on Travis. I started using [waitForConditionTimeout](https://github.com/ethereum/browser-solidity/pull/94/files#diff-38662eda33df859414dcfaa49bac04b2R21), set to 60 seconds.

chriseth:
@redsquirrel please reabes

redsquirrel:
It's rebased. BUT, the browser tests are inexplicably failing again. 😞  Do not merge. 😞

redsquirrel:
@chriseth The tests passed. I suggest merging this. If the tests are flaky in master, then remove the call to nightwatch from travis and we can at least run them locally.

PR #97, opened 2016-06-13
Issue 92 documentation improvements

Victory:
I had a typo in the rebuild command. This Fixes the typo. Also mentions a version of Ubuntu where the build works. Older versions you can still get problems with memdown.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Run:
     nvm --version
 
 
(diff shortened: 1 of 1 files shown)


Victory:
I have also cross referenced the Ubuntu version with this SE page:

http://ethereum.stackexchange.com/questions/2673/testrpc-2-0-0-cannot-find-module-memdown

The question actually mentions an issue with 14.04 LTS but i was able to get it to run. (Xbuntu 14.04.4 LTS to be exact).

chriseth:
Thanks!

PR #102, opened 2016-06-15
Update browser-solidity.css

iFA88:
On Firefox the element with "dragbar" ID covers 90% of the element with "input" ID scroll bar. So it is really hard to drag the scroll bar.
I recommend to change in the #dragbar CSS element the Width from 1em to 0.5em.

### DIFF
--- a/assets/css/browser-solidity.css
+++ b/assets/css/browser-solidity.css
@@ -389,7 +389,7 @@ body {
 #dragbar {
     background-color: transparent;
     position: absolute;
-    width: 1em;
+    width: 0.5em;
     right: -3px;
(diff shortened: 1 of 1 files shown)


axic:
@iFA88 can you change the commit message to be more meaningful? i.e. put the real reason on the first line

iFA88:
@axic I have modified my comment. Its okay now? Just because when anyone check my file update they know immediately whats wrong, i should not need a word for that. :)

axic:
@iFA88 are you sure? The commit seems unaffected.

Also please rebase as this is against an older version of the tree which has failing tests.

PR #105, opened 2016-06-15
Fixing browser testing

redsquirrel:
Do not merge. I'm using this PR to investigate browser testing issues.

### DIFF
--- a/test-browser/new-file-test.js
+++ b/test-browser/new-file-test.js
@@ -2,7 +2,7 @@ module.exports = {
   'New file test': function (browser) {
     browser
       .url('http://127.0.0.1:8080')
-      .waitForElementVisible('.newFile', 10000)
+      .waitForElementVisible('.newFile', 60000)
       .click('.newFile')
       .assert.containsText('.active', 'Untitled')
       .end();
--- a/test-browser/smoke-test.js
+++ b/test-browser/smoke-test.js
@@ -2,7 +2,7 @@ module.exports = {
   'Smoke test': function (browser) {
     browser
       .url('http://127.0.0.1:8080')
-      .waitForElementVisible('#righthand-panel', 10000)
+      .waitForElementVisible('#righthand-panel', 60000)
       .assert.containsText('#righthand-panel', 'Solidity version')
       .end();
   }


redsquirrel:
I'm in touch with the TravisCI support team. Trying to resolve this. In the meantime, I suggest merging #104 in order for the build process to work.

redsquirrel:
Here is the full email from Travis support:

```
Hey Dave -

Thanks for your reply! Great questions. 

Restarting builds essentially send them back to the queue to run again, usually when they faild 
for some external reason that does not require code changes to get tests passing. There's a
little refresh circlular arrow on the Current build page on Travis for your repo - if you click
that, the build will restart (there's a similar question on Stack Overflow here with a nice
screenshot). Alternatively, you can use our CLI tool from the travis ge m to do the same thing.
Does it seem like one of those two options would work? 

As to access to debugging tools - of course! You can trigger a debug job as follows: 

1. Get the API token using the CLI and send a POST request to /job/:job_id/debug replacing the
TOKEN and JOB_ID values below:

#! /usr/bin/env bash

curl -s -X POST \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -H "Travis-API-Version: 3" \
  -H "Authorization: token <TOKEN>" \
  https://api.travis-ci.com/job/<JOB_ID>/debug

2. Head back to the web UI and in the log of your job. you should see the following lines to 
connect to the VM:

Setting up debug tools.
Preparing debug sessions.
Use the following SSH command to access the interactive debugging environment:
ssh ukjiuCEkxBBnRAe32Y8xCH0zj@ny2.tmate.io

3. Connect from your computer using SSH into the interactive session, and once you're
done, just type exit and your build will terminate. 

Please note that when the debug build is initiated like this, the job will skip the remaining
phases after debug. Also, if you're all set with your log after running the debug build,
please consider removing it. 

I hope this helps - and let us know if you give this option a try or if there's anything else
we can do to help :) 

Have a great day!
Anna
```

In order to troubleshoot this myself, I'm going to need write access to Github's browser-solidity project. If someone with write access wants to dig into this, go for it. 😄  We can also leave the browser testing as-is, and have it be a manual step (see README for how to run it locally) in the PR merge process rather than as part of the automated travis build.

chriseth:
Thanks, gave write access to you.

Issue #106, opened 2016-06-17
Provide warnings when vulnerable code is detected


redsquirrel:
This project seems well positioned to help new Solidity developers avoid [known vulnerabilities](http://hackingdistributed.com/2016/06/16/scanning-live-ethereum-contracts-for-bugs/). It seems like one thing that browser-solidity could do is to detect calls to `send` and provide a warning.

Are there other vulnerabilities we could point out?

chriseth:
Good idea!

flyswatter:
We just wrote a bytecode analyzer for this at MetaMask earlier this week!

Feel free to use our code:
https://twitter.com/metamask_io/status/743644432000417793

axic:
Given #107 was closed with

> Similar warnings have been added directly to the compiler, so I'm closing this PR.

Should this issue be considered closed?

PR #107, opened 2016-06-18
Introducing vulnerability warnings

redsquirrel:
By adding more vulnerability checks to `src/app/sol-warnings.js`, we can help educate Solidity developers on possible vulnerabilities in their contract code. Attempting to address #106.


kumavis:
@redsquirrel im doing lots of evm byte code analysis these days, can help there

redsquirrel:
@kumavis Have a look at ded722b6047df9ad01d973231e36a00cd2258a9c. This gives you access to AST, bytecode, opcodes, etc.

kumavis:
nice!

src/app/sol-warnings.js
```diff
@@ -0,0 +1,34 @@
...
+// compiledResult: an object with different keys based on success/failure of compilation.
+//   success keys: 'contracts' and 'sources' (contains AST)
+//   error key: 'error' (contains compile error string)
+
+function find (source, compiledResult) {
+  var messages = [];
+
+  if (source) {
+    if (source.match(/\.\s*send\s*\(/)) {
+      messages.push('Caution: You are using the send method.\nPlease read <a target="_blank" href="http://hackingdistributed.com/2016/06/16/scanning-live-ethereum-contracts-for-bugs/">this</a> before deploying your code.');
```
chriseth: I think it would be better to link to the documentation, also the same warning could be issued for any function call.

Also what about adding a general warning "Interacting with unknown contracts (including sending ether) can cause arbitrary side-effects including recursive calls into the current contract" before the call returns."?

redsquirrel: I agree a link to documentation would be better. But first we need better documentation. Until then, perhaps a link to this blog post is the best option? And that link should be provided when we detect calls to `send`, `call`, or any external function. Is that right?

Re: the general warning. Are you suggesting providing that regardless of what code is live in the editor?

chriseth: General warning: Yes - if we add such a warning, it should be always present. We cannot detect any external function call and we certainly do not want this warning for every single external function call.

redsquirrel: > We cannot detect any external function call

Just curious. Can't we detect this from the AST?

chriseth: Ok, you are right. What we probably cannot detect is whether the contract is "known" or not (i.e. generic).

redsquirrel:
Switched from using Regexps on the string source to searching the AST.

src/app/sol-warnings.js
```diff
@@ -0,0 +1,54 @@
+module.exports = { find: find };
+
+// source: just a big string of the current tab's source code.
+//
+// compiledResult: an object with different keys based on success/failure of compilation.
+//   success keys: 'contracts' (contains opcodes, bytecode) and 'sources' (contains AST)
+//   error key: 'error' (contains compile error string)
+
+function find (source, compiledResult) {
```
chriseth: `analyze` would be more appropriate, I think.

src/app/sol-warnings.js
```diff
@@ -0,0 +1,54 @@
+module.exports = { find: find };
```
chriseth: The file itself could be called `staticAnalysis`

src/app/sol-warnings.js
```diff
@@ -0,0 +1,54 @@
...
+    if (compiledResult.contracts) {
+      Object.keys(compiledResult.contracts).forEach(function (contractName) {
+        // @TODO: use compiledResult.contracts[contractName].opcodes, interface, etc
+      });
+    }
+    if (compiledResult.sources) {
+      Object.keys(compiledResult.sources).forEach(function (tabName) {
+        var ast = compiledResult.sources[tabName].AST;
+        if (searchAST(ast, {member_name: 'send'})) {
+          messages.push('Caution: You are using the send method.\nPlease read <a target="_blank" href="http://hackingdistributed.com/2016/06/16/scanning-live-ethereum-contracts-for-bugs/">this</a> before deploying your code.');
```
chriseth: I would rather warn on `call` or `call.value` because that is much more dangerous.

chriseth: Also, don't we have the source mapping as part of the ast? The message should be displayed at the correct place.

redsquirrel: I'm not sure, I will look into it.

redsquirrel: re: warn on `call` ... the idea here is to establish a place to warn on as many things as we want to warn on, along with providing links to learn more about the warning. happy to hear more suggestions on what we should warn on and where we could send people for further reading.

src/app/static-analysis.js
```diff
@@ -0,0 +1,54 @@
...
+    if (compiledResult.contracts) {
+      Object.keys(compiledResult.contracts).forEach(function (contractName) {
+        // @TODO: use compiledResult.contracts[contractName].opcodes, interface, etc
+      });
+    }
+    if (compiledResult.sources) {
+      Object.keys(compiledResult.sources).forEach(function (tabName) {
+        var ast = compiledResult.sources[tabName].AST;
+        if (searchAST(ast, {member_name: 'send'})) {
+          messages.push('Caution: You are using the send method.\nPlease read <a target="_blank" href="http://hackingdistributed.com/2016/06/16/scanning-live-ethereum-contracts-for-bugs/">this</a> before deploying your code.');
```
chriseth: Unchecked send is already reported by the compiler.

Furthermore, a core principle of the compiler is that any error and warning can be silenced by the user in a certain way (either by correcting the error or adding a certain pattern that says "yes, I know what I am doing").
Adding a warning for every single send will just cause the users to ignore all warnings, even the serious ones.

I would suggest creating a new "pane" where the user can see results of static analysis, with checkboxes that allow certain analyses to be performed and an explanation what the analysis is for.

redsquirrel: A new "pane" is a good idea. I'm going to close this for now.

redsquirrel:
Similar warnings have been added directly to the compiler, so I'm closing this PR.

src/app/static-analysis.js
axic: That would be a great idea! I think you could utilise https://github.com/ConsenSys/solidity-parser.

PR #112, opened 2016-06-23
Update gas limit to 4700000 in Universal Dapp

Denton-L:


redsquirrel:
Is 4,700,000 the standard amount?

Denton-L:
Gas limit is currently [4,712,388](https://etherscan.io/block/1760349). Just picking 4,700,000 to be in line with #91.

redsquirrel:
Thanks.

axic:
@Denton-L the gas limit you have set is not `4700000`, but `4700000000`.

It is not a change as simple as that.  The web3 option will run `estimateGas` and always execute the transaction with the needed gas limit.  The vm option doesn't have this, but an arbitrarily high limit.

I have some changes which make this configurable via the UI. I think that should be the proper way to resolve it.

redsquirrel:
@axic Do your changes allow a person to specify/override the gas sent in a transaction? /cc @simondlr

Denton-L:
Sorry, @axic... My bad...

axic:
@redsquirrel @Denton-L fixed it in #125 :)

We could go with 4,7M as a default, but that is the block gas limit, which depends on the miners.

PR #114, opened 2016-06-24
Fix #113

### LINKED ISSUE #113: Minor Thing: Double 0x
Upon creation of a new contract using the "At Address" option, the contract's address appears as "0x0xdeadbeef..."

Denton-L:
This should fix #113.


axic:
I think the check could be simply removed as both the vm and web3 will return 0x prefixed addresses.

Denton-L:
It's possible for a user to enter an address without the leading '0x', though. In that case, it'll be displayed without the leading '0x' as well.

thebalaa:
@Denton-L this broke things pretty bad, see #118

Denton-L:
Thanks, I'll fix it.

PR #119, opened 2016-07-02
Change value of address

### LINKED ISSUE #118: AssertionError: The field to must have byte length of 20
My previously working contracts no longer work in browser-solidity ( I cannot call any functions without getting the error shown below.

Even when calling 'fallback` on an empty contract. I did try with previous versions of solidity (by selecting from the dropdown) to no avail.

screen shot 2016-07-01 at 12 59 10 pm

```
contract Test {
}
```

Denton-L:
This should fix #118 because I broke it in #114.

@chriseth, this is a relatively urgent bug fix so if you could merge this PR soon, thanks!


thebalaa:
@chriseth 1+ for getting this merged and deployed asap

PR #120, opened 2016-07-08
sauce tests

yann300:


.travis.yml
```diff
@@ -1,17 +1,9 @@
...
-#  - npm run serve &
 script:
   - npm run lint && npm run test && npm run build
-# && nightwatch --env remote
-#  - pkill node
-#addons:
-#  sauce_connect:
-#    username: "chriseth"
-#    access_key: "b781828a-9e9c-43d8-89d4-2fbb879595ca"
+  - bash ci/browser_tests.sh
```
chriseth: Do you need `bash`? If the file is executable it should just work.

ci/browser_tests.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
chriseth: Please use `#!/usr/bin/env bash`

ci/browser_tests.sh
```diff
@@ -0,0 +1,43 @@
...
+
+
+SAUCECONNECT_URL="http://saucelabs.com/downloads/sc-4.3.16-linux.tar.gz"
```
chriseth: Please use `https`

ci/browser_tests.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
```
chriseth: Please add `set -e` (will make the script fail if a single command fails).

ci/browser_tests.sh
```diff
@@ -0,0 +1,43 @@
...
+done
+
+function updateTestExitCode() {
+  if [ $? -eq 1 ]
+  then
+    TEST_EXITCODE=1
+  fi
+}
+
+npm run browser-test-remote-firefox
```
chriseth: If you add `set -e` above, this of course does not work anymore :-)

But it can be replaced by: `npm run browser-test-remote-firefox || TEST_EXITCODE=1`

chriseth:
Amazing :-)

PR #121, opened 2016-07-08
Cache build

Denton-L:
This will ensure that Travis caches the node_modules folder between builds. This will ensure faster build times as Travis won't have to do a complete reinstall when calling `npm install` each build.


redsquirrel:
Thanks for fixing those whitespace issues. ❤️

PR #123, opened 2016-07-12
Self contained Travis build - work without Internet connection

axic:
Don't rely on the availability of https://ethereum.github.io/solc-bin/bin/list.js. First step towards #2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "browser-test-remote-ie": "nightwatch --config nightwatch.js --env ie",
     "browser-test-remote-chrome": "nightwatch --config nightwatch.js --env chrome",
     "browser-test-remote-safari": "nightwatch --config nightwatch.js --env safari",
-    "build": "mkdir -p build; browserify src/index.js -g yo-yoify -o build/app.js",
+    "build": "mkdir -p build; rm soljson.js; wget https://ethereum.github.io/solc-bin/soljson.js; browserify src/index.js -g yo-yoify -o build/app.js",
     "lint": "semistandard",
     "serve": "http-server ."
   },
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "browser-test-remote-ie": "nightwatch --config nightwatch.js --env ie",
     "browser-test-remote-chrome": "nightwatch --config nightwatch.js --env chrome",
     "browser-test-remote-safari": "nightwatch --config nightwatch.js --env safari",
-    "build": "mkdir -p build; rm soljson.js; wget https://ethereum.github.io/solc-bin/soljson.js; browserify src/index.js -g yo-yoify -o build/app.js",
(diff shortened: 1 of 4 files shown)


axic:
Testing needed! Any takers @redsquirrel @chriseth @Denton-L? :smiley:

axic:
Do we actually need to include the ZIP file? Github has an option to download a branch as a ZIP file.

axic:
Another possible way is to download soljson.js during `npm run build`, which seem to make more sense. Added it as a commit.

src/app.js
```diff
@@ -415,6 +400,34 @@ var run = function () {
 
   $('.asmOutput button').click(function () { $(this).parent().find('pre').toggle(); });
 
+  // ----------------- version selector-------------
+
+  // var soljsonSources is provided by bin/list.js
+
+  var versionList = true;
+  try {
+    soljsonSources;
```
chriseth: `window.soljsonSource === undefined`? (or document, whatever the global context is)

package.json
```diff
@@ -9,7 +9,7 @@
     "browser-test-remote-ie": "nightwatch --config nightwatch.js --env ie",
     "browser-test-remote-chrome": "nightwatch --config nightwatch.js --env chrome",
     "browser-test-remote-safari": "nightwatch --config nightwatch.js --env safari",
-    "build": "mkdir -p build; browserify src/index.js -o build/app.js",
+    "build": "mkdir -p build; rm soljson.js; wget https://ethereum.github.io/solc-bin/soljson.js; browserify src/index.js -o build/app.js",
```
chriseth: `rm -f`?

src/app.js
axic:
```
missingVar === undefined
Uncaught ReferenceError: missingVar is not defined(…)(anonymous function) @ VM2543:1
```

vs.

```
window.missingVar === undefined
true
```

Weird. Did the first and had that issue.

ci/deploy_from_travis.sh
```diff
@@ -10,6 +10,9 @@ git checkout --orphan gh-pages
 git rm --cached -r .
 echo "# Automatic build" > README.md
 echo "Built website from {$SHA}. See https://github.com/ethereum/browser-solidity/ for details." >> README.md
+echo "To use an offline copy, download browser-solidity-$SHA.zip." >> README.md
+# ZIP the whole directory
+zip -r browser-solidity-$SHA.zip .
```
chriseth: We should exclude the .git directory

package.json
chriseth: Is that a good thing to do in the build step? This way it always accesses the internet.

axic: True, maybe it makes more sense to have a separate call for it

chriseth:
I just merged the change that uses the solc npm package. How did you plan that this one here works with the other one?

axic:
@chriseth I don't think it is affected. Let me rebase it first.

PR #124, opened 2016-07-12
UDapp: don't supply a hardcoded limit to estimageGas, but judge based on the response

axic:
This way the actual required gas is displayed and once we have a way to change the gas limit on the UI, the user could change it to the required one.


axic:
And #125 should allow changing the gas limits in the UI.

PR #126, opened 2016-07-12
Use solc.linkBytecode for linking

axic:


src/universal-dapp.js
```diff
@@ -556,15 +557,7 @@ UniversalDApp.prototype.linkBytecode = function (contractName, cb) {
...
-    var libLabel = '__' + libraryName + Array(39 - libraryName.length).join('_');
-    var hexAddress = address.toString('hex');
-    if (hexAddress.slice(0, 2) === '0x') {
-      hexAddress = hexAddress.slice(2);
-    }
-    hexAddress = Array(40 - hexAddress.length + 1).join('0') + hexAddress;
-    while (bytecode.indexOf(libLabel) >= 0) {
-      bytecode = bytecode.replace(libLabel, hexAddress);
-    }
+    bytecode = solc.linkBytecode(bytecode, { [libraryName]: ethJSUtil.addHexPrefix(address.toString('hex')) });
```
chriseth: Is it really `[libraryName]`?

axic: `libraryName` is a literal, while `[libraryName]` will use the referenced value as key

axic:
I'm not sure what is broken with tests, but I get seemingly unrelated issues:
> Testing if element <.active> contains text: "Untitled".  - expected "Untitled" but got: ""

chriseth:
These test errors mean that either the application did not start up at all in the browser or some js did not initialize correctly.

chriseth:
Please rebase

chriseth:
Please rebase.

axic:
This actually won't work in this form, because `linkBytecode` is not a static method of the solc wrapper - it needs a soljson loaded. We need to pass down the loaded compiler to the udapp in that case.

axic:
Depends on https://github.com/ethereum/solc-js/pull/150

axic:
@yann300 updated this, it should work, but haven't tested.

axic:
@yann300 rebased!

yann300:
moved to remix repo

PR #131, opened 2016-07-15
remix integration

yann300:


src/app/debugger.js
```diff
@@ -0,0 +1,23 @@
+var Remix = require('ethereum-remix')
+var $ = require('jquery')
+
+function Debugger (executionContext, _id) {
+  this.el = document.querySelector(_id)
+  this.debugger = new Remix(executionContext.web3())
+  this.el.appendChild(this.debugger.render())
+  
+  this.web3 = executionContext.web3()
```
axic: Could this somehow be integrated with the VM-mode?

yann300: yes that's the next step

src/app/debugger.js
```diff
@@ -0,0 +1,21 @@
+var $ = require('jquery');
+
+function Debugger (executionContext, _id) {
+  this.el = document.querySelector(_id);
+  this.debugger = new window.remix.Debugger(executionContext.web3());
```
chriseth: Do we really need to fetch the `remix` object from `window`?

yann300: I think this is the easier solution.
here is how remix is loaded initially
https://github.com/ethereum/remix/blob/master/src/index.js

using something like require('ethereum-remix') // with ethereum-remix refering to the (remix) app.js was giving error while compiling browser solidity (i think because app.js is already a browserified file).

using something like require('ethereum-remix') // with ethereum-remix refering directly to ./ethdebugger.js
is not possible because in that case string templates are not yo-yoified (and string template are not supported by a few browsers)

@axic any thoughts about that? is there a better way?

chriseth:
Test is failing.

package.json
```diff
@@ -55,5 +57,8 @@
       "nightwatch.js",
       "ci/sauceDisconnect.js"
     ]
+  },
+  "dependencies": {
+    "ethereum-remix": "0.0.2-alpha.0.0.1"
```
chriseth: This is also only a dev dependency.

src/app.js
```diff
@@ -89,21 +90,23 @@ var run = function () {
   var editor = new Editor(loadingFromGist, storage);
 
   // ----------------- tabbed menu -------------------
-
   $('#options li').click(function (ev) {
     var $el = $(this);
-    var match = /[a-z]+View/.exec($el.get(0).className);
+    document.querySelector('ul#options').selectTab($el);
+  });
+  document.querySelector('ul#options').selectTab = function (el) {
```
chriseth: Please don't add functions to dom elements.

src/app/debugger.js
```diff
@@ -0,0 +1,20 @@
+var Remix = require('ethereum-remix');
```
chriseth: We could make that lowercase.

src/universal-dapp.js
```diff
@@ -352,6 +352,16 @@ UniversalDApp.prototype.getCallButton = function (args) {
     return $('<div class="' + returnCls + '">').html('<strong>' + returnName + ':</strong> ' + JSON.stringify(result, null, 2));
   };
 
+  var getDebugTransaction = function (result) {
+    var $debugTx = $('<div class="debugTx">');
+    var $button = $('<button class="debug" >Debug Transaction</button>');
```
chriseth: Surplus space in `"debug" >`

src/universal-dapp.js
```diff
@@ -514,6 +524,7 @@ UniversalDApp.prototype.getCallButton = function (args) {
       } else {
         clearOutput($result);
         $result.append(getReturnOutput(result)).append(getGasUsedOutput(result));
+        $result.append(getDebugTransaction(result));
```
chriseth: This should be present in other cases, too - we want to debug constant functions.

chriseth:
I think we can merge this one as a first demo, but the user interface can certainly be improved. Also I think that this is not too useful without the ability to debug vm transactions.

chriseth:
We should not display the "connection" setting as part of the remix ui, it should just use the one supplied by browser-solidity.

PR #132, opened 2016-07-15
Include a gas limit for web3 constant calls

axic:
Missed this in #124

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -645,6 +645,7 @@ UniversalDApp.prototype.runTx = function (data, args, cb) {
       value: value
     };
     if (constant && !isConstructor) {
+      tx.gas = gasLimit;
       self.web3.eth.call(tx, cb);
     } else {
       self.web3.eth.estimateGas(tx, function (err, resp) {


redsquirrel:
Just a general question: why would you need to specify (or spend) gas on an `eth_call`?

axic:
@redsquirrel albeit there is no gas being spent on `eth_call`, it can still be used to set a limit on how much time a call take, i.e. to avoid a deadlock on the node, when running an infinite loop in a contract.

redsquirrel:
@axic Really? If you wouldn't mind explaining, how does that work? Does `eth_call` still operate under gas constraints, but just doesn't withdraw the balance?

axic:
@redsquirrel I think it depends on the implementation, but in general it should work like that.

`call` should be able to do (and access) everything a transaction can, except all changes made during the `call` are *ignored*.

During the `call` an `sstore` and subsequent `sload` should work as desired. So either one implements a specific sandbox or just rolls back the changes after.

chriseth:
Shouldn't the gas limit be much higher for a call?

axic:
@chriseth hmm, maybe.

That change belongs to a separate PR in my opinion as the VM mode needs to be changed too. And perhaps the UI as well? Or let us just settle at 10 times the normal gas?

PR #140, opened 2016-07-22
Provide formal verification input.

chriseth:


index.html
```diff
@@ -125,6 +126,19 @@
...
+						This feature is still in development and thus also not yet well documented,
+						but you can find some information
+						<a href="http://solidity.readthedocs.io/en/latest/security-considerations.html#formal-verification">here</a>.
+						The compiler generates input to be verified
+						(or report errors). Please paste the text below into
+						<a href="http://why3.lri.fr/try/">http://why3.lri.fr/try/</a>
+						to actually perform the verification.
+						We plan to support direct integration in the future.</p>
+						<textarea id="formalVerificationInput" readonly="readonly"></textarea>
+						<div id="formalVerificationErrors"></div>
```
axic: Maybe this block could be controlled from the script so that the `textarea` is not displayed in case of errors?

src/app/formalVerification.js
```diff
@@ -0,0 +1,30 @@
...
+FormalVerification.prototype.compiling = function () {
+  $('#formalVerificationInput', this.outputElement).val('');
+  $('#formalVerificationErrors').empty();
+};
+
+FormalVerification.prototype.compilationFinished = function (compilationResult) {
+  if (compilationResult.formal !== undefined) {
+    if (compilationResult.formal['why3'] !== undefined) {
+      $('#formalVerificationInput', this.outputElement).val(
+        '(* copy this to http://why3.lri.fr/try/ *)' +
```
axic: This text is duplicated on the HTML.

src/app/formalVerification.js
```diff
@@ -0,0 +1,30 @@
...
+  this.renderer = renderer;
+}
+
+FormalVerification.prototype.compiling = function () {
+  $('#formalVerificationInput', this.outputElement).val('');
+  $('#formalVerificationErrors').empty();
+};
+
+FormalVerification.prototype.compilationFinished = function (compilationResult) {
+  if (compilationResult.formal !== undefined) {
```
axic: Shouldn't it display a message if formal verification is lacking in the solc version used?

src/app/formalVerification.js
chriseth: The problem is that I want to keep the textarea small. Its only purpose is to be able to copy it to the clipboard - you can read it after you pasted it into the why3 tool. Because it is small, it will only display the first line of text and that is (due to whatever reason) empty, so I added this prefix here to make it clear _what_ to copy.

index.html
chriseth: Added the feature.

src/app/formalVerification.js
chriseth: Yep!

axic:
@chriseth I've missed one thing. The file naming convention so far was lowercase with dashes.

PR #142, opened 2016-07-25
remix - vm integration

yann300:
merge after https://github.com/ethereum/remix/pull/63

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "webworkify": "^1.2.1",
     "yo-yo": "^1.2.2",
     "yo-yoify": "^3.3.0",
-    "ethereum-remix": "0.0.2-alpha.0.0.4"
+    "ethereum-remix": "0.0.2-alpha.0.0.7"
   },
   "repository": {
     "type": "git",
--- a/src/app/renderer.js
+++ b/src/app/renderer.js
@@ -4,7 +4,7 @@ var UniversalDApp = require('../universal-dapp.js');
 
 var utils = require('./utils');
 
-function Renderer (editor, executionContext, updateFiles, transactionDebugger) {
+function Renderer (editor, executionContext, updateFiles, transactionDebugger, vm) {
   var detailsOpen = {};
 
   function renderError (message, container, noAnnotations) {
@@ -61,6 +61,9 @@ function Renderer (editor, executionContext, updateFiles, transactionDebugger) {
       });
     }
 
+    vm.stateManager.revert(function () {
+      vm.stateManager.checkpoint();
+    });
     var dapp = new UniversalDApp(udappContracts, {
       mode: executionContext.isVM() ? 'vm' : 'web3',
       web3: executionContext.web3(),
@@ -81,7 +84,7 @@ function Renderer (editor, executionContext, updateFiles, transactionDebugger) {
           .append(textRow('uDApp', combined(contractName, contract['interface'], contract.bytecode), 'deploy'))
           .append(getDetails(contract, source, contractName));
       }
-    }, transactionDebugger);
+    }, transactionDebugger, vm);
 
     var $contractOutput = dapp.render();
 
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -18,6 +18,8 @@ function UniversalDApp (contracts, options, transactionDebugger) {
 
   self.web3 = options.web3;
   self.transactionDebugger = transactionDebugger;
+  transactionDebugger.addProvider('EXTERNAL', self.web3);
+  transactionDebugger.switchProvider('EXTERNAL')
   if (options.mode === 'vm') {
     // FIXME: use `options.vm` or `self.vm` consistently
     options.vm = true;
@@ -25,7 +27,8 @@ function UniversalDApp (contracts, options, transactionDebugger) {
     self.accounts = {};
 
     self.vm = new EthJSVM(null, null, { activatePrecompiles: true, enableHomestead: true });
-
+    transactionDebugger.addProvider('VM', self.vm);
+    transactionDebugger.switchProvider('VM')
     self.addAccount('3cd7232cd6f3fc66a57a6bedc1a8ed6c228fff0a327e169c2bcc5e869ed49511');
     self.addAccount('2ac6c190b09897cd8987869cc7b918cfea07ee82038d492abce033c75c1b1d0c');
   } else if (options.mode !== 'web3') {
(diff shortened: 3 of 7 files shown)


src/universal-dapp.js
```diff
@@ -650,6 +666,7 @@ UniversalDApp.prototype.runTx = function (data, args, cb) {
 
   var tx;
   if (!self.vm) {
+    self.transactionDebugger.switchProvider('EXTERNAL');
```
chriseth: Why do you switch the provider here and not at the point where the radiobutton is changed?

yann300: Just to avoid making executionContext dependent of the debugger.
I thought to put the addProvider('VM') and addProvider('EXTERNAL') function in app.js but at that time the vm is not instanciated. so maybe better to setup everything on the same place.
But I can change that if you think it's clearer.

chriseth: I think it is still better for now to make it depend on the debugger. There is a task to convert browser-solidity into an event-based environment where a provider switch would be an event everyone could listen on and act accordingly. It is better to have these calls in one place.

chriseth:
I think the "debug" button that is part of the transaction should be a bit smaller. Perhaps a small button with a bug-symbol not larger than a text line is better.

package.json
```diff
@@ -32,7 +32,7 @@
     "webworkify": "^1.2.1",
     "yo-yo": "^1.2.2",
     "yo-yoify": "^3.3.0",
-    "ethereum-remix": "0.0.2-alpha.0.0.4"
+    "ethereum-remix": "0.0.2-alpha.0.0.7"
```
axic: Out of curiosity, why does this have such a weird versioning system?

src/universal-dapp.js
```diff
@@ -700,7 +718,12 @@ UniversalDApp.prototype.runTx = function (data, args, cb) {
         transactions: [],
         uncleHeaders: []
       });
-      self.vm.runTx({block: block, tx: tx, skipBalance: true, skipNonce: true}, cb);
+      self.vm.runTx({block: block, tx: tx, skipBalance: true, skipNonce: true}, function (err, result) {
+        if (self.vm) {
```
axic: This check is not needed, since it is run on self.vm.runTx.

src/universal-dapp.js
```diff
@@ -700,7 +718,12 @@ UniversalDApp.prototype.runTx = function (data, args, cb) {
...
+          result.transactionHash = self.transactionDebugger.web3().releaseCurrentHash(); // used to keep track of the transaction
```
axic: Where is this hash coming from? What is the relation between web3 and the vm?

package.json
yann300: when you publish to npm, we always have to incr the version (i did not found any other solution). the suffix will not be kept for normal version. This is only used when i have to publish to npm often.

src/universal-dapp.js
yann300: web3() is returning a web3 provider (the one loaded here is a custom provider that rely on the VM).
this hash is used to keep track of the transaction so that it is possible to retrieve it later (for now the instance of the  VM does not include any memory support). but it will certainly come later.

package.json
axic: You could also just increase the patch level, `0.0.2` to `0.0.3`.

Also if you dont really need an npm release (yet) and the only reason it is on npm to use it here, you could replace the whole with a github url.  Npm will download it that way also.

chriseth:
Could you move the debug button next to the close button, and not on its own line?

src/universal-dapp.js
```diff
@@ -18,14 +17,13 @@ function UniversalDApp (contracts, options, transactionDebugger) {
...
   self.web3 = options.web3;
   self.transactionDebugger = transactionDebugger;
+
   if (options.mode === 'vm') {
     // FIXME: use `options.vm` or `self.vm` consistently
     options.vm = true;
 
     self.accounts = {};
-
-    self.vm = new EthJSVM(null, null, { activatePrecompiles: true, enableHomestead: true });
```
axic: I think it is a slippery slope moving this out of udapp.  It would be better designing a real interface.

src/app/renderer.js
```diff
@@ -61,6 +61,9 @@ function Renderer (editor, executionContext, updateFiles, transactionDebugger) {
       });
     }
 
+    vm.stateManager.revert(function () {
+      vm.stateManager.checkpoint();
```
axic: Is revert rolling back *and* removing the checkpoint that's why you need/want to add it again?

yann300: yes

src/universal-dapp.js
yann300: we need to move the instanciation of the vm away from udapp because it is possible now to switch the provider used by the debugger (web3 or VM) independently of udapp (udapp may not be instanciated when the user switch to the vm provider).
Also as the vm has kind of the same role as web3 maybe it makes sense to wrap the vm into an object which provides the same interface as web3. 
There's certainly room for improvement, but maybe better to address that in an another PR.
switching to an event based env could be a nice opportunity

axic: The big problem with wrapping it into `web3` is that the VM supports a much more advanced interface, which due to RPC limitations, cannot exists in `web3`.

There's no way to get detailed execution and output information from web3.

chriseth:
@axic is your comment blocking the merge?

axic:
@yann300 @chriseth I've reviewed now how this works in Remix.  I think the best would be having the trace interface defined in Remix (and not part of the Web3 providers).  And we could implement that trace interface in udapp.

My reasoning is that the VM interface is a superset of Web3, while the current situation makes us to take Web3 as a superset.

Actually, part of the udapp interface is already doing that - moving web3 and vm under a single interface.

yann300:
yes I see your concern about the fact the vm has much more advanced features than web3.
Generally remix need to be compliant with the web3 interface. This is why web3vmprovider wraps the vm into the web3 interface.
about implementing the trace interface in udapp, I thought about that and it was actually working like that before I created the PR. But having the vm trace implemented in remix will allow us to use the combination remix/vm outside of browser solidity.

chriseth:
Some glue-layer would make sense, but perhaps we could do this as a subsequent refactoring (to be done as immediate next stop)? I am not a fan of long-standing PRs.

chriseth:
I think moving the vm creation out of udapp is the correct decision. What still needs to be done is some proper interface for accessing it. Merging this so that we can do a subsequent refactor of both udapp and the debugger interface.

PR #149, opened 2016-07-27
Exclude abstract (interface) contracts from the right hand panel

axic:
Fixes https://github.com/chriseth/browser-solidity/issues/77.

The check is to exclude any contract which doesn't have a bytecode.

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -152,6 +152,12 @@ UniversalDApp.prototype.render = function () {
     self.$el.append(self.getABIInputForm());
   } else {
     for (var c in self.contracts) {
+      // This is an abstract contract. Do not display.
+      // FIXME: maybe have a flag for this in the JSON?
+      if (self.contracts[c].bytecode.length === 0) {
+        continue;
+      }
+
       var $contractEl = $('<div class="contract"/>');
 
       if (self.contracts[c].address) {
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -216,11 +216,19 @@ UniversalDApp.prototype.getCreateInterface = function ($container, contract) {
   var $atButton = $('<button class="atAddress"/>').text('At Address').click(function () { self.clickContractAt(self, $container.find('.createContract'), contract); });
   $createInterface.append($atButton);
 
+  var $newButton = self.getInstanceInterface(contract);
+  $createInterface.append($newButton);
+
   // Only display creation interface for non-abstract contracts.
   // FIXME: maybe have a flag for this in the JSON?
-  if (contract.bytecode.length !== 0) {
-    var $newButton = self.getInstanceInterface(contract);
-    $createInterface.append($newButton);
+  // FIXME: maybe fix getInstanceInterface() below for this case
+  if (contract.bytecode.length === 0) {
+    var $createButton = $newButton.find('.constructor .call');
+
+    // NOTE: we must show the button to have CSS properly lined up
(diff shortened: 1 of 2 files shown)


Denton-L:
This seems to be broken, only the title is removed; everything else is still there.
2016-07-28-105409_1366x731_scrot

Denton-L:
Also, would it be possible to remove the Bytecode textbox since it's empty?

axic:
No abstract contract should be visible at all.

The contract `B` is abstract because it defines only function interfaces.

The contract `A` is not abstract, because it only has fully defined functions. In this case the fallback interface.

Until https://github.com/ethereum/solidity/issues/649 is sorted, it is a bit crude figuring out what is abstract.

axic:
Ah I see actually the fields are visible. Weird, it seems to work locally here.

axic:
@Denton-L fixed

Denton-L:
Alright, it works now!

Question though, why is the "At Address" button removed? Take, for example, this contract:

```
contract Abstract {
    function doThing();
}


contract Inherit is Abstract {
    function doThing() {}
}
```

Shouldn't it be possible to add `Abstract` using "At Address" where you're feeding in an instance of `Inherit`?

axic:
Abstract contract is never deployed to the blockchain, because it hasn't got any bytecode. It is only used during compilation.

Denton-L:
Abstract contracts aren't deployed to the blockchain, yes, but it's possible for a contract that inherits from the abstract contract to be deployed onto the blockchain.

What if I wanted to attach to an inheriting contract, but to the inheriting parent? Using the example above, that should be possible because any instance of `Inherit` should also be an instance of `Abstract`.

axic:
@Denton-L I see, I guess that could be possible, albeit I've never used the "at address" button.

axic:
@Denton-L it could be implemented in a future PR as it requires a lot of changes. Right now the "at address" is tied to having the HTML created by the "contract" button.

Denton-L:
Could you possibly stick in a `$('.call').remove()` in there somewhere?

axic:
@Denton-L I'm afraid it is really messy.

Denton-L:
Alright, thanks anyway! 👍

chriseth:
I think we should not remove the "at address" feature for abstract contracts, so this cannot be merged as is.

axic:
Changed it to show a disabled Create button.

src/universal-dapp.js
```diff
@@ -208,9 +208,23 @@ UniversalDApp.prototype.getCreateInterface = function ($container, contract) {
...
+
+  // Only display creation interface for non-abstract contracts.
+  // FIXME: maybe have a flag for this in the JSON?
+  // FIXME: maybe fix getInstanceInterface() below for this case
+  if (contract.bytecode.length === 0) {
+    var $createButton = $newButton.find('.constructor .call');
+
+    // NOTE: we must show the button to have CSS properly lined up
+    $createButton.attr('disabled', 'disabled');
+    $createButton.text('Create');
```
chriseth: We could perhaps provide a better hint to the user why this button is disabled? I cannot come up with a better text for the button, but perhaps we can provide a title-popup?

axic: The button has the special cursor if you hover over. The text could be changed though if you have an idea.

chriseth: `title="This contract does not implement all functions and thus cannot be created."`

axic: Added!

PR #151, opened 2016-07-27
Introduce CSS lint

axic:
Running csslint currently show 271 errors/warnings.

### DIFF
--- a/assets/css/universal-dapp.css
+++ b/assets/css/universal-dapp.css
@@ -54,7 +54,6 @@
 
 .udapp .title {
 	margin-bottom: 0.4em;
-	display: inline-block;
 	padding: 0.2em;
 	background-color: rgba( 255,255,255,0.5 );
 	display: block;
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - stable
 script:
-  - npm run lint && npm run test && npm run downloadsolc && npm run build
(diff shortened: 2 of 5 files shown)


axic:
@Denton-L since you have looked at CSS before, are you interested in looking at the warnings reported and see if it makes sense fixing any of them?

axic:
Disabled the rule for alphabetical order. Maybe it is useful?

We could disable other rules too if needed.

Denton-L:
I'm not very well-versed at CSS but I'll give it a look.

Denton-L:
This is outside the scope of my knowledge as I'm not really a web developer... I'll have to hand this off to someone else.

axic:
@chriseth @Denton-L can you test and review this? It only fails on serious errors but reports the warnings also.

Denton-L:
In my browser, (Chromium 52.0.2743.116 on Linux), the size of the contract seems to be displayed strangely. See below.

2016-08-31-143147_442x401_scrot

axic:
Have look at the last two changes. Im not sure the first or the second of the duplicate entries take precedence.

assets/css/universal-dapp.css
```diff
@@ -57,7 +57,6 @@
 	display: inline-block;
 	padding: 0.2em;
 	background-color: rgba( 255,255,255,0.5 );
-	display: block;
```
chriseth: The other one should be removed.

axic: Fixed.

PR #156, opened 2016-07-28
Dont display empty compiler output fields

### LINKED ISSUE #155: Don't display empty compiler output fields
Currently, in some cases (abstract contracts, empty contracts) the bytecode or interface fields are empty.

Empty fields probably shouldn't be displayed.

(This may depend on https://github.com/ethereum/solidity/issues/758)

axic:
Fixes #155.

### DIFF
--- a/src/app/renderer.js
+++ b/src/app/renderer.js
@@ -77,7 +77,9 @@ function Renderer (editor, executionContext, updateFiles, transactionDebugger) {
         if (contract.bytecode) {
           $contractOutput.append(textRow('Bytecode', contract.bytecode));
         }
+
         $contractOutput.append(textRow('Interface', contract['interface']));
+
         if (contract.bytecode) {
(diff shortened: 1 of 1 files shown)


axic:
@Denton-L here's the other change to make it even more intelligent :)

Denton-L:
One small issue (sorry!), for contract `A` in the following example, I feel like it should display the interface, even though it's only `[]` because `[]` is still a valid interface. But this is just a minor edge-case.

2016-07-28-135832_1366x731_scrot

axic:
> One small issue (sorry!), for contract A in the following example, I feel like it should display the interface, even though it's only [] because [] is still a valid interface.

Not sure how useful that is, but I could be convinced.

Denton-L:
I suppose it's just semantics but if there isn't an interface line displayed, it would imply that `A` does not have an interface at all, which isn't true because all contracts have interfaces, whereas displaying `[]` would tell the user that `A` has an interface, and it's `[]`.

Issue #145, opened 2016-07-27
Formal verification errors are not removed


chriseth:

chriseth:
This seems to work now.

PR #167, opened 2016-08-04
Refactoring:: add events

yann300:
link to https://github.com/ethereum/browser-solidity/issues/138

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -28,7 +28,7 @@ window.addEventListener('message', function (ev) {
   }
 }, false);
 /*
-  trigger selectTab
+  trigger tabChanged
 */
 var run = function () {
   var self = this;
@@ -411,7 +411,6 @@ var run = function () {
     setEditorSize(hidingRHP ? 0 : storage.getEditorSize());
     $('.toggleRHP i').toggleClass('fa-angle-double-right', !hidingRHP);
     $('.toggleRHP i').toggleClass('fa-angle-double-left', hidingRHP);
-    if (!hidingRHP) compiler.compile();
   });
 
   // ----------------- editor resize ---------------
@@ -436,9 +435,9 @@ var run = function () {
     $('#output').append($('<div/>').append($('<pre/>').text('Loading github.com/' + root + '/' + path + ' ...')));
     return $.getJSON('https://api.github.com/repos/' + root + '/contents/' + path, cb);
   }
-  var transactionDebugger = new Debugger('#debugger');
-  var executionContext = new ExecutionContext();
 
+  var executionContext = new ExecutionContext();
+  var transactionDebugger = new Debugger('#debugger', executionContext.event);
   transactionDebugger.addProvider('VM', executionContext.vm());
   transactionDebugger.switchProvider('VM');
   transactionDebugger.addProvider('INTERNAL', executionContext.web3());
@@ -456,43 +455,25 @@ var run = function () {
     transactionDebugger.debug(data);
   });
 
-  var renderer = new Renderer(editor, executionContext.web3(), updateFiles, udapp, executionContext);
-  var formalVerification = new FormalVerification($('#verificationView'));
-
-  formalVerification.event.register('compilationError', this, function (message, container, noAnnotations) {
-    renderer.error(message, container, noAnnotations);
-  });
-
   var compiler = new Compiler(editor, queryParams, handleGithubCall, updateFiles);
+  var formalVerification = new FormalVerification($('#verificationView'), compiler.event);
+  var renderer = new Renderer(editor, executionContext.web3(), updateFiles, udapp, executionContext, formalVerification.event, compiler.event); // eslint-disable-line
 
   executionContext.event.register('contextChanged', this, function (context) {
-    $('#output').empty();
-    context = context === 'vm' ? 'VM' : context;
-    context = context === 'injected' ? 'EXTERNAL' : context;
-    context = context === 'web3' ? 'INTERNAL' : context;
-    transactionDebugger.switchProvider(context);
     compiler.compile();
   });
+
   executionContext.event.register('web3EndpointChanged', this, function (context) {
-    $('#output').empty();
     compiler.compile();
   });
+
+  executionContext.event.register('compilerLoaded', this, function (context) {
+    compiler.compile();
+  });
+
   compiler.event.register('compilerLoaded', this, function (version) {
     setVersionText(version);
     compiler.compile();
-  });
-  compiler.event.register('compilationError', this, function (data) {
-    renderer.error(data);
-  });
-  compiler.event.register('compilationSucceed', this, function (data, source) {
-    if (!hidingRHP) {
-      renderer.contracts(data, source);
-      formalVerification.compilationFinished(data);
-    }
-  });
-  compiler.event.register('isCompiling', this, function () {
-    $('#output').empty();
-    formalVerification.compiling();
   });
 
   function setVersionText (text) {
--- a/src/app.js
+++ b/src/app.js
@@ -16,7 +16,7 @@ var ExecutionContext = require('./app/execution-context');
 var UniversalDApp = require('./universal-dapp.js');
 var Debugger = require('./app/debugger');
 var FormalVerification = require('./app/formalVerification');
-var util = require('./lib/util');
+var EventManager = require('./lib/eventManager');
 
 // The event listener needs to be registered as early as possible, because the
 // parent will send the message upon the "load" event.
@@ -32,7 +32,7 @@ window.addEventListener('message', function (ev) {
 */
 var run = function () {
   var self = this;
-  util.makeEventCapable(this);
+  this.event = new EventManager();
   var storage = new Storage(updateFiles);
 
   function loadFiles (files) {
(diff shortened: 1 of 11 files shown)


src/lib/eventManager.js
```diff
@@ -0,0 +1,32 @@
...
+eventManager.prototype.unregister = function (eventName, obj) {
+  for (var reg in this.registered[eventName]) {
+    if (this.registered[eventName][reg] && this.registered[eventName][reg].obj === obj) {
+      this.registered[eventName].splice(reg, 1)
+      return
+    }
+  }
+}
+
+eventManager.prototype.register = function (eventName, obj, func) {
```
chriseth: Can we make `obj` optional?

src/lib/util.js
```diff
@@ -0,0 +1,14 @@
...
+var EventManager = require('./eventManager')
+
+module.exports = {
+  extend: function (destination, source) {
+    for (var property in source) {
+      destination[property] = source[property]
+    }
+  },
+
+  produceEvent: function (destination) {
```
chriseth: Better to call this something like `makeEventCapable`?

On the other hand: Especially because the functions `register` and `trigger` have extremely common names, it would perhaps be better to not just extend the object, but rather have a property called `event` that is the EventManager.

Then we would have `this.event.trigger('selectTab',  [cls])` and `otherObject.event.register('changed', function() { >>> })`.

src/app.js
```diff
@@ -108,6 +111,7 @@ var run = function () {
       el.removeClass('active');
       $('#optionViews').removeClass(cls);
     }
+    this.trigger('selectTab',  [cls])
```
chriseth: double space

axic:
Is there a way to somehow split this into smaller incremental changes? It is hard to review this way.

yann300:
Basically this PR remove unneeded dependencies and replace them by events (also I moved some ui functions to ui-helper.js)
There are indeed a lot of changes but it is hard to be more atomic.
This is not a new feature but an architectural change that imply those changes. 
Maybe better to review it as a new code instead of checking code changes.
Also it is a good opportunity to write ui tests (i don't think there are so many), so we can verify the whole integrity (i can help with that)

src/lib/util.js
```diff
@@ -0,0 +1,14 @@
...
+var EventManager = require('./eventManager');
+
+module.exports = {
+  extend: function (destination, source) {
+    for (var property in source) {
+      destination[property] = source[property];
+    }
+  },
+
+  makeEventCapable: function (destination) {
```
chriseth: Perhaps it would be better to just make the `destination.event = new EventManager()` assignment explicit inside destination? At the point wheer `makeEventCapable` is called, it is not really visible what happens.

src/lib/eventManager.js
```diff
@@ -0,0 +1,41 @@
...
+  for (var reg in this.registered[eventName]) {
+    if (this.registered[eventName][reg] &&
+      this.registered[eventName][reg].obj === obj && (!func || this.registered[eventName][reg].func === func)) {
+      this.registered[eventName].splice(reg, 1);
+      return;
+    }
+  }
+};
+
+eventManager.prototype.register = function (eventName, obj, func) {
```
chriseth: Please document.

src/app.js
```diff
@@ -433,19 +437,63 @@ var run = function () {
...
+    compiler.compile();
+  });
+  compiler.event.register('compilerLoaded', this, function (version) {
+    setVersionText(version);
+    compiler.compile();
+  });
+  compiler.event.register('compilationError', this, function (data) {
+    renderer.error(data);
+  });
+  compiler.event.register('compilationSucceed', this, function (data, source) {
```
chriseth: `compilationSucceeded`

src/app.js
```diff
@@ -433,19 +437,63 @@ var run = function () {
...
+    if (!hidingRHP) {
+      renderer.contracts(data, source);
+      formalVerification.compilationFinished(data);
+    }
+  });
+  compiler.event.register('isCompiling', this, function () {
```
chriseth: `compilationStarted`?

src/app.js
```diff
@@ -433,19 +437,63 @@ var run = function () {
...
+    $('#output').empty();
+    formalVerification.compiling();
```
chriseth: Would it be better if the formal verification component could register its events itself rather than having to go through the dapp as a hub? I think it would be better if everything could be more self-contained, at least if we restrict to some basic events that are global and general like "compilation started", "compilation finished" and so on.

src/app/compiler.js
```diff
@@ -115,13 +118,12 @@ function Compiler (editor, renderer, queryParams, handleGithubCall, outputField,
 
     if (missingInputs !== undefined && missingInputs.length > 0) {
       compile(missingInputs);
-    } else if (noFatalErrors && !hidingRHP()) {
-      renderer.contracts(data, editor.getValue());
-      formalVerification.compilationFinished(data);
+    } else if (noFatalErrors) {
+      self.event.trigger('compilationSucceed', [data, editor.getValue()]);
```
chriseth: I think we should have another event `compilationFailed`, or have one event `compilationFinished` where we pass success condition. It would also be good if other components could register that event and retrieve the list of errors as part of that callback.

chriseth:
Please document the events of a class at the top of each class.

src/lib/util.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+module.exports = {
+  extend: function (destination, source) {
```
chriseth: Is this still used somewhere?

src/app/renderer.js
```diff
@@ -1,276 +1,118 @@
...
+        column: errCol,
+        text: message,
+        type: type
       });
     }
+    $error.click(function (ev) {
+      if (errFile !== '' && errFile !== utils.fileNameFromKey(this.editor.getCacheFile()) && this.editor.hasFile(errFile)) {
+        // Switch to file
+        this.editor.setCacheFile(utils.fileKey(errFile));
+        this.updateFiles();
```
chriseth: Can `updateFiles` be done via events?

src/app.js
```diff
@@ -411,7 +411,6 @@ var run = function () {
     setEditorSize(hidingRHP ? 0 : storage.getEditorSize());
     $('.toggleRHP i').toggleClass('fa-angle-double-right', !hidingRHP);
     $('.toggleRHP i').toggleClass('fa-angle-double-left', hidingRHP);
-    if (!hidingRHP) compiler.compile();
```
chriseth: Where is this done now?

src/app/compiler.js
```diff
@@ -24,6 +24,18 @@ function Compiler (editor, queryParams, handleGithubCall, updateFiles) {
 
   var compileTimeout = null;
 
+  executionContextEvent.register('contextChanged', this, function (context) {
```
chriseth: Hm, are you sure about this? I think these three would probably be better done in the app.

src/app/formalVerification.js
```diff
@@ -4,9 +4,16 @@ var EventManager = require('../lib/eventManager');
...
+  var self = this;
+  compilerEvent.register('compilationSucceeded', this, function (data, source) {
+    self.compilationFinished(data);
+  });
+  compilerEvent.register('compilationStarted', this, function () {
+    self.compiling();
+  });
 }
 
 FormalVerification.prototype.compiling = function () {
```
chriseth: Do we need this as its own function? Can it just go into the event callback?

src/app.js
yann300: Before contracts were only rendered if (hidingRHP ==  false), now contracts are rendered in each case (even if the panel is hidden) so we don't need to recompile.
I know that it could look like a waste of resources to recompile even if the panel is hidden but:
 - In most of the case, the panel will not be hidden cause compilation errors are displayed in this panel.
 - Basicallly i wanted to remove getHidingRHP as a dependency of renderer.js. It looked like an unnecessary dependency.

chriseth: ok, that's fine. It is better to have a separate button to disable compilation.

src/app/compiler.js
```diff
@@ -115,13 +118,12 @@ function Compiler (editor, renderer, queryParams, handleGithubCall, outputField,
 
     if (missingInputs !== undefined && missingInputs.length > 0) {
       compile(missingInputs);
-    } else if (noFatalErrors && !hidingRHP()) {
-      renderer.contracts(data, editor.getValue());
-      formalVerification.compilationFinished(data);
+    } else if (noFatalErrors) {
+      self.event.trigger('compilationSucceeded', [data, editor.getValue()]);
```
chriseth: I think it would be better to have a `compilationFinished` event, with a flag that tells if it was successful or not.

src/app/compiler.js
```diff
@@ -102,14 +102,14 @@ function Compiler (editor, queryParams, handleGithubCall, updateFiles) {
...
-      self.event.trigger('compilationError', [data['error']]);
+      self.event.trigger('compilationFinished', [false, data['error'], editor.getValue()]);
       if (utils.errortype(data['error']) !== 'warning') {
         noFatalErrors = false;
       }
     }
     if (data['errors'] !== undefined) {
       data['errors'].forEach(function (err) {
-        self.event.trigger('compilationError', [err]);
+        self.event.trigger('compilationFinished', [false, err, editor.getValue()]);
```
chriseth: I thought `compilationError` was for just reporting a single error? `compilationFinished` should only be triggered once per compilation run.

chriseth:
Ready for merge from my side.

chriseth:
Fixes #145

PR #172, opened 2016-08-11
Only use latest released version by default.

chriseth:


src/app.js
```diff
@@ -466,7 +469,13 @@ var run = function () {
...
 
   if (window.soljsonSources !== undefined) {
-    loadVersion(queryParams.get().version || 'soljson-latest.js');
+    var latestRelease = 'soljson-latest.js';
+    if (window.soljsonReleases !== undefined) {
+      for (var release in window.soljsonReleases) {
+        latestRelease = release;
+      }
+    }
+    loadVersion(queryParams.get().version || release);
```
yann300: lastestRelease instead of release

axic:
Once we get https://github.com/ethereum/solc-bin/pull/4 merged, we could make this much simpler.

chriseth:
Indeed, I actually looked at it, but it was closed. We now have node at the deploy machine, so no problem in running javascript.

PR #176, opened 2016-08-17
source code debugging

yann300:


src/app/sourceHighlighter.js
```diff
@@ -0,0 +1,206 @@
...
+    } else {
+      if (self.currentRange) {
+        self.currentMarker = self.editor.addMarker(self.currentRange, 'highlightcode')
+      }
+    }
+  })
+
+  /* update compilation data */
+  compilerEvent.register('compilationFinished', this, function (success, data, source) {
+    if (!self.isDebugging) {
```
chriseth: Do we really want this check? Don't we usually start debugging _after_ the compilation finished?

src/app/sourceHighlighter.js
```diff
@@ -0,0 +1,206 @@
...
+      self.reset()
+      self.compilationData = success ? data : null
```
chriseth: Is it possible to store only part of the result data?

src/app/sourceHighlighter.js
```diff
@@ -0,0 +1,206 @@
...
+            self.sourceMappingDecoder.getLineColumnPosition(rawPos, self.currentLineColumnLayout, function (error, pos) {
+              if (!error) {
+                self.highlight(pos, rawPos.file)
+              }
+            })
+          }
+        })
+      } else {
+        self.sourceMappingDecoder.getLineColumnPosition(rawPos, self.currentLineColumnLayout, function (error, pos) {
+          if (!error) {
```
chriseth: If there was an error, we should not just keep the previous highlight range.

src/app/sourceHighlighter.js
```diff
@@ -0,0 +1,206 @@
...
+    * highlight the statement with the given @arg position
+    *
+    * @param {Object} position - the position to highlight { start: {line, column}, end: {line, column} }
+*/
+SourceHighlighter.prototype.highlight = function (position, fileIndex) {
+  var name = utils.fileNameFromKey(this.editor.getCacheFile()) // current opened tab
+  var source = this.compilationData.sourceList[parseInt(fileIndex)] // auto switch to that tab
+  this.removeCurrentMarker()
+  if (name !== source) {
+    this.switchToFile(source) // command the app to swicth to the curent file
```
chriseth: Will that trigger a recompile?

src/app/compiler.js
```diff
@@ -119,7 +123,11 @@ function Compiler (editor, queryParams, handleGithubCall, updateFiles) {
     if (missingInputs !== undefined && missingInputs.length > 0) {
       compile(missingInputs);
     } else if (noFatalErrors) {
-      self.event.trigger('compilationFinished', [true, data, editor.getValue()]);
+      self.lastCompilationResult = {
+        data: data,
+        source: source
+      }
+      self.event.trigger('compilationFinished', [true, data, source]);
```
chriseth: Is the event triggered twice if there were only warnings?

src/app/debugger.js
```diff
@@ -1,39 +1,132 @@
...
+  // register selected code item, highlight the corresponding source location
+  this.debugger.codeManager.register('changed', this, function (code, address, index) {
+    this.debugger.sourceLocationTracker.getSourceLocation(address, index, self.lastCompilationResult.data.contracts, function (error, rawLocation) {
+      if (!error) {
+        if (!self.cache.lineBreakPositionsByContent[address]) {
+          self.cache.lineBreakPositionsByContent[address] = self.sourceMappingDecoder.getLinebreakPositions(self.editor.getFile(self.lastCompilationResult.data.sourceList[rawLocation.file]))  
+        }
+        var lineColumnPos = self.sourceMappingDecoder.convertOffsetToLineColumn(rawLocation, self.cache.lineBreakPositionsByContent[address])
+        self.highlight(lineColumnPos, rawLocation);
+      }
```
chriseth: Unset highlight if there was an error?

src/app/debugger.js
```diff
@@ -1,39 +1,132 @@
...
 };
 
+/**
+ * highlight the given @arg lineColumnPos
+ *
+ * @param {Object} lineColumnPos - position of the source code to hightlight {start: {line, column}, end: {line, column}}
+ * @param {Object} rawLocation - raw position of the source code to hightlight {start, length, file, jump}
+ */
+Debugger.prototype.highlight = function (lineColumnPos, rawLocation) {
+  var name = utils.fileNameFromKey(this.editor.getCacheFile()); // current opened tab
```
chriseth: Could also be a different file, but I think it's better to do that in a separate PR because we first need the ability to switch tabs without triggering recompilation.

src/app/ui-helper.js
```diff
@@ -146,11 +146,12 @@ module.exports = {
       details.append(this.tableRow('Runtime Bytecode', contract.runtimeBytecode));
     }
 
+    /*
```
chriseth: This is still useful information.

src/app/debugger.js
yann300: can we still keep that for this PR and improve this in another PR?

src/app/compiler.js
yann300: no it is no triggered twice

src/app/debugger.js
chriseth: yes

src/app/debugger.js
```diff
@@ -1,39 +1,135 @@
...
+
+  this.editor.onChangeSetup(function () {
+    if (arguments.length > 0) { // if arguments.length === 0 this is a session change, we don't want to stop debugging in that case
+      self.debugger.unLoad()
+    }
+  });
+
+  // register selected code item, highlight the corresponding source location
+  this.debugger.codeManager.register('changed', this, function (code, address, index) {
+    this.debugger.sourceLocationTracker.getSourceLocation(address, index, self.lastCompilationResult.data.contracts, function (error, rawLocation) {
```
chriseth: This works for now, but in the future, we should not just use "last compilation results". This does not work if you want to debug a past transaction, for example.

src/app/compiler.js
chriseth: Why? If there is an `errors` property, but it only contains warnings, then the event is triggered in line 115 and in this line.

yann300:
contains fix for https://github.com/ethereum/browser-solidity/issues/185

chriseth:
Please merge as soon as the functionality from remix is available in npm.

PR #182, opened 2016-08-23
Correct how to find latest release version.

chriseth:
Still to do:
 - [x] somehow old versions are not loaded correctly

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,6 +1,7 @@
 /* global alert, confirm, prompt, Option, Worker, soljsonSources */
 
 var $ = require('jquery');
+var semver = require('semver');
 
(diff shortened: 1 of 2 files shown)


chriseth:
Loading old versions does seem to work.

src/app.js
```diff
@@ -498,10 +499,15 @@ var run = function () {
   };
 
   if (window.soljsonSources !== undefined) {
-    var latestRelease = 'soljson-latest.js';
```
axic: This was used as a fallback name for the built-in verison.

chriseth: But it still works like that, doesn't it?

PR #188, opened 2016-08-30
Update ethereumjs-abi dependency to 0.6.4.

chriseth:


chriseth:
I'm actually not sure if this is needed or if the new version is picked up automatically with the next build.

axic:
@chriseth it should be picked up as it was a patch release. Because it is major=0, any increase in the minor will not be picked up.

axic:
For some reason the last build didn't take it.

PR #197, opened 2016-08-31
Order entires alphabetically in package.json

axic:
Using `npm install --save <new package>` will not rewrite the package.json after this.


axic:
Obsoleted by #151 (it contained the same change).

PR #200, opened 2016-08-31
Move the solc lookup URL to the proper location

axic:
This is the first part of a set of changes.


src/app.js
```diff
@@ -352,7 +352,6 @@ var run = function () {
   $('#versionSelector').append(new Option('latest local version', 'soljson.js'));
 
   $('#versionSelector').change(function () {
-    queryParams.update({ version: $('#versionSelector').val() });
```
chriseth: Where is this done now?

axic: `loadVersion` does the same already, it is being done twice currently

PR #202, opened 2016-08-31
Correctly fall back to soljson-latest in the offline case.

chriseth:


chriseth:
@axic did you take a look at this?

PR #205, opened 2016-09-01
Contract detail cleanup

axic:


src/app/ui-helper.js
```diff
@@ -21,6 +21,13 @@ module.exports = {
       cls);
```
chriseth: I think we should also use `<input readonly="readonly"` here

src/app/ui-helper.js
```diff
@@ -125,35 +136,34 @@ module.exports = {
...
-      details.append($('<span class="col1">Gas Estimates</span>'));
-      details.append($('<pre/>').text(gasEstimates));
+      details.append(this.preRow('Gas Estimates', gasEstimates));
     }
 
     if (contract.runtimeBytecode && contract.runtimeBytecode.length > 0) {
       details.append(this.tableRow('Runtime Bytecode', contract.runtimeBytecode));
     }
 
+    if (contract.opcodes !== '') {
```
chriseth: Can you change it to `!== undefined && !== ''` in case we remove from abstract contracts it in the future?

src/app/ui-helper.js
axic: The textarea has the benefit that it can be resized and the data made readable. With input it needs to be copy pasted to the outside.

chriseth: If we use input, it looks much more uniform. Also I don't think that anyone reads one of these outputs - they have to be reformatted before they are readable, so you have to copy-past them into a smart editor anyway.

chriseth:
Oh another thing: is the "udapp" output field actually useful?

src/app/ui-helper.js
axic: We could pretty print the JSON. Maybe it should be part of a new feature.

axic:
> Oh another thing: is the "udapp" output field actually useful?

I don't think it useful currently. The same applies to the `geth deploy` field. Why should anyone use that?

src/app/ui-helper.js
chriseth: The json is really not meant for human consumption.

chriseth:
`geth deploy` can at least be pasted somewhere. And it is used in the tutorial.

PR #206, opened 2016-09-01
Remove Solidity Interface from the output

axic:


chriseth:
This also removes it if an older compiler version is selected. Also included this feature here: https://github.com/ethereum/browser-solidity/pull/207

axic:
Yes it does, but is it needed with an older one?

chriseth:
It is not "needed", but we can provide it.

axic:
Wouldn't that be a cause of complaint? (*Why does it work with an older version and not with the latest one?*)

chriseth:
The better argument is probably: We should not provide an incentive to use old, potentially buggy compiler versions.

PR #211, opened 2016-09-01
Cleanup version loading

axic:
This reorder loading code for clarity and disables the selector if there's only one version.


axic:
@chriseth changed one more thing: renamed `soljson.js` to builtin.

PR #213, opened 2016-09-06
Parse source into json for event.

chriseth:


axic:
It needs to be added to `compiler-worker` too.

chriseth:
@axic I think in compiler-worker, it never leaves the json-format.

src/app/compiler.js
```diff
@@ -144,7 +144,7 @@ function Compiler (editor, queryParams, handleGithubCall, updateFiles) {
   function loadInternal (url) {
     delete window.Module;
     // Set a safe fallback until the new one is loaded
-    setCompileJSON(function (source, optimize) { compilationFinished('{}'); });
+    setCompileJSON(function (source, optimize) { compilationFinished({}); });
```
axic: Good catch. Shouldn't it be `{ error: 'Compiler not found.' }` or something similar?

src/app/compiler.js
```diff
@@ -173,12 +173,14 @@ function Compiler (editor, queryParams, handleGithubCall, updateFiles) {
           break;
         case 'compiled':
           var result;
+          var source;
           try {
             result = JSON.parse(data.data);
+            source = JSON.parse(data.source);
```
axic: I would prefer to have two have two separate try/catch blocks for this. The `source` one should never trigger.

src/app/compiler.js
```diff
@@ -189,7 +200,9 @@ function Compiler (editor, queryParams, handleGithubCall, updateFiles) {
       compilationFinished({ error: 'Worker error: ' + msg.data });
     });
     compileJSON = function (source, optimize) {
-      worker.postMessage({cmd: 'compile', source: JSON.stringify(source), optimize: optimize});
+      jobs[nextJobId] = {sources: source};
```
axic: `nextJobId` can be replaced here locally with `jobs.length`

chriseth: No, it is not an array. Not sure how javascript implements arrays, but since we usually only have one concurrent job, I think an object is better.

axic: With arrays and `jobs.length` it should work without any issues.

chriseth: oh wow, today I learned about "empty array slots" :-)

axic: Oh you are using delete. Well actually that works and you'll have empty slots in JS (just like in Lua).

Issue #194, opened 2016-08-31
Use the JSON solc-bin list (solc-bin/list.json)


axic:
Currently browser-solidity uses `solc-bin/list.js` which creates a global for releases and builds and there's special code to check if it was loaded.

Loading from JSON is more explicit and controllable from browser-solidity.

axic:
Done in #215.

PR #215, opened 2016-09-06
Load versions from list.json

axic:


src/app.js
```diff
@@ -514,36 +511,52 @@ var run = function () {
...
+      loadVersion('builtin');
+      return;
+    }
+
+    //
+    // populate version dropdown with all available compiler versions
+    //
+
+    // NOTE: the list currently is oldest first
+    $.each(data.builds.reverse(), function (i, build) {
```
axic: Do we want to keep the builds in oldest first order?

axic:
Do not merge yet, only after #216.

src/app.js
chriseth: If we rename the files it does not matter, we can just sort them according to semver.

axic: So you want to sort them according to semver for the dropdown?

src/app.js
```diff
@@ -526,8 +526,14 @@ var run = function () {
     // populate version dropdown with all available compiler versions
     //
 
-    // NOTE: the list currently is oldest first
-    $.each(data.builds.reverse(), function (i, build) {
+    // Sort builds according to semver
+    var builds = data.builds.sort(function (a, b) {
+      a = a.version + '-' + a.prerelease;
+      b = b.version + '-' + b.prerelease;
+      return semver.lt(a, b);
```
axic: Added sorting here, but it doesn't seem to work, the output is fairly random. Any ideas?

chriseth: Are you sure? My local tests look rather good.

src/app.js
```diff
@@ -513,36 +510,58 @@ var run = function () {
...
-  if (window.soljsonSources !== undefined) {
-    // populate selector list with available versions
-    $.each(soljsonSources, function (i, file) {
-      if (file) {
-        var version = file.replace(/soljson-(.*).js/, '$1');
-        $('#versionSelector').append(new Option(version, file));
-      }
+  $.getJSON('https://ethereum.github.io/solc-bin/bin/list.json', function (data, status) {
+    //
+    // loading failed for some reason, fall back to local compiler
```
chriseth: Is this excessive spacing really required at that level?

src/app.js
```diff
@@ -513,36 +510,58 @@ var run = function () {
...
+      return;
+    }
+
+    //
+    // populate version dropdown with all available compiler versions
+    //
+
+    // Sort builds according to semver
+    var builds = data.builds.sort(function (a, b) {
+      a = a.version + '-' + a.prerelease;
```
chriseth: In the future, `a.prerelease` might be empty. In that case, we also have to drop the `'-'`

src/app.js
chriseth: What about `semver.compare(a, b)`? I think this function is required to return -1 / 0 / 1 and not true / false.

axic: That seems to do the trick.

src/app.js
axic: Tried to separate the pieces with comments, reworked that a bit.

src/app.js
axic: Have a method called `buildVersion()` for that now. With change of versioning, we need to update that method only.

src/app.js
```diff
@@ -513,36 +510,58 @@ var run = function () {
...
+    // loading failed for some reason, fall back to local compiler
+    if (status !== 'success') {
+      $('#versionSelector').append(new Option('latest local version', 'builtin'));
+
+      loadVersion('builtin');
+      return;
+    }
+
+    function buildVersion (build) {
+      if (build.prerelease) {
```
chriseth: check for length?

src/app.js
```diff
@@ -513,36 +510,58 @@ var run = function () {
...
+        return build.version;
       }
+    }
+
+    // Sort builds according to semver
+    var builds = data.builds.sort(function (a, b) {
+      a = buildVersion(a);
+      b = buildVersion(b);
+      return semver.compare(a, b);
+    }).reverse();
```
axic: I guess instead of `reverse()` I could use `1 - semver.compare(a, b)`, but this way it is more readable.

axic: Actually, swapping a/b should work

chriseth: Yes, but then please add a comment.

axic:
Fixed #194.

PR #219, opened 2016-09-07
Compile button and option to disable automatic compilation

### LINKED ISSUE #101: Add Feature: Compile Only On Demand
Currently, browser-solidity recompiles every time a change has been made to the file. This may be problematic, however, if the user is working with very large files as the whole browser freezes up to compile every time a change is made to a large project.

There should be an option to only compile on demand.

axic:
Fixes  #101

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -436,6 +436,10 @@ var run = function () {
 
   var renderer = new Renderer(editor, executionContext.web3(), updateFiles, udapp, executionContext, formalVerification.event, compiler.event); // eslint-disable-line
 
+  $('#compile').click(function () {
+    compiler.compile();
+  });
+
   executionContext.event.register('contextChanged', this, function (context) {
     compiler.compile();
   });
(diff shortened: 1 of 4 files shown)


src/app.js
```diff
@@ -436,6 +436,41 @@ var run = function () {
 
   var renderer = new Renderer(editor, executionContext.web3(), updateFiles, udapp, executionContext, formalVerification.event, compiler.event); // eslint-disable-line
 
+  var autoCompile = document.querySelector('#autoCompile').checked;
+
+  document.querySelector('#autoCompile').addEventListener('change', function () {
```
chriseth: We have jquery here, so you could do some live stuff, but it's of course also fine like that.

src/app.js
```diff
@@ -436,6 +436,41 @@ var run = function () {
...
+  var compileTimeout = null;
+
+  function editorOnChange () {
+    if (!autoCompile) {
+      return;
+    }
+
+    var input = editor.getValue();
+    if (input === '') {
+      editor.setCacheFileContent('');
```
chriseth: Don't we want to do this even if we do not compile?

src/app.js
axic: This matches the rest of the file. I think it should be updated at a later stage to only use jQuery.

src/app.js
axic: I don't fully know what is the point of that code, but moved the check past it.

PR #238, opened 2016-09-13
Simplify runTx

axic:


src/universal-dapp.js
```diff
@@ -737,7 +736,7 @@ UniversalDApp.prototype.runTx = function (data, args, cb) {
       data: data,
       value: value
     };
-    if (constant && !isConstructor) {
+    if (args.isCall) {
```
chriseth: Perhaps rename that to `useCall`?

PR #244, opened 2016-09-15
update remix version

yann300:
-> -> 0.0.2-alpha.0.0.9


axic:
I wonder why don't you just keep incrementing the patch version (or follow semver)? :)

chriseth:
Merging for now, but I think in the long run, we should switch to direct github url.

PR #250, opened 2016-10-04
Fix testing and add compilation test.

chriseth:
This removes some unnecessary timeouts from tests and especially adds the first test that compiles a source file.

Lessons learnt from these tests:
- It takes more than 30 seconds to load the compiler in the test environment

Because of that, I would propose to put as many tests as possible into single test cases that only need to load the compiler once.


chriseth:
This is still not too stable. I fear that there is no way around mocking the compiler. One way to do that is to provide a mock file `soljson.js` in the local repository at the time the tests are run and then use the url `http://127.0.0.1:8080/#version=builtin` for the tests.

In order for this to run smoothly, the tests could provide a list of anything they will compile and then travis compiles these sources before the UI tests are run and puts these compilations results inside the mock soljson.js.

ci/browser_tests.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-
-SAUCECONNECT_URL="https://saucelabs.com/downloads/sc-4.3.16-linux.tar.gz"
+SC_VERSION="4.4.0"
```
axic: Can you split out the changes to `browser_tests.sh` into another PR and merge that?

nightwatch.js
```diff
@@ -68,7 +68,7 @@ module.exports = {
 
     'local': {
       'launch_url': 'http://localhost:8080',
-      'selenium_port': 4444,
+      'selenium_port': 4445,
```
chriseth: Weird - but it seems to work, perhaps that option is ignored.

ci/browser_tests.sh
chriseth: Will do, the actual tests will probably need more change as mentioned above.

axic:
Is this obsoleted by #273?

axic:
Rebased.

chriseth:
I will update this once https://github.com/ethereum/browser-solidity/pull/283 is merged.

test-browser/compiling.js
```diff
@@ -0,0 +1,35 @@
+module.exports = {
+  'Compile Simple Contract': function (browser) {
+    browser
+      .url('http://127.0.0.1:8080')
```
axic: Should this load a specific version? Like 0.4.0?

yann300:
this PR also fix an ie/safari bug
throw "attempted to assign to readonly property exception" here:
https://github.com/ethereum/browser-solidity/pull/250/files#diff-c2ddeee397dd6378fa147d13ec494e31R830

src/universal-dapp.js
```diff
@@ -826,8 +827,11 @@ UniversalDApp.prototype.rawRunTx = function (args, cb) {
         ++self.blockNumber
       }
       self.vm.runTx({block: block, tx: tx, skipBalance: true, skipNonce: true}, function (err, result) {
-        result.transactionHash = self.txdebugger.web3().releaseCurrentHash() // used to keep track of the transaction
```
axic: I sitll don't get where is `result` readonly? Safari just takes any input parameter as readonly?

axic: Also the web3 mode must be updated with this change.

test-browser/tests/new-file-test.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     browser
       .waitForElementVisible('.newFile', 10000)
       .click('.newFile')
-      .pause('10000')
+      .pause(500)
```
axic: Is the new value better / enough?

test-browser/tests/smoke-test.js
```diff
@@ -10,7 +10,6 @@ module.exports = {
   'Smoke test': function (browser) {
     browser
       .waitForElementVisible('#righthand-panel', 10000)
-      .pause('10000')
```
axic: Is this pause not required anymore?

axic:
Moved the readonly fix to #305.

test-browser/tests/smoke-test.js
yann300: no, we wait just before

Issue #24, opened 2016-05-21
Add option to change block-level parameters in VM mode


axic:
Follow up form https://github.com/chriseth/browser-solidity/issues/90. The following block-level parameters could changed for each transaction and some of them could be configurable by the user:
- ~~blockhash~~ it is different
- coinbase
- ~~block number~~ #269 added this
- difficulty
- block gas limit

PR #269, opened 2016-10-19
Implement steadily increasing `block.number`.

lapo-luchini:
This allows testing contracts such as <https://chainy.info/>.
Known bug: it increments even when reading a `constant` function.
Partially implements #24.

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -49,6 +49,7 @@ UniversalDApp.prototype.reset = function (contracts, getAddress, getValue, getGa
     this._addAccount('dae9801649ba2d95a21e688b56f77905e5667c44ce868ec83f82e838712a2c7a');
     this._addAccount('d74aa6d18aa79a05f3473dd030a97d3305737cbc8337d940344345c1f6b72eea');
     this._addAccount('71975fbf7fe448e004ac7ae54cad0a383c3906055a65468714156a07385e96ce');
+    this.blockNumber = 1150000; // The VM is running in Homestead mode, which started at this block.
   }
 };
 
@@ -749,12 +750,16 @@ UniversalDApp.prototype.runTx = function (args, cb) {
       tx.sign(account.privateKey);
       var block = new EthJSBlock({
         header: {
-          // FIXME: support coinbase, difficulty, number and gasLimit
-          timestamp: new Date().getTime() / 1000 | 0
+          // FIXME: support coinbase, difficulty and gasLimit
+          timestamp: new Date().getTime() / 1000 | 0,
+          number: self.blockNumber
         },
         transactions: [],
         uncleHeaders: []
       });
+      if (!args.useCall) {
+        ++self.blockNumber;
+      }
       self.vm.runTx({block: block, tx: tx, skipBalance: true, skipNonce: true}, function (err, result) {
         result.transactionHash = self.txdebugger.web3().releaseCurrentHash(); // used to keep track of the transaction
         cb(err, result);


axic:
Thanks @lapo-luchini!

Can you merge the two commits and perhaps remove the link the comment?

The state-changing aspect is a current problem we need to address (#270), however you could add a check to not increment the block is `runCall` is specified.

src/universal-dapp.js
```diff
@@ -49,6 +49,7 @@ UniversalDApp.prototype.reset = function (contracts, getAddress, getValue, getGa
     this._addAccount('dae9801649ba2d95a21e688b56f77905e5667c44ce868ec83f82e838712a2c7a');
     this._addAccount('d74aa6d18aa79a05f3473dd030a97d3305737cbc8337d940344345c1f6b72eea');
     this._addAccount('71975fbf7fe448e004ac7ae54cad0a383c3906055a65468714156a07385e96ce');
+    this.block = 1000000;
```
axic: Please rename `block` to `blockNumber`.

lapo-luchini:
Sure… but I'm not sure how to «merge the commits» except producing a new fork a re-creating the PR.

axic:
@lapo-luchini you can use `git rebase` to do that. Try `git rebase -i HEAD~3`.

src/universal-dapp.js
```diff
@@ -749,8 +750,9 @@ UniversalDApp.prototype.runTx = function (args, cb) {
       tx.sign(account.privateKey);
       var block = new EthJSBlock({
         header: {
-          // FIXME: support coinbase, difficulty, number and gasLimit
-          timestamp: new Date().getTime() / 1000 | 0
+          // FIXME: support coinbase, difficulty and gasLimit
+          timestamp: new Date().getTime() / 1000 | 0,
+          number: self.blockNumber++,
```
axic: Don't increment if `runCall` is specified.

lapo-luchini: I don't know about `runCall`, but I inspected a `args.useCall` that is true on constants and false on transactions, I guess that's that.

lapo-luchini:
Ah, I needed to force push, but seems to have worked.

…sorry, I can't wrap my brain around `git` probably because I usually use (either SVN or) `mtn`, which is the project that inspired Linus writing `git`… but it's quite different in approach. (also, `rebase` would be utterly impossible)

lapo-luchini:
(re-forced as I did use 4-spaces indentation instead of your 2)

src/universal-dapp.js
```diff
@@ -49,6 +49,7 @@ UniversalDApp.prototype.reset = function (contracts, getAddress, getValue, getGa
     this._addAccount('dae9801649ba2d95a21e688b56f77905e5667c44ce868ec83f82e838712a2c7a');
     this._addAccount('d74aa6d18aa79a05f3473dd030a97d3305737cbc8337d940344345c1f6b72eea');
     this._addAccount('71975fbf7fe448e004ac7ae54cad0a383c3906055a65468714156a07385e96ce');
+    this.blockNumber = 1000000;
```
axic: Actually we should start from 1,150,000, because the VM is set to Homestead.

axic: Actually we should start from 1,150,000, because the VM is set to Homestead.

lapo-luchini:
OK, changed the value and the commit message with the actual reason instead of the example.

axic:
@lapo-luchini thank your for the work, sorry for not mentioning this last bit in the previous comment.

chriseth:
@axic shall I merge or do you still want this comment to be included?

axic:
@chriseth I would like to get the comment before we forget about it (i.e. we need to update the block number and the comment when browser-solidity moves to EIP150)

chriseth:
@axic then please change your review.

lapo-luchini:
Should be ok now.

axic:
@lapo-luchini thank you. We'll apply it as soon as the test issues are sorted.

axic:
@lapo-luchini can you please rebase against latest master? It has the testing suite fixed. After the rebase the tests here should pass and we can merge it.

lapo-luchini:
@axic rebased, but tests still fail. (seems to be some `semistandard`)

src/universal-dapp.js
```diff
@@ -749,12 +750,15 @@ UniversalDApp.prototype.runTx = function (args, cb) {
       tx.sign(account.privateKey);
       var block = new EthJSBlock({
         header: {
-          // FIXME: support coinbase, difficulty, number and gasLimit
-          timestamp: new Date().getTime() / 1000 | 0
+          // FIXME: support coinbase, difficulty and gasLimit
+          timestamp: new Date().getTime() / 1000 | 0,
+          number: self.blockNumber,
```
axic: Please remove the comma. (semistandard)

lapo-luchini:
Tests are now green, yay!

axic:
Thanks, merged!

PR #273, opened 2016-10-20
use mock compiler during browser tests

yann300:


ci/makeMockCompiler.js
```diff
@@ -0,0 +1,65 @@
...
+    key: key.trim(),
+    source: source,
+    optimization: optimization,
+    missingInputs: missingInputs,
+    result: result
+  }
+}
+
+function replaceSolCompiler () {
+  fs.readFile('./test-browser/mockcompiler/compiler.js', 'utf8', function (error, data) {
```
chriseth: You could also include the compiler as a literal string here and combine it with the compilation results, i.e. write everything directly into soljson.js.

test-browser/mockcompiler/.gitignore
```diff
@@ -0,0 +1 @@
+results.json
```
chriseth: Why not use the global gitignore file?

test-browser/mockcompiler/compiler.js
```diff
@@ -0,0 +1,58 @@
...
+    _compileJSONCallback: {},
+    _compileJSON: {}
+  };
+
+  loadJSON('./test-browser/mockcompiler/results.json', function (data) {
+    window.Module.mockData = JSON.parse(data);
+  });
+}
+
+function loadJSON (url, callback) {
```
chriseth: The suggestion above would also not require such a request.

test-browser/simpleContract.js
```diff
@@ -0,0 +1,65 @@
...
+};
+
+function runTests (browser, testData) {
+  browser
+    .url('http://127.0.0.1:8080/#version=builtin')
+    .waitForElementVisible('.newFile', 10000)
+  browser.assert.notEqual(testData, null)
+  /*
+  TODO push contracts to the editor and test
+  setEditorContent(browser, testData.testSimpleContract.sources.Untitled, function () {
```
chriseth: In another test I used
```
.clearValue('#input textarea')
.setValue('#input textarea', source)
```

test-browser/simpleContract.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+var fs = require('fs');
+
+module.exports = {
+  'Simple Contract': function (browser) {
```
chriseth: Isn't it better to use nodejs require for that?
Another solution (perhaps for the next iteration) would be to declare sources inside each test with a special delimiter and scan all test files when running `makeMockCompiler`.

test-browser/simpleContract.js
```diff
@@ -0,0 +1,65 @@
...
+
+function runTests (browser, testData) {
+  browser
+    .url('http://127.0.0.1:8080/#version=builtin')
+    .waitForElementVisible('.newFile', 10000)
+  browser.assert.notEqual(testData, null)
+  /*
+  TODO push contracts to the editor and test
+  setEditorContent(browser, testData.testSimpleContract.sources.Untitled, function () {
+    getCompiledContracts(browser, function (contracts) {
```
chriseth: I think you need some kind of wait loop here. I'm not sure if my solution in `compiling.js` works, but you could take a look.

axic:
@yann300 this is a great idea, I reckon it will speed up (and fix) the tests once and for all :+1:

test-browser/simpleContract.js
yann300: this line is just about declaring the test.

test-browser/simpleContract.js
yann300: could point me to the line please? i can't find it

ci/makeMockCompiler.js
```diff
@@ -0,0 +1,57 @@
...
+    result: result
+  };
+}
+
+function replaceSolCompiler (results) {
+  fs.readFile('./test-browser/mockcompiler/compiler.js', 'utf8', function (error, data) {
+    if (error) {
+      return console.log(error);
+    }
+    data = data + '\n\nvar mockData = \'' + JSON.stringify(results).replace(/(\\t|\\n|'|\\")/g, ' ') + '\'';
```
chriseth: What is the purpose of the replacement? I think this modifies the compilation results. Why not just `var mockData = ` + JSON.stringify(results);` (without the quotes) - JSON is valid javascript.

ci/makeMockCompiler.js
```diff
@@ -0,0 +1,57 @@
...
+    fs.writeFile('./soljson.js', data, 'utf8', function (error) {
+      if (error) {
+        return console.log(error);
```
chriseth: What is vital here: Make the script return a failure exit code so that the test will stop.

src/app.js
```diff
@@ -554,6 +554,13 @@ var run = function () {
   });
 
   $.getJSON('https://ethereum.github.io/solc-bin/bin/list.json').done(function (data) {
+    function buildVersion (build) {
```
chriseth: Why include this change here?

test-browser/simpleContract.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+var fs = require('fs');
+
+module.exports = {
+  'Simple Contract': function (browser) {
+    initTestData(function (error, testData) {
+      if (error) {
+        console.log(error);
```
chriseth: Again, this should result in a failure.

src/app.js
yann300: that function has been removed from a previous commit. but it is used

axic: Oh, true! It is used in the option list. **Damn tests**, we need them so badly.

axic: Moved this to #275.

chriseth:
This does not yet call the compiler.

chriseth:
@yann300 ok, let's merge this even if it does not have a test yet. Did you get the ballot working? If not, it's fine for now.

src/app.js
```diff
@@ -520,15 +520,7 @@ var run = function () {
     } else {
       url = 'https://ethereum.github.io/solc-bin/bin/' + version;
     }
-    var isFirefox = typeof InstallTrigger !== 'undefined';
```
axic: Why is this being removed? Is the URL fix in #277 solving this problem?

yann300: I just want to check if running without workers fix all my tests in firefox. #277 seems to not fix the issue

axic: Nothing seems to fix that issue and I cannot even reproduce it :(

axic: @yann300 true means using worker, so with removing it in every case it is using in the browser thread.

yann300: which firefox version are you running?
so https://ethereum.github.io/browser-solidity/#version=builtin
is working for you?

axic: No, it isn't: `Worker error: undefined`.

But the other worker issue I cannot reproduce.

yann300: I was only referring to this issue `Worker error: undefined`.
@chriseth the build is ok if we don't use the worker for firefox. 
the Ballot is included but not yet tested (see the comment in the code).
We could merge this now and fix the worker in another pr.

chriseth: @axic @yann300 let's first discuss and fix the worker issue in #277

chriseth:
Rebased to test the worker.

PR #277, opened 2016-10-24
Use absolute url for worker.

chriseth:


src/app/compiler.js
```diff
@@ -201,6 +201,13 @@ function Compiler (editor, handleGithubCall) {
       jobs.push({sources: source});
       worker.postMessage({cmd: 'compile', job: jobs.length - 1, source: JSON.stringify(source), optimize: optimize});
     };
+    if (!url.startsWith('http') && !url.startsWith('file')) {
```
axic: Since we're limiting it for the `https` case and using a local file, can you add a comment explaining the same (and reasons, i.e. https doesn't allow that)?

chriseth: I'm limiting to the case where we probably have a relative path.

src/app.js
```diff
@@ -516,7 +516,13 @@ var run = function () {
     queryParams.update({ version: version });
     var url;
     if (version === 'builtin') {
-      url = 'soljson.js';
+      var location = window.document.location;
+      location = location.protocol + '//' + location.host + '/' + location.pathname;
```
chfast: Why not `location.href`?

chriseth: Because it includes the hash part (`#version=...')

axic: The case we're dealing with, the href is `http://ethereum.github.io/browser-solidity/#version=builtin` and we want an end result of `http://ethereum.github.io/browser-solidity/soljson.js`.

PR #287, opened 2016-10-27
Split runTx (and make settings async)

axic:
This became a monster change set. The core idea is to make the setting inputs asynchronous, so that eventually it can be event driven and/or we can prompt the user which account to send a transaction from.

Fixes two theoretical bugs:
- invalid account is selected in VM mode (not really possible currently as the dropdown is populated by us)
- if personal mode is used in web3, it would still try to use the non-personal account

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "test": "node test/index.js"
   },
   "devDependencies": {
+    "async": "^2.1.2",
     "babel-cli": "^6.16.0",
     "babel-plugin-transform-es2015-block-scoping": "^6.15.0",
     "babel-plugin-transform-es2015-template-literals": "^6.8.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -63,5 +63,8 @@
       "src/mode-solidity.js",
       "soljson.js"
     ]
+  },
+  "dependencies": {
+    "async": "^2.1.2"
```
yann300: that could be a dev dependency no?

src/universal-dapp.js
```diff
@@ -679,34 +680,94 @@ function tryTillResponse (web3, txhash, done) {
...
-  if (self.getGasLimit) {
-    try {
-      gasLimit = self.getGasLimit()
-    } catch (e) {
-      return cb(e)
-    }
-  }
+  async.waterfall([
+    // query gas limit
+    function (callback) {
```
yann300: Is that possible to move these functions (also getGasLimit maybe) outside of universal-dapp.js?
cause if we add more functionality here, the whole universal-dapp.js will become not so readable.

package.json
axic: Fixed

PR #292, opened 2016-10-28
Save file contents periodically

### LINKED ISSUE #265: Files are not saved to storage with auto compile disabled
It seems that files are saved to storage during compile. If "auto compile" is disabled files are only saved when the "compile" button is pressed.

A further side affect of this is that when importing files the saved version of the file is imported, rather than the live version. This leads to much confusion. It happens a lot because files to be imported are not typically compiled on their own and therefore are not saved.

axic:
And don't rely on the auto-compile feature to do so.

Fixes #265.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -455,10 +455,13 @@ var run = function () {
 
   function editorOnChange () {
     var input = editor.getValue()
+    // save contents
+    editor.setCacheFileContent(input)
+
     if (input === '') {
-      editor.setCacheFileContent('')
       return
     }
+
     if (input === previousInput) {
       return
     }
--- a/src/app.js
+++ b/src/app.js
@@ -522,6 +522,10 @@ var run = function () {
     }
   })
 
+  compiler.event.register('compilationStarted', this, function () {
+    editor.clearAnnotations()
+  })
+
   function startdebugging (txHash) {
     transactionDebugger.debug(txHash)
     selectTab($('ul#options li.debugView'))
--- a/src/app.js
+++ b/src/app.js
@@ -452,20 +452,31 @@ var run = function () {
 
   var previousInput = ''
   var compileTimeout = null
+  var saveTimeout = null
 
   function editorOnChange () {
     var input = editor.getValue()
-    // save contents
-    editor.setCacheFileContent(input)
 
-    if (input === '') {
-      return
-    }
-
+    // if there's no change, don't do anything
     if (input === previousInput) {
       return
     }
     previousInput = input
+
+    // fire storage update
+    // NOTE: save at most once per 5 seconds
+    if (saveTimeout) {
+      window.clearTimeout(saveTimeout)
+    }
+    saveTimeout = window.setTimeout(function () {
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -455,10 +455,13 @@ var run = function () {
 
   function editorOnChange () {
     var input = editor.getValue()
+    // save contents
+    editor.setCacheFileContent(input)
```
yann300: The content will be saved each time the content is modified ?
I would perhaps just add a save button next to the compile button, and save the content automatically in a *.save.sol file every X minutes. So users could retrieve the content if something goes wrong.
Cause i guess saving a 500 lines contract every time will slowdown everything.

ethernomad: If saving every time the file is modified is too slow it could save on tab close. Also, all files would need to be saved when any file is compiled because of imports.

chriseth: I would not add a save button. Save buttons always complicate things because you have to keep track of two different states of your data. In our case, this is even more complicated because the files are not really stored in files. People might expect the data turn up on their harddrive. We already have unlimited undo history. I think it would make sense to add an 'autocommit to git' in the future, but the localstorage should always reflect the current state.

Concerning this change itself: I think we had a save every 5 minutes. It might be worth checking whether hold-pressing a button is slower without such a delay.

axic: There was a 300ms timeout in the previous one, that translates to 0.3 seconds.

We could make this proper and say, save every 5 seconds. Question is: should we save before every compilation too?

chriseth: Sorry, that was a typo, I did not mean 5 minutes.
I think we should save before each compilation and roughly once per second, depending on performance.

src/app.js
```diff
@@ -452,18 +452,32 @@ var run = function () {
...
       return
     }
     previousInput = input
 
+    // fire storage update
+    // NOTE: save at most once per 5 seconds
+    if (saveTimeout) {
+      window.clearTimeout(saveTimeout)
+    }
+    saveTimeout = window.saveTimeout(function () {
```
yann300: is that really in `window` ?

axic: It is used that way in the code below. I've just copied it.

yann300: I see, cause i never saw `window.saveTimeout`
it should be `window.setTimeout` I think
And also we should retrigger the timer each time. setTimeout does not act as a loop.. but maybe this is already done elsewhere.

src/app.js
```diff
@@ -452,18 +452,32 @@ var run = function () {
...
       return
     }
     previousInput = input
 
+    // fire storage update
+    // NOTE: save at most once per 5 seconds
+    if (saveTimeout) {
+      window.clearTimeout(saveTimeout)
+    }
+    saveTimeout = window.setTimeout(function () {
```
axic: In order to use `setInterval` we need to cancel it every time we switch tabs.  I think we should redesign the editor class and the events to make this more effective.

I would also like to remove the editor dependency in the compiler.

PR #298, opened 2016-11-03
static analysis framework

yann300:
story https://github.com/ethereum/browser-solidity/issues/252

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -613,7 +613,7 @@ var run = function () {
       selectedVersion = queryParams.get().version
     }
 
-    loadVersion(selectedVersion)    
+    loadVersion(selectedVersion)
   }).fail(function (xhr, text, err) {
     // loading failed for some reason, fall back to local compiler
     $('#versionSelector').append(new Option('latest local version', 'builtin'))
--- a/src/app.js
+++ b/src/app.js
@@ -19,6 +19,7 @@ var UniversalDApp = require('./universal-dapp.js')
 var Debugger = require('./app/debugger')
 var FormalVerification = require('./app/formalVerification')
 var EventManager = require('./lib/eventManager')
+var StaticAnalysis = require('./app/staticanalysis/staticAnalysisView')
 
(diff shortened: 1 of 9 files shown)


src/app.js
```diff
@@ -610,6 +611,10 @@ var run = function () {
     }
 
     loadVersion(selectedVersion)
+
+    // static analysis
+    var staticanalysis = new StaticAnalysis(compiler, renderer)
```
chriseth: Why do you add it here? It should be rather similar to the "Formal Verification" tab, which is added right at the beginning.

src/app/staticanalysis/staticAnalysisView.js
```diff
@@ -0,0 +1,79 @@
...
+  return toRun
+}
+
+staticAnalysisView.prototype.run = function () {
+  if (!this.view) {
+    return
+  }
+  var selected = this.selectedModules()
+  var warningContainer = $('#staticanalysisresult')
+  warningContainer.html('')
```
chriseth: `.empty()`?

src/app/staticanalysis/staticAnalysisRunner.js
```diff
@@ -0,0 +1,30 @@
...
+var AstWalker = require('ethereum-remix').util.AstWalker
+var list = require('./modules/list')
+
+function staticAnalysisRunner () {
+}
+
+staticAnalysisRunner.prototype.run = function (ast, toRun, callback) {
+  var walker = new AstWalker()
+  for (var k in ast) {
+    walker.walk(ast[k].AST, {'*': function (node) {
```
chriseth: I think we could improve that in the future. Furthermore, there might be some things to check which need the AST as is, and not just as a callback on each node.

yann300:
TODO 
 - highlight static analysis tab is warning - trigger a run for each compilation)
 - error for each source location
 - nothing if no warning
 - add test

PR #301, opened 2016-11-04
Cleanup the compiler

### LINKED ISSUE #295: Refactor: remove editor from compiler

axic:
Fixes #295.

### DIFF
--- a/src/app/compiler.js
+++ b/src/app/compiler.js
@@ -39,12 +39,8 @@ function Compiler (editor, handleImportCall) {
     })
   }
 
-  var compile = function () {
+  var compile = function (files) {
     self.event.trigger('compilationStarted', [])
-    var input = editor.getValue()
-
-    var files = {}
-    files[utils.fileNameFromKey(editor.getCacheFile())] = input
     internalCompile(files)
   }
   this.compile = compile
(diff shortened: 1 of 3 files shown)


axic:
This is work in progress, moving of the `editor.getFile` must be done also.

src/app/compiler.js
```diff
@@ -121,7 +117,7 @@ function Compiler (editor, handleGithubCall) {
       self.event.trigger('compilationFinished', [false, data, source])
     } else if (missingInputs !== undefined && missingInputs.length > 0) {
       // try compiling again with the new set of inputs
-      internalCompile(source.sources, missingInputs)
+      internalCompile(source.sources, source.target, missingInputs)
```
axic: This is the part I didn't finish. `source.target` must be added the compilation phase (it must be passed back).

Alternatively, we could wrap the whole compilation into a context (compiler instance) and have these there and not passed back and forth.

axic:
Will adjust this after #354.

axic:
After merging the changes from #354, the `editor` dependency can be removed here and job's done (tm).

axic:
I'm not sure why this doesn't work yet with the tests, it does work locally with a real compiler. @yann300 could it because of the mock compiler?

axic:
The `setCompileJSON` part should be removed, it is only used by `test/compiler-test.js`, which instead should just use the mock compiler.

axic:
@chriseth @yann300 it works now, review please

PR #303, opened 2016-11-05
Update to ethereumjs-vm 2.0.1

axic:


axic:
I don't get why this breaks the tests, a local build works for me.

chriseth:
Tests succeed for firefox but fail for the other browsers.

src/app/execution-context.js
```diff
@@ -16,7 +16,10 @@ if (typeof window.web3 !== 'undefined') {
   web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:8545'))
 }
 
-var vm = new EthJSVM(null, null, { activatePrecompiles: true, enableHomestead: true })
```
yann300: Why did you removed the first two parameters? I might be one the reasons why tests are failing

axic: That is the breaking change of the 2.x version.

PR #304, opened 2016-11-05
Support useCall in VM mode

### LINKED ISSUE #241: Support calls in runTx in VM mode
Needed for #127.

axic:
Fixes #241.


yann300:
I tried it that's fine

chriseth:
Will this break if there is a tx call while the call call is running? Can we create a copy of the state for the call instead?

chriseth:
As discussed, this runs in the ui thread anyway, so probably not a big issue.

PR #290, opened 2016-10-27
Added LESS to compile CSS

jcbrmn:
Was able to add in less and a clean css plugin that minifies all of the css. Color variables are at the top of each less file. I just left them for now, but you can easily split those into a different file and in your styles.less file just make sure you call the variables file before the other less files.

Let me know if you have any questions.

PR #312, opened 2016-11-08
Use LESS instead of CSS

### LINKED ISSUE #203: Use LESS or another CSS build tool
Currently, CSS files are places in `assets/css/`. It would be nice moving `browser-solidity.css`, `styles.css`, `universal-dapp.css` to LESS and including it in the build process.

axic:
Replaces #290. Fixes #203.

Changes needed:
- [x] add spaces between definitions
- [x] fix errors introduced (duplicate properties)
- [x] ~~fix package.json? it works on my computer, but not on travis~~ (have to run build before csslint)
- [ ] verify the output hasn't changed
- [ ] merge in changes from #340
- [ ] merge in changes from #352


axic:
Rebased.

axic:
Errors:
```
app.css
9: error at line 221, col 5
Duplicate property 'position' found.
    position: static;

app.css
13: error at line 340, col 3
Duplicate property 'width' found.
  width: 33%;

app.css
14: error at line 343, col 3
Duplicate property 'padding' found.
  padding: 0.25em;

app.css
17: error at line 352, col 3
Duplicate property 'width' found.
  width: 33%;

app.css
18: error at line 355, col 3
Duplicate property 'padding' found.
  padding: 0.25em;
```

axic:
We need to review the "Removesduplicates from less" commit. Are the second definitions in CSS overriding the previous ones?

axic:
All the colors are mixed up, because the global color declarations get overwritten. (See the top of each .less file).

serapath:
Is this set in stone?

I think the biggest potential problem in a big project like an IDE is, that many different visual components having their CSS in a global CSS file and different developers adding styles means it's easy to grow the CSS into a lot of bloat over time.

It needs a lot of discipline to not mess up CSS and it's getting harder and harder to reason about it - especially layout related bugs.

The main reason behind is the "cascading" (which means, all css selectors pollute the global namespace - and that's as bad as using tons of global variables in programming languages)

A solution would be to have **scoped css**, which is solved by the BEM convetion from yandex, but still a pain in the ass to manually follow through and really always do all the namespacing everywhere.

A step up from that and state of the art is using `css-in-js` solutions (https://github.com/MicheleBertoli/css-in-js)

My current favourite one is `csjs` which i'm using on the `file-explorer` branch.
It still allows to use css, but it gives special powers better than less and it still allows designers and coders to re-use their css knowledge, because it's css in js :-)

Now, even the simplest css class names become local variables automatically not breaking any global styles anymore and they are now co-located with the component they are meant for.

When I open a component - may it be a "search bar" or a "button" or an "image gallery" - the only thing i personally want is to find all the necessary stuff (like javascript, css and html) related to that component in one place with no need to look up tons of different places in a bloated directory structure scrolling to tons of unrelated stuff so that i even need to utilize editors search capabilities and i'm still not sure if there is not some other side effect that messes with my stuff :-)


I can highly recommend using local scope for css through javascript which automatically gives even better powers than pre-processors like sass, stylus or less

chriseth:
csjs looks like a good thing. I actually like the fact that styling and code are in the same file. Optimally, you should of course separate layout and functionality, but I don't think it is possible in reality - starting from the fact that you have to order divs in the code depending on the final layout. This separation should not be done between html and css but between UI and non-UI code.

axic:
I guess this can be closed since most of the CSS was moved to csjs.

PR #316, opened 2016-11-09
static analysis: Highlight static analysis tab if warning and trigger a run for each compilation

yann300:
MERGE AFTER https://github.com/ethereum/browser-solidity/pull/311/


src/app.js
```diff
@@ -528,6 +554,8 @@ var run = function () {
 
   compiler.event.register('compilationStarted', this, function () {
     editor.clearAnnotations()
+    var input = editor.getValue()
+    editor.setCacheFileContent(input)
```
chriseth: all this should already be done in app.js (also clearannotations)

src/app/staticanalysis/modules/txOrigin.js
```diff
@@ -17,9 +17,12 @@ txOrigin.prototype.visit = function (node) {
 }
 
 txOrigin.prototype.report = function (node) {
-  var report = this.txOriginNode.length + ' use of tx.origin\n'
+  var report = []
   this.txOriginNode.map(function (item, i) {
-    report += item.src + '\n'
+    report.push({
+      warning: 'use of tx.origin: "tx.origin" is useful only in very exceptional cases.\nIf you use it for authentication, you usually want to replace it by "msg.sender", because otherwise any contract you call can act on your behalf.',
```
chriseth: Please also break this line in the source code.

PR #330, opened 2016-11-14
Add gas checker.

chriseth:
Closes https://github.com/ethereum/browser-solidity/issues/162

This includes a refactoring of the error reporting mechanism (add options as an object), adds a feature to it (use "span" instead of "pre") and refactors the static analyzer in response to different requirements, as expected.

### DIFF
--- a/src/app/renderer.js
+++ b/src/app/renderer.js
@@ -40,7 +40,12 @@ Renderer.prototype.error = function (message, container, options) {
   if (!opt.type) {
     opt.type = utils.errortype(message)
   }
-  var $pre = $(opt.useSpan ? '<span />' : '<pre />').text(message)
+  var $pre
+  if (opt.isHTML) {
+    $pre = $(opt.useSpan ? '<span />' : '<pre />').html(message)
+  } else {
+    $pre = $(opt.useSpan ? '<span />' : '<pre />').text(message)
+  }
   var $error = $('<div class="sol ' + opt.type + '"><div class="close"><i class="fa fa-close"></i></div></div>').prepend($pre)
   if (container === undefined) {
     container = $('#output')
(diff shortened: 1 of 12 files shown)


axic:
This has a lot of non-gaschecker code and is missing the actual gaschecker file.

chriseth:
@axic thanks, added the gas checker.

src/app/renderer.js
```diff
@@ -34,13 +34,14 @@ function Renderer (editor, updateFiles, udapp, executionContext, formalVerificat
...
   var self = this
-  if (!type) {
-    type = utils.errortype(message)
+  var opt = options || {}
+  if (!opt.type) {
+    opt.type = utils.errortype(message)
   }
-  var $pre = $('<pre />').text(message)
-  var $error = $('<div class="sol ' + type + '"><div class="close"><i class="fa fa-close"></i></div></div>').prepend($pre)
+  var $pre = $(opt.useSpan ? '<span />' : '<pre />').text(message)
```
yann300: why using span instead of pre ?

chriseth: The pre was mostly needed to format the pointer to the source location properly, which is generated by the compiler. If we use `span` here, the browser breaks the lines and it is much easier to read.

yann300: cause if you put `\n` in it and use a span. it will not add a new line

chriseth: Yes, I noticed that, too. But then you do want the browser to break. Perhaps we can go with html and use `<br/>` for newlines?

yann300: yes

test-browser/tests/staticanalysis.js
```diff
@@ -28,33 +29,22 @@ module.exports = {
...
-      })
-    })
-  browser.assert.equal(found, true)
-}
-
 function runTests (browser) {
   browser
     .waitForElementVisible('.newFile', 10000)
   contractHelper.testContracts(browser, sources.sources.Untitled, ['TooMuchGas', 'test1', 'test2'], function () {
+    console.log('pppzzz?')
```
chriseth: Please remove.

Issue #223, opened 2016-09-07
Support the new ABI keywords


axic:
```
[
  {
    "constant": false,
    "inputs": [],
    "name": "rsaverify",
    "output": [
      {
        "name": "",
        "type": "bool"
      }
    ],
    "payable": false,
    "type": "function"
  },
  {
    "inputs": [],
    "type": "constructor"
  },
  {
    "payable": false,
    "type": "fallback"
  }
]
```

UDapp only supports `function` and doesn't care about the new keywords.

Todo:
- [x] support the `constructor` keyword (right now it is created manually)
- [x] support the `fallback` keyword (right now it is created manually)
- [x] support the `payable` keyword and introduce an icon / colour for payable calls
- [x] register for the `compilerLoaded` event, because the current behaviour must be enabled for old compiler versions

axic:
It may make more sense adding a filtering layer to the compiler: for "old" compiler versions, the interface would be extended with:
- a `constructor` entry
- a `fallback` entry
- `payable` keyword on every `constructor`, `fallback`, `function` entry

axic:
Fixed by #331.

PR #331, opened 2016-11-14
Properly handle the interface ABI

axic:
See #223.

Depends on https://github.com/ethereum/solc-js/pull/77.

### DIFF
--- a/src/app/compiler.js
+++ b/src/app/compiler.js
@@ -2,6 +2,7 @@
 
 var solc = require('solc/wrapper')
 
+var semver = require('semver')
 var webworkify = require('webworkify')
 var utils = require('./utils')
 
@@ -19,6 +20,8 @@ function Compiler (editor, handleGithubCall) {
 
   var cachedRemoteFiles = {}
   var worker = null
+
+  var currentVersion
 
   var optimize = false
 
@@ -53,6 +56,7 @@ function Compiler (editor, handleGithubCall) {
   this.setCompileJSON = setCompileJSON // this is exposed for testing
 
   function onCompilerLoaded (version) {
+    currentVersion = version
     self.event.trigger('compilerLoaded', [version])
   }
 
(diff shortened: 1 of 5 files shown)


axic:
This is not a final version. I think the JSON manipulation should be in solc-js.

src/app/compiler.js
```diff
@@ -270,6 +276,58 @@ function Compiler (editor, handleGithubCall) {
...
+
+        // add "payable" to everything
+        // FIXME: only for compiler <0.4.0
+        if (semver.lt(currentVersion, '0.4.0')) {
+          item.payable = true
+        }
+      }
+
+      // FIXME: only for compiler <0.x.y
+      //        0.1.2 from Aug 2015 had it. The code has it since May 2015 (e7931ade)
```
axic: @chriseth do you remember which released version it was added in?

chriseth: I think we started versioning in July 2015.

axic:
Build needs restart once https://github.com/ethereum/solc-js/pull/77 is merged.

axic:
This needs a bit more work.

axic:
Works now.

src/universal-dapp.js
```diff
@@ -598,10 +618,19 @@ UniversalDApp.prototype.getCallButton = function (args) {
     })
   }
 
+  var title
+  if (isConstructor) {
+    title = 'Create'
```
yann300: Does it make sense to name it (Create Contract), (Deploy Contract) or (Create Instance) ..?
so it does not overlap with a potential `Create` function

axic: That is only shown on the top for `Create` and never mixed with actual functions. Note that the behaviour of that is unchanged.

axic:
This works locally on my browser but fails on the karma. @yann300 do you have any idea?

axic:
Thank you @yann300!

PR #335, opened 2016-11-21
Split tx run + Queue txs

yann300:
fix https://github.com/ethereum/browser-solidity/issues/248

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -3,13 +3,12 @@
 
 var $ = require('jquery')
 var ethJSUtil = require('ethereumjs-util')
-var EthJSTX = require('ethereumjs-tx')
 var ethJSABI = require('ethereumjs-abi')
-var EthJSBlock = require('ethereumjs-block')
 var BN = ethJSUtil.BN
 var EventManager = require('./lib/eventManager')
 var crypto = require('crypto')
 var async = require('async')
+var TxRunner = require('./app/txRunner')
 
 /*
   trigger debugRequested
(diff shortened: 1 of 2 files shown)


yann300:
fix https://github.com/ethereum/remix/issues/144

src/universal-dapp.js
```diff
@@ -34,10 +33,18 @@ function UniversalDApp (executionContext, options, txdebugger) {
   self.executionContext.event.register('contextChanged', this, function (context) {
     self.reset(self.contracts)
   })
+  self.txRunner = new TxRunner(executionContext, {
+    queueTxs: true,
+    personalMode: true
```
axic: I think this will break because we've never tested personalMode.

src/universal-dapp.js
```diff
@@ -733,7 +725,7 @@ UniversalDApp.prototype.runTx = function (args, cb) {
             return callback(err)
           }
 
-          tx.from = ret
+          tx.from = accounts[ret]
```
axic: `accounts` is specific to VM mode (it contains private keys).

src/universal-dapp.js
```diff
@@ -746,115 +738,19 @@ UniversalDApp.prototype.runTx = function (args, cb) {
             return callback('No accounts available')
           }
 
-          tx.from = ret[0]
+          tx.from = accounts[ret[0]]
```
axic: Same here, this will break web3 mode.

src/universal-dapp.js
yann300: yep :) i have fixed that in another pr

PR #339, opened 2016-11-22
Send compilation result to the debugger

yann300:
+ use only compiler compilation  result property


yann300:
merge after https://github.com/ethereum/remix/pull/145

PR #341, opened 2016-11-23
Fix send tx through web3

### LINKED ISSUE #342: Invalid address error when using injected provider
Over at MetaMask, we had [an issue submitted](https://github.com/MetaMask/metamask-plugin/issues/880) by @drcode regarding MetaMask/browser-solidity compatibility, and I could use your help to track down.

The error thrown is `Uncaught Error: invalid address(…)`, thrown from browser-solidity when the payload has an `undefined` `from` field.

For some reason it seems that browser-solidity is no longer successfully building its transaction parameters using the currently injected selected address at `web3.eth.accounts[0]` or `web3.eth.coinbase`.

In the `Method.prototype.buildCall` function, you can see this line:
```
var payload = method.toPayload(Array.prototype.slice.call(arguments));
```

The input arguments are very minimal, and the resulting `payload` value lacks a defined sender:
```javascript
{from: undefined, to: undefined, data: "0x606060405234610000575b33600060006101000a8154817000816000905550600101610594565b5090565b50505b5056", value: "0x0", gas: 449552}
```

This makes it look like the `method.toPayload()` method is failing to pull out the current address from the injected `web3`.

Unfortunately, when I did [a search of this repository](https://github.com/ethereum/browser-solidity/search?utf8=%E2%9C%93&q=toPayload), I saw no `toPayload` reference anywhere, so now I'm confused.

This *seems to be* a browser-solidity bug to me right now, but without the original source code to search & reference, it will be hard for me to track down.  Any clues?

yann300:
Fixes #342.

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -57,17 +57,6 @@ UniversalDApp.prototype.reset = function (contracts, getAddress, getValue, getGa
     this._addAccount('dae9801649ba2d95a21e688b56f77905e5667c44ce868ec83f82e838712a2c7a')
     this._addAccount('d74aa6d18aa79a05f3473dd030a97d3305737cbc8337d940344345c1f6b72eea')
     this._addAccount('71975fbf7fe448e004ac7ae54cad0a383c3906055a65468714156a07385e96ce')
-  } else {
-    var self = this
-    this.getAccounts(function (error, accounts) {
-      if (error) {
-        console.log('cannot retrieve accounts from web3')
-      } else {
-        accounts.map(function (item, index) {
-          self.accounts[item] = item
-        })
-      }
-    })
   }
 }
 
@@ -695,7 +684,6 @@ UniversalDApp.prototype.runTx = function (args, cb) {
     data: args.data,
     useCall: args.useCall
   }
-  var accounts = this.accounts
   async.waterfall([
     // query gas limit
     function (callback) {
@@ -739,7 +727,8 @@ UniversalDApp.prototype.runTx = function (args, cb) {
             return callback(err)
           }
 
-          tx.from = accounts[ret]
+          tx.from = self.executionContext.isVM() ? self.accounts[ret] : ret
+
           callback()
         })
       } else {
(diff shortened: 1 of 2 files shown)


axic:
This goes against the previous design, where we have "prototype.getAccounts" doing this abstraction. Mixing external RPC addresses into the private key light might not be the best idea.

yann300:
I thought that it is better to use the property `accounts` even in web3 mode than to leave it blank
while in web3 mode (which could loo inconsistent)
the VM and web3 make use of accounts differently (the vm need private key, web3 not), but both are still representing an account.
But I see your concern. I pushed changes

src/universal-dapp.js
```diff
@@ -549,11 +549,14 @@ UniversalDApp.prototype.getCallButton = function (args) {
 
     var decoded
     self.runTx({ to: args.address, data: data, useCall: args.abi.constant && !isConstructor }, function (err, txResult) {
-      var result = txResult.result
+      var result
       if (err) {
         replaceOutput($result, $('<span/>').text(err).addClass('error'))
       // VM only
-      } else if (self.executionContext.isVM() && result.vm.exception === 0 && result.vm.exceptionError) {
```
axic: This seems to be wrong, all the conditions below are error cases.

Maybe it would be better adding an extra condition before all these:
```
if (!err) {
  result = txResult.result
}
```

axic:
@yann300 updated with a change to txResult

kumavis:
tests are failing @yann300

yann300:
yes that's fine now

kumavis:
@axic ready for you

src/universal-dapp.js
```diff
@@ -58,6 +54,10 @@ UniversalDApp.prototype.reset = function (contracts, getAddress, getValue, getGa
     this._addAccount('d74aa6d18aa79a05f3473dd030a97d3305737cbc8337d940344345c1f6b72eea')
     this._addAccount('71975fbf7fe448e004ac7ae54cad0a383c3906055a65468714156a07385e96ce')
   }
+  this.txRunner = new TxRunner(this.executionContext, this.accounts, {
```
axic: Since we have the `opts` field, shouldn't `accounts` (and maybe even `executionContext`) be part of it? I'm just wondering.

axic:
@yann300 it seems you've force pushed over my commit?

axic:
Added back my change to the 'fix txResult' commit.

yann300:
oh ok sorry for that

src/universal-dapp.js
yann300: hmm so i see in opts just a param used to modify the behavior of txrunner.
executionContext and accounts are more kind of required parameters.

axic: Well, we can cleanup the `txRunner` in another PR 😉

PR #345, opened 2016-11-29
Extend state manager (VM)

yann300:


axic:
Is there a way we could extend the class, either by `require('inherit')` or ES6 classes?

src/app/execution-context.js
```diff
@@ -16,10 +18,48 @@ if (typeof window.web3 !== 'undefined') {
...
+  }
+}
+
+var stateManager = new StateManagerCommonStorageDump({})
 var vm = new EthJSVM({
   enableHomestead: true,
   activatePrecompiles: true
 })
+
+vm.stateManager = stateManager
```
axic: Can you add a note here that this should move into the `EthJSVM` constructor?

PR #346, opened 2016-12-02
Display metadata hash if found

axic:


src/app/renderer.js
```diff
@@ -91,6 +91,13 @@ Renderer.prototype.contracts = function (data, source) {
     })
   }
 
+  var retrieveMetadataHash = function (bytecode) {
+    var match = /a165627a7a72305820([0-9a-f]{64})0029$/.exec(bytecode)
```
chriseth: case-insensitive regex to be more flexible?

src/app/renderer.js
```diff
@@ -101,6 +108,12 @@ Renderer.prototype.contracts = function (data, source) {
 
     if (contract.bytecode) {
       $contractOutput.append(uiHelper.preRow('Web3 deploy', uiHelper.gethDeploy(contractName.toLowerCase(), contract['interface'], contract.bytecode), 'deploy'))
+
+      // check if there's a metadata hash appended
+      var metadataHash = retrieveMetadataHash(contract.bytecode)
+      if (metadataHash) {
+        $contractOutput.append(uiHelper.tableRow('Metadata location', 'bzzr://' + metadataHash))
```
chriseth: We could use `'http://swarm-gateways.net/bzzr:/' + metadataHash` until we have support for the bzzr protocol

src/app/renderer.js
axic: I think we should rather just lowercase the bytecode so the output is consistent.

src/app/renderer.js
axic: Since we're not solving the upload process yet I thought just to display the actual metadata and the URL.

PR #353, opened 2016-12-07
Support deterministically changing coinbase/difficulty/blockGasLimit on every block

### LINKED ISSUE #24: Add option to change block-level parameters in VM mode

axic:
Fixes #24.

### DIFF
--- a/src/app/txRunner.js
+++ b/src/app/txRunner.js
@@ -97,11 +97,16 @@ TxRunner.prototype.execute = function () {
         data: new Buffer(data.slice(2), 'hex')
       })
       tx.sign(account.privateKey)
+
+      const coinbases = [ '0x0e9281e9c6a0808672eaba6bd1220e144c9bb07a', '0x8945a1288dc78a6d8952a92c77aee6730b414778', '0x94d76e24f818426ae84aa404140e8d5f60e10e7e' ]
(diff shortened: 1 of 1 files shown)


src/app/txRunner.js
```diff
@@ -97,11 +97,16 @@ TxRunner.prototype.execute = function () {
...
+      const difficulties = [ new BN('69762765929000', 10), new BN('70762765929000', 10), new BN('71762765929000', 10) ]
       var block = new EthJSBlock({
         header: {
-          // FIXME: support coinbase, difficulty and gasLimit
           timestamp: new Date().getTime() / 1000 | 0,
-          number: self.blockNumber
+          number: self.blockNumber,
+          coinbase: coinbases[self.blockNumber % coinbases.length],
+          difficulty: difficulties[self.blockNumber % difficulties.length],
+          gasLimit: new BN(gasLimit, 10).imuln(2)
```
yann300: why `.imuln(2)` ?

axic: Double the transaction limit, just to assume the the transaction wasn't the only one in the "block".

PR #354, opened 2016-12-16
Move import processing from the compiler

axic:


axic:
@chriseth @yann300 please review

axic:
@chriseth 0.4.7 and 0.2.2 with github and local (other tab) imports.

PR #355, opened 2016-12-20
Support importing Swarm files

axic:
Implements #356 / #357. Depends on #366.

This requires CORS enabled on swarm-gateways.net, which is in progress (https://github.com/ethereum/go-ethereum/issues/3350).

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -531,7 +531,7 @@ var run = function () {
         files.addReadOnly(url, content)
         cb(null, content)
       })
-    } else if ((match = /^bzzr:\/\/?([0-9a-fA-F]{64}$)/.exec(url))) {
+    } else if ((match = /^(bzzr?:\/\/?.*)$/.exec(url))) {
       $('#output').append($('<div/>').append($('<pre/>').text('Loading ' + url + ' ...')))
       swarmgw.get(match[1], function (err, content) {
         if (err) {
--- a/src/app.js
+++ b/src/app.js
@@ -542,7 +542,8 @@ var run = function () {
         files.addReadOnly(url, content)
         cb(null, content)
       })
-    } else if ((match = /^(bzzr?:\/\/?.*)$/.exec(url))) {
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -429,13 +430,23 @@ var run = function () {
...
+      handleGithubCall(match[3], match[4], function (err, content) {
+        if (err) {
+          cb('Unable to import "' + url + '": ' + err)
+          return
+        }
+
+        cachedRemoteFiles[url] = content
+        cb(null, content)
+      })
+    } else if ((match = /^bzzr:\/\/?([0-9a-fA-F]{64}$)/.exec(url))) {
```
chriseth: I think we should also support `bzz://` because that supports directory structures.

axic: The rule for that would be something like `^(bzz:\/\/?[0-9a-fA-F]{64}/.*)$`. Probably better to catch both with `bzzr?` and let the Swarm client do its work.

axic:
Need to update `swarmgw` first to support other than `bzzr://<hash>` before merging this.

axic:
Added a question to geth: https://github.com/ethereum/go-ethereum/issues/3523

axic:
There's still no CORS on the endpoint...

axic:
Supports all three Swarm URLs and CORS is now enabled on the gateways. Can be merged.

src/app.js
```diff
@@ -515,12 +516,36 @@ var run = function () {
       })
   }
 
+  function handleSwarmImport (url, cb) {
+    swarmgw.get(url, function (err, content) {
+      // retry if this failed and we're connected via RPC
+      if (err && !executionContext.isVM()) {
```
chriseth: At some point, we should reverse the order here, I tihnk.

axic: Unfortunately we have to remove the RPC support because it will be removed, I have been told by the Swarm team.

PR #361, opened 2017-01-04
Check the current block gas limit before sending the tx

yann300:


src/app/txRunner.js
```diff
@@ -65,25 +65,26 @@ TxRunner.prototype.execute = function () {
         })
       })
     } else {
-      self.web3.eth.estimateGas(tx, function (err, resp) {
+      self.web3.eth.estimateGas(tx, function (err, gasEstimation) {
```
axic: Can you please split this into two commits? (changing `resp` to `gasEstimation` and the actual change)

src/app/txRunner.js
```diff
@@ -65,25 +65,26 @@ TxRunner.prototype.execute = function () {
...
         if (err) {
-          return callback(err, resp)
+          return callback(err, gasEstimation)
         }
+        self.web3.eth.getBlock('latest', function (err, block) {
+          if (err) {
+            return callback(err)
+          } else {
+            var blockGasLimit = Math.floor(block.gasLimit - block.gasLimit / 1024)
+            tx.gas = blockGasLimit < gasEstimation ? blockGasLimit : gasEstimation
```
axic: Is there any point running the transaction with a truncated estimate?

src/app/txRunner.js
```diff
@@ -65,25 +65,26 @@ TxRunner.prototype.execute = function () {
...
 
-        if (resp > gasLimit) {
-          return callback('Gas required exceeds limit: ' + resp)
```
axic: I think this is useful, because the user can set a hard limit (`gasLimit`) in the UI.

yann300: yes ok

src/app/txRunner.js
```diff
@@ -69,21 +69,24 @@ TxRunner.prototype.execute = function () {
...
-
-        sendTransaction(tx, function (err, resp) {
+        self.web3.eth.getBlock('latest', function (err, block) {
           if (err) {
-            return callback(err, resp)
-          }
+            return callback(err)
+          } else {
+            var blockGasLimit = Math.floor(block.gasLimit - block.gasLimit / 1024)
+            tx.gas = blockGasLimit < resp ? blockGasLimit : resp
```
axic: I'm just wondering, shouldn't we just take it as an error if the estimate is higher than the block limit? It will very likely fail.

src/app/txRunner.js
```diff
@@ -65,25 +65,28 @@ TxRunner.prototype.execute = function () {
...
-
-        sendTransaction(tx, function (err, resp) {
+        self.web3.eth.getBlock('latest', function (err, block) {
           if (err) {
-            return callback(err, resp)
-          }
+            return callback(err)
+          } else {
+            var blockGasLimit = Math.floor(block.gasLimit - block.gasLimit / 1024)
+            tx.gas = blockGasLimit < gasEstimation ? blockGasLimit : gasEstimation
```
axic: I'm just wondering, shouldn't we just take it as an error if the estimate is higher than the block limit? It will very likely fail.

chriseth: The original problem here was that go-ethereum returns an estimate that is higher than the block gas limit if an exception occurs and this PR hopefully fixes it. Note that you would want to execute failing transactions exactly for debugging where the exception comes from.

axic: Ah I see, so practically this is just a workaround for our inability to tell that the estimate failed.

axic: Added a comment into the code.

PR #365, opened 2017-01-06
Merge ui-helper and renderer

axic:


axic:
The reason for this is that ui-helper is not a helper and is only used by the renderer. Additionally, the main and the details section of the contract data was split into two files.

PR #367, opened 2017-01-06
Editor cleanup [WIP]

### LINKED ISSUE #64: Group storage related methods

axic:
Fixes #64.


chriseth:
Tests fail.

axic:
Need to introduce some events for the app to hook onto, such as:
- `fileSwitched` (switched to new session)
- ?

axic:
Obsoleted by #401.

PR #372, opened 2017-01-11
Move storage.editorSize to app.js

axic:


axic:
Does it actually makes sense keeping the editor size in storage and not in cookies? Would simplify the code by removing the `fileNameFromKey` stuff.

chriseth:
Let's please not use cookies.

Issue #374, opened 2017-01-11
Create browser-solidity configuration


axic:
It should be a JSON file named `.browser-solidity.json` as stored in the same storage layout as other files. Potentially it should be hidden in the editor.

The `editorSize` settings should move into this configuration. In the future it could also hold other variables such as auto-compilation is enabled or not.

Benefit: simple configuration file which can be saved and reloaded together with other files and/or gists.

chriseth:
I don't really see the benefit in having this two-layer abstraction for the settings. What about `localStorage['settings']['editorSize']`?

axic:
Each localStorage entry is a string, you cannot have `localStorage['settings']['editorSize']`. Will it automatically JSONify it?

The benefit is it can downloaded using the download feature, chrome would synchronize it and also gists could contain it.

axic:
Also: we could move all the settings we have in the URL hash into this file and save it to gists. That would mean only the gist hash is needed to load a gist with all the original settings.

axic:
Heh, I had opened a similar issue #150 a while ago.

axic:
Implemented by #375.

PR #375, opened 2017-01-11
Add config file API

axic:
Implements #374. Depends on #376.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -11,6 +11,7 @@ var GistHandler = require('./app/gist-handler')
 var gistHandler = new GistHandler()
 
 var Storage = require('./app/storage')
+var Config = require('./app/config')
 var Editor = require('./app/editor')
 var Renderer = require('./app/renderer')
 var Compiler = require('./app/compiler')
@@ -40,6 +41,7 @@ var run = function () {
   var self = this
   this.event = new EventManager()
   var storage = new Storage()
+  var config = new Config(storage)
 
   function loadFiles (files) {
     for (var f in files) {
--- a/src/app.js
+++ b/src/app.js
@@ -370,6 +370,7 @@ var run = function () {
   // ----------------- resizeable ui ---------------
 
   var EDITOR_SIZE_KEY = 'editor-size-cache'
+  var EDITOR_WINDOW_SIZE = 'editorWindowSize'
 
   var dragging = false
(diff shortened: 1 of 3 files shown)


src/app.js
```diff
@@ -418,10 +421,18 @@ var run = function () {
     }
   })
 
+  // convert old browser-solidity
```
axic: Do we really need to carry over this single config option? Probably not such a big deal.

axic: #377 removes a clean way to carry this on, I would say it is not a big deal losing this information

chriseth: Carrying over config is not too important.

src/app/config.js
```diff
@@ -0,0 +1,30 @@
...
+
+var CONFIG_FILE = utils.fileKey('.browser-solidity.json')
+
+function Config (storage) {
+  this.keys = {}
+
+  // load on instantiation
+  var config = storage.get(CONFIG_FILE)
+  if (config) {
+    this.keys = JSON.parse(config)
```
chriseth: catch exception?

src/app/config.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
+
+var utils = require('./utils')
+
+var CONFIG_FILE = utils.fileKey('.browser-solidity.json')
+
+function Config (storage) {
+  this.keys = {}
```
chriseth: This does not only contains the keys but also the values.

src/app.js
axic: Realized that I still want to keep there the removal of the key, so for the moment this can stay. In #377 it will only delete key.

axic:
@chriseth updated

src/app/config.js
```diff
@@ -0,0 +1,36 @@
...
+function Config (storage) {
+  this.items = {}
+
+  // load on instantiation
+  try {
+    var config = storage.get(CONFIG_FILE)
+    if (config) {
+      this.items = JSON.parse(config)
+    }
+  } catch (exception) {
```
chriseth: We should really set the editor into some fail-safe mode if this happens. Currently, we do not yet store much valuable information in the config, but as soon as we do, we should not overwrite existing config files which are malformed without the user's confirmation.

axic: Can you add an issue for that so we don't forget?

PR #377, opened 2017-01-11
Storage reworked

axic:
This changes the `storage` API to only support files, which have a path and internally prefixes them with `sol:` for `localStorage`.

e.g. `ballot.sol` is stored as `sol:ballot.sol`, while `std/owned.sol` is `sol:std/owned.sol`.

Basically this will remove the need for `utils.fileKey`, `utils.fileNameFromKey`, `utils.isCachedFile` and all the complexity those cause.

Will need to be applied after #367 and #375.

A follow up change will be improving the editor to validate filenames and display a directory structure.

### DIFF
--- a/src/app/storage.js
+++ b/src/app/storage.js
@@ -19,7 +19,7 @@ function Storage () {
   }
 
   this.keys = function () {
-    return safeKeys().filter(function (item) { return item.replace('sol:', '') })
+    return safeKeys().filter(function (item) { return item.replace(/^\.sol:/, '') })
   }
 
   this.remove = function (name) {
@@ -45,7 +45,7 @@ function Storage () {
   safeKeys().forEach(function (name) {
     if (name.indexOf('sol-cache-file-', 0) === 0) {
       var content = window.localStorage.getItem(name)
-      window.localStorage.setItem('sol:' + name.replace('sol-cache-file-', ''), content)
+      window.localStorage.setItem(name.replace(/^sol-cache-file-/, 'sol:'), content)
       window.localStorage.removeItem(name)
     }
   })
--- a/src/app/storage.js
+++ b/src/app/storage.js
@@ -49,6 +49,9 @@ function Storage () {
       window.localStorage.removeItem(name)
     }
   })
+
+  // remove obsolete key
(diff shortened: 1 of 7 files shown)


axic:
Alternatively, it might be cleaner to go this way first, and apply renaming/tidying up changes from #367.

axic:
@chriseth I think we should merge this instead and then adjust #367 (it will be significantly less changes that way).

chriseth:
Tests are failing.

src/app/storage.js
```diff
@@ -6,20 +6,24 @@ function Storage () {
...
   }
 
-  this.keys = function () {
+  function safeKeys () {
     // NOTE: this is a workaround for some browsers
     return Object.keys(window.localStorage).filter(function (item) { return item !== null && item !== undefined })
   }
 
+  this.keys = function () {
+    return safeKeys().filter(function (item) { return item.replace('sol:', '') })
```
chriseth: Better use `replace(/^\.sol:/, '')`?

src/app/storage.js
```diff
@@ -36,6 +40,18 @@ function Storage () {
     }
     this.set(filename, content)
   }
+
+  // on startup, upgrade the old storage layout
+  safeKeys().forEach(function (name) {
+    if (name.indexOf('sol-cache-file-', 0) === 0) {
+      var content = window.localStorage.getItem(name)
+      window.localStorage.setItem('sol:' + name.replace('sol-cache-file-', ''), content)
```
chriseth: `name.replace(/^sol-cache-file-/, 'sol:')`?

axic:
> Tests are failing.

Still not sure why.

src/app/storage.js
axic: > Better use replace(/^\.sol:/, '')

Yep that's better.

src/app/storage.js
axic: Added.

src/app/storage.js
axic: > Also ignore any items that do not start with `sol:`?

Added.

axic:
@chriseth fixed, ready to merge 👍

PR #380, opened 2017-01-12
Introduce (metadata) publish button

### LINKED ISSUE #362: Include publish button
This will publish the source codes and the metadata on Swarm/IPFS.

It should be placed next to the "Create" button. It should be disabled if the "Create" button is disabled.

axic:
Fixes #362.

### DIFF
--- a/assets/css/universal-dapp.css
+++ b/assets/css/universal-dapp.css
@@ -173,7 +173,7 @@
 
 .udapp input,
 .udapp button {
-	width: 33%;
+	width: 20%;
 	display: block;
 	float: left;
 	padding: 0.25em;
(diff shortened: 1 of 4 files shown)


axic:
This doesn't have the actual publishing implementing, but it depends on https://github.com/ethereum/go-ethereum/issues/3523.

axic:
Since this now displays a warning message ("Feature not implemented."), we might as well merge it.

chriseth:
I don't think we should add a button that is not working.

axic:
> I don't think we should add a button that is not working.

Have you seen the Reddit button? 😉

axic:
Added publishing of metadata. Still need to include way to publish sources.

It should also verify that the published hash matches the pre-calculated hash.

chriseth:
What is missing for publishing the source?

axic:
I have improved version I need to push here, after that we should be able to merge it.

axic:
@yann300 @chriseth I think this can be merged

Issue #133, opened 2016-07-15
Support IPFS/Swarm for files


axic:
Use the ipfs / swarm gateways for now.
- [ ] can we use gateway.ipfs.io for that?
- [ ] does it support CORS?

chriseth:
They can already be loaded just fine via the gateways :-)

axic:
Yes, but we could support Swarm through RPC also. If it makes sense.

axic:
The *public IPFS gateway* doesn't have CORS enabled (yet): https://github.com/ipfs/go-ipfs/issues/2239

PR #381, opened 2017-01-12
Support importing IPFS files

axic:
Implements #133. Blocked by not having CORS.

Since I've added this to try out, it may make sense having it here hoping it will be supported soon.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -529,6 +529,22 @@ var run = function () {
       }
     })
   }
+
+  function handleIPFS (url, cb) {
+    // replace ipfs:// with /ipfs/
+    url = url.replace(/^ipfs:\/\/?/, '/ipfs/')
+
+    $('#output').append($('<div/>').append($('<pre/>').text('Loading ' + url + ' ...')))
+    return $.get('https://gateway.ipfs.io/' + url)
+      .done(function (data) {
+        cb(null, data)
+      })
+      .fail(function (xhr, text, err) {
+        // NOTE: on some browsers, err equals to '' for certain errors (such as offline browser)
+        cb(err || 'Unknown transport error')
+      })
+  }
+
   function handleImportCall (url, cb) {
     var match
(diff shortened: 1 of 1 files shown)


axic:
There's still no CORS on the endpoint...

axic:
@yann300 @chriseth this is working now 😉

yann300:
cool!
If you have time could you just quick update the doc?
https://github.com/ethereum/remix/blob/master/docs/tutorial_import.rst
is this possible to push to ipfs also?

axic:
> is this possible to push to ipfs also?

The public gateway only supports reading.

src/app.js
```diff
@@ -557,6 +572,18 @@ var run = function () {
         files.addReadOnly(url, content)
         cb(null, content)
       })
+    } else if ((match = /^(ipfs:\/\/?.+)/.exec(url))) {
+      $('#output').append($('<div/>').append($('<pre/>').text('Loading ' + url + ' ...')))
+      handleIPFS(match[1], function (err, content) {
+        if (err) {
+          cb('Unable to import "' + url + '": ' + err)
```
chriseth: There is lots of code duplication between the three remote imports. Can you reduce this to something like
```js
var importers = [{match:  /^(https?:\/\/)?(www.)?github.com\/([^/]*\/[^/]*)\/(.*)/, importer = handleGithubImport}, {...}]
for (...) { ...}
```
?

axic: A planned to address that in a future PR. The only thing I've did in the direction is to move the log 😉

chriseth: Please to it in this pr, if possible. It's not a complicated change and three times the same code is already two times too much :-)

axic:
@chriseth @yann300 updated to reduce code duplication.

axic:
Moved the refactoring to a separate PR because it makes it really hard to review.

PR #400, opened 2017-01-18
Implement the files API

axic:


src/app/files.js
```diff
@@ -0,0 +1,98 @@
...
+var EventManager = require('../lib/eventManager')
+
+function Files (storage) {
+  var event = new EventManager()
+  this.event = event
+  var readonly = {}
+
+  this.exists = function (path) {
+    // NOTE: ignore the config file
+    if (path == '.browser-solidity.json') {
```
chriseth: This is kind of a "special file" or system file, so perhaps we can hide it only at the UI level? Although that might be more difficult.

src/app/files.js
```diff
@@ -0,0 +1,98 @@
...
+
+  this.set = function (path, content) {
+    // NOTE: ignore the config file
+    if (path == '.browser-solidity.json') {
+      return
+    }
+
+    if (!this.isReadOnly(path)) {
+      if (storage.exists(path)) {
+        event.trigger('fileAdded', [path])
```
chriseth: The cases have to be inverted.

src/app/files.js
```diff
@@ -0,0 +1,98 @@
...
+      } else {
+        event.trigger('fileChanged', [path])
```
chriseth: I think it is better to set the content before triggering the event.

src/app/files.js
```diff
@@ -0,0 +1,98 @@
...
+      }
+      storage.set(path, content)
+    }
+  }
+
+  this.addReadOnly = function (path, content) {
+    readonly[path] = content
```
chriseth: Shouldn't we check if the file already exists in storage?

chriseth:
With request to "who has access to storage": Since we are adding another layer of paths, I'm specifically thinking about the case where we have a name clash and storage paths are added underneath readonly files.

src/app/files.js
axic: In the future I would move it outside this and change the config API to support that (i.e. catch events modifying this file). That way the user could both change in the editor as well as in the UI.

src/app/files.js
axic: Fixed.

src/app/files.js
axic: Fixed.

src/app/files.js
axic: Fixed.

axic:
> How do we want to handle errors in that API? 

I was thinking to return true/false to indicate success.

> Also, are we certain that this is the only module that can modify storage?

There will be two modules able to do that: files and config.

> What happens if the same website is open in two browser windows, do they share access to storage?

They do. The conflicting file will be the currently open session as saving that will overwrite it in storage.

> I'm specifically thinking about the case where we have a name clash and storage paths are added underneath readonly files.

As a next step I would want to add path resolution and rejecting such cases. e.g. if `this/is/a/file` is present trying to add `this/is/a/file/andthistoo` should fail.

axic:
Also we could add storage caching here, though not sure that is beneficial. As long as the outside API of Files is nailed down we have the flexibility to change the internals.

src/app/files.js
```diff
@@ -0,0 +1,104 @@
...
+      return readonly[path]
+    } else {
+      return storage.get(path)
+    }
+  }
+
+  this.set = function (path, content) {
+    // NOTE: ignore the config file
+    if (path === '.browser-solidity.json') {
+      return
```
chriseth: Please add error handling, this should return false, the other cases should return true.

src/app/files.js
```diff
@@ -0,0 +1,104 @@
...
+      return true
+    } else {
+      return storage.exists(path)
+    }
+  }
+
+  this.get = function (path) {
+    // NOTE: ignore the config file
+    if (path === '.browser-solidity.json') {
+      return
```
chriseth: Could we use an explicit `return null` here?

src/app/files.js
```diff
@@ -0,0 +1,104 @@
...
+
+function Files (storage) {
+  var event = new EventManager()
+  this.event = event
+  var readonly = {}
+
+  this.exists = function (path) {
+    // NOTE: ignore the config file
+    if (path === '.browser-solidity.json') {
+      return
```
chriseth: `return false`?

src/app/files.js
axic: I'm not sure whether `localStrage.getItem` returns `null` or `undefined`, but if we return an explicit value these should match.

src/app/files.js
axic: One more thing to consider: `localStorage.setItem` doesn't return a value, but throws an exception if a fault happens. We could either catch that in the Storage API and return true/false or catch it here.

src/app/files.js
axic: According to Mozilla:
> A DOMString containing the value of the key. If the key does not exist, null is returned.

src/app/storage.js
```diff
@@ -28,12 +33,16 @@ function Storage () {
 
   this.remove = function (name) {
     window.localStorage.removeItem('sol:' + name)
+    return true
   }
 
   this.rename = function (originalName, newName) {
     var content = this.get(originalName)
-    this.set(newName, content)
+    if (!this.set(newName, content)) {
```
chriseth: Do we really want this to succeed in case the new name already exists?

src/app/files.js
```diff
@@ -0,0 +1,119 @@
...
+  this.event = event
+  var readonly = {}
+
+  this.exists = function (path) {
+    // NOTE: ignore the config file
+    if (path === '.browser-solidity.json') {
+      return false
+    }
+
+    if (this.isReadOnly(path)) {
```
chriseth: What about `return this.isReadOnly(path) || storage.exists(path);`? And similar below.

src/app/files.js
```diff
@@ -0,0 +1,119 @@
...
+      return true
+    }
+
+    return false
+  }
+
+  this.addReadOnly = function (path, content) {
+    if (!storage.exists(path)) {
+      readonly[path] = content
+      event.trigger('fileAdded', [path])
```
chriseth: We could supply the path and the readonly property, so that the UI does not have to ask again.

src/app/storage.js
axic: The check for that is currently handled outside of storage.

src/app/files.js
axic: I was thinking about that, but the time I wrote this the feature wasn't used by the app. I can add it though.

src/app/files.js
axic: I don't mind. The code currently consistently has `if (this.isReadOnly(path)) {`, but it wouldn't make much of a difference.

src/app/storage.js
axic: POSIX:

> The rename() system call causes the link named old to be renamed as new.  If new exists, it is first removed.

src/app/files.js
chriseth: I think the former is easier to read and understand.

src/app/storage.js
chriseth: I don't think we should follow posix here. What is the use-case of overwriting by renaming? I know that we do not make atomic promises here, but I still think that it is better to fail in this case. After all, this is an IDE and not an operating system. If you rename a file you are currently working on, you want to change its name and not another file to disappear at the same time.

axic: Note that the IDE part is not interacting with this API at all, only through the "files API" which does this check.

axic: As discussed, `set()` operates in the same manner - overwrites if it exists. In order to make it consistent, that would have to change too.

I think I would postpone this one item until the next step when we see which layer it makes the most sense in. Happy to change it if it turns out to be useful here.

Issue #384, opened 2017-01-16
Storage overhaul


axic:
- [x] ~consolidate editor/storage/session calls (#367)~
- [x] move editorSize to app (#372)
- [x] move Chrome syncing to app (#373)
- [x] add config file API (#374)
- [x] document all uses of storage / editor
- [x] group remote-import code in app (split between `storage.loadFiles`, `editor.packageFiles`, `app:loadFiles`, `app:copyOver`)
- [x] add storage key prefix (#377)
- [x] implement the files API (#400)
- [x] add editor events (replace `onSetChanged` with an event) (#403)
- [x] add storage events (`fileAdded`/`fileChanged`, `fileRemoved`, `fileRenamed`) (#400)
- [x] sync file before session switch (i.e. switching tabs and before adding a new file) (#427)
- [x] rewrite editor (#401)

(This replaces #64 - see it for some details on events/methods)

axic:
I'm thinking to change "editor" to have three state variables:
- list of files
- sessions
- storage items

List of files contains all the available files with flags (e.g. read-only, temporary). This is used to show tabs, etc.

Sessions contains the currently loaded sessions - right now that equals to list of files, but when we would have a lot files, it probably doesn't make sense to have all of them always loaded.

Storage items only contain non-readonly/non-temporary files.

axic:
Files
- exists(path)
- get(path) -> content
- set(path, content)
- setReadOnly(bool)
- remove(path)
- rename(path, path)
- list() -> paths
- emits events:
  - fileAdded
  - fileChanged
  - fileRemoved
  - fileRenamed

Storage
- exists(path)
- get(path) -> content
- set(path, content)
- keys() -> paths
- remove(path)
- rename(path, path)

Editor
- open(path, content)
- openReadOnly(path, content)
- get(path)
- discard(path)
- emits event: changed

Basically the editor only handles the "current" file, but it can keep a cache of sessions.

(Note, Files / Storage should emit more events at some point.)

chriseth:
Sounds good! Just one tiny thing: I think there should be a single function call to add new content that is read only. Otherwise we have to deal with things like files being open in the editor and suddenly turning read-only.

axic:
Actually my implementation does exactly that: get/set is a thin wrapper over stroage, while addReadOnly adds a readonly file.

PR #401, opened 2017-01-19
Rewrite the editor based on the  files API

### LINKED ISSUE #344: Readonly tabs
When importing sources from github or swarm, it is useful to be able to display them. For that, read-only tabs should be available.

axic:
Fixes #344 and #384.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -47,8 +47,11 @@ var run = function () {
   var config = new Config(storage)
   var currentFile
 
+  // return all the files, except the temporary/readonly ones
   function packageFiles () {
-    return files.list().map(function (path) { return { content: files.get(path) } })
+    return files.list()
+      .filter(function (path) { if (!files.isReadOnly(path)) { return path } })
+      .map(function (path) { return { content: files.get(path) } })
   }
 
   function createNonClashingName (path) {
--- a/src/app.js
+++ b/src/app.js
@@ -123,11 +123,11 @@ var run = function () {
         console.log('comparing to cloud', key, resp)
         if (typeof resp[key] !== 'undefined' && obj[key] !== resp[key] && confirm('Overwrite "' + key + '"? Click Ok to overwrite local file with file from cloud. Cancel will push your local file to the cloud.')) {
           console.log('Overwriting', key)
-          storage.set(key, resp[key])
+          files.set(key, resp[key])
           updateFiles()
         } else {
           console.log('add to obj', obj, key)
-          obj[key] = storage.get(key)
+          obj[key] = files.get(key)
         }
         done++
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -40,16 +43,33 @@ var run = function () {
   var self = this
   this.event = new EventManager()
   var storage = new Storage()
+  var files = new Files(storage)
   var config = new Config(storage)
+  var currentFile
+
+  function packageFiles () {
+    return files.list().map(function (path) { return { content: files.get(path) } })
```
chriseth: Does this include the readonly files?

src/app.js
```diff
@@ -463,23 +493,19 @@ var run = function () {
...
     } else if ((githubMatch = /^(https?:\/\/)?(www.)?github.com\/([^/]*\/[^/]*)\/(.*)/.exec(url))) {
       handleGithubCall(githubMatch[3], githubMatch[4], function (err, content) {
         if (err) {
           cb('Unable to import "' + url + '": ' + err)
           return
         }
 
-        cachedRemoteFiles[url] = content
+        // FIXME: at some point we should invalidate the cache
+        files.addReadOnly(url, content)
```
chriseth: Wohoo! :-)

src/app.js
axic: Yes, but we can add a simple line `.filter(function (path) { if (!files.isReadOnly(path) return path })` to exclude readonly.

chriseth: Sure, and I think we should.

axic: Added it last week, but won't push until I can rebase from #400 and #403.

axic:
@chriseth rebased, but this will need some more work. Nevertheless a quick review is appreciated.

src/app.js
```diff
@@ -269,43 +289,56 @@ var run = function () {
...
+
+    if (files.isReadOnly(file)) {
+      editor.openReadOnly(file, files.get(file))
+    } else {
+      editor.open(file, files.get(file))
+    }
   }
 
+  function switchToNextFile () {
+    switchToFile(Object.keys(files.list())[0])
```
yann300: I checked browser test failure. It might fails because of that. 
Are we sure that the array is always not empty?
and I don't get why this will switch to the next file (is that because the previous files has been "poped" ?

axic: > Are we sure that the array is always not empty?

It should not be empty on startup, but it could become empty, that's true. Let me fix that.

> don't get why this will switch to the next file

It switched to the first file in the list, not actually the next file. Should rename the method I guess.

axic:
@yann300 rebased this, please review

PR #403, opened 2017-01-23
Introduce editor events

axic:


src/app.js
```diff
@@ -573,7 +575,9 @@ var run = function () {
     compileTimeout = window.setTimeout(runCompiler, 300)
   }
 
-  editor.onChangeSetup(editorOnChange)
+  editor.event.register('changed', editorOnChange)
```
chriseth: I'm wondering whether we should rename the two events to `contentChanged` and `fileSwitched` to make a better distinction between the two events.

axic: Went for `sessionSwitched` and `contentChanged.`

PR #407, opened 2017-02-01
remove e6-shim

yann300:
address https://github.com/ethereum/browser-solidity/issues/383

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "brace": "^0.8.0",
     "browserify": "^13.0.0",
     "csslint": "^1.0.2",
-    "es6-shim": "^0.35.1",
     "ethereum-remix": "https://github.com/ethereum/remix",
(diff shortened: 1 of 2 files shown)


axic:
@yann300 you need also to remove the `require()` statement in `index.js`

axic:
@yann300 do we have a sauceconnect test browser version which is non-ES6?

axic:
Firefox <50 seems like a good candidate to include in the tests.

yann300:
I'm not sure which is the current firefox version that sauceconnect use. but IE11 is the ie version sc use for testing and it is not es6.
In any case it could make sense to add also Firefox<50

axic:
Would be nice to have both Firefox 49 and >50 as a definitive test case.

yann300:
yes will add a task

yann300:
tests are failing because of

PR #411, opened 2017-02-02
Breakpoints

yann300:


src/app/debugger.js
```diff
@@ -16,19 +16,51 @@ function Debugger (id, editor, compiler, executionContextEvent, switchToFile, of
...
+        break
+      }
+    }
+    source.row = row
+    return source
+  }
+
+  editor.event.register('breakpointCleared', (fileName, row) => {
+    if (self.compiler.lastCompilationResult.data) {
+      this.breakPointManager.remove(convertSourceLocation(this, fileName, row))
```
chriseth: The breakpoints should be file name based, the index is only valid for a single compilation run, but the breakpoints are not recreated.

src/app/debugger.js
```diff
@@ -61,21 +93,44 @@ Debugger.prototype.debug = function (txHash) {
   })
 }
 
+Debugger.prototype.switchFile = function (rawLocation) {
+  var name = this.editor.getCacheFile() // current opened tab
+  var source = this.compiler.lastCompilationResult.data.sourceList[rawLocation.file] // auto switch to that tab
```
chriseth: what if lastCompilationResult does not exist?

src/app/editor.js
```diff
@@ -17,6 +17,31 @@ function Editor (doNotLoadStorage, storage) {
...
 
+  var self = this
+  editor.on('guttermousedown', function (e) {
+    var target = e.domEvent.target
+    if (target.className.indexOf('ace_gutter-cell') === -1) {
+      return
+    }
+    var row = e.getDocumentPosition().row
+    var brkpoints = e.editor.session.getBreakpoints()
+    for (var k in brkpoints) {
```
chriseth: please change to `breakpoints`

axic:
Needs to be rebased.

PR #416, opened 2017-02-07
Fix geth deploy

yann300:
fix https://github.com/ethereum/browser-solidity/issues/412

### DIFF
--- a/src/app/renderer.js
+++ b/src/app/renderer.js
@@ -165,6 +165,7 @@ Renderer.prototype.contracts = function (data, source) {
       code += 'var ' + inp.name + ' = /* var of type ' + inp.type + ' here */ ;\n'
     })
 
+    contractName = contractName.replace(/[:./]/g, '_')
     code += 'var ' + contractName + 'Contract = web3.eth.contract(' + jsonInterface.replace('\n', '') + ');' +
       '\nvar ' + contractName + ' = ' + contractName + 'Contract.new('
 


src/app/renderer.js
```diff
@@ -165,8 +165,8 @@ Renderer.prototype.contracts = function (data, source) {
       code += 'var ' + inp.name + ' = /* var of type ' + inp.type + ' here */ ;\n'
     })
 
-    code += 'var ' + contractName + 'Contract = web3.eth.contract(' + jsonInterface.replace('\n', '') + ');' +
-      '\nvar ' + contractName + ' = ' + contractName + 'Contract.new('
+    code += 'var ' + contractName.replace(':', '_') + 'Contract = web3.eth.contract(' + jsonInterface.replace('\n', '') + ');' +
```
chriseth: I think you also have to replace `/`

axic: `:`, `.`, `/` are replaced in solc, probably should follow the same here.

src/app/renderer.js
```diff
@@ -165,6 +165,7 @@ Renderer.prototype.contracts = function (data, source) {
       code += 'var ' + inp.name + ' = /* var of type ' + inp.type + ' here */ ;\n'
     })
 
+    contractName = contractName.replace(/(:|\.|\/|\s)/g, '_')
```
chriseth: I would propose to use`/[:./]/g`

yann300: ok

PR #418, opened 2017-02-08
use API [WIP]

yann300:


src/app.js
```diff
@@ -497,7 +497,47 @@ var run = function () {
...
 
-  var transactionDebugger = new Debugger('#debugger', editor, compiler, executionContext.event, switchToFile, offsetToLineColumnConverter)
+  var contentToolAPI = {
+    offsetToLineColumn: (location, file) => {
+      return offsetToLineColumnConverter.offsetToLineColumn(location, file, compiler.lastCompilationResult)
+    }
+  }
+
+  var ethToolAPI = {
+    toWei: (value, unit) => {
```
chriseth: This does not seem like it should be passed in.

src/app.js
```diff
@@ -497,7 +497,47 @@ var run = function () {
...
+    lastCompilationResult: () => {
+      return compiler.lastCompilationResult
+    }
+  }
+
+  var editorAPI = {
+    currentOpenedFile: () => {
+      return editor.getCacheFile()
+    },
+    addMarker: (range, css) => {
```
chriseth: Replace addMarker and removeMarker by `currentSourceLocation` - or should it perhaps be an event?

src/app.js
```diff
@@ -497,7 +497,47 @@ var run = function () {
...
+    }
+  }
+
+  var compilerAPI = {
+    lastCompilationResult: () => {
+      return compiler.lastCompilationResult
+    }
+  }
+
+  var editorAPI = {
```
chriseth: Please group by the component you pass this into, not by the component you take it from (and also adjust the naming to the nomenclature of the component you pass this object into).

PR #423, opened 2017-02-14
Make udapp.newAccount() async

axic:


src/universal-dapp.js
```diff
@@ -55,18 +55,19 @@ UniversalDApp.prototype.reset = function (contracts, transactionContextAPI, rend
   })
 }
 
-UniversalDApp.prototype.newAccount = function (password) {
+UniversalDApp.prototype.newAccount = function (password, cb) {
```
axic: It is not used yet, see #425.

PR #428, opened 2017-02-15
Show alerts on file errors

axic:


axic:
@yann300 @chriseth bad news, tab renaming / removal fails with error

src/app.js
```diff
@@ -654,23 +668,28 @@ var run = function () {
...
+      files.set(currentFile, input)
+    }
   }
 
   var previousInput = ''
   var compileTimeout = null
   var saveTimeout = null
 
   function editorOnChange () {
+    if (!currentFile) {
```
axic: Seems like my comment I made from my phone yesterday didn't go through.

This shouldn't be called at all of there's no session (aka no `currentFile`).

Issue #404, opened 2017-01-24
Remix naming


axic:
Since we're thinking to merge the two projects, I think the following layout might make sense:
- remix-core / remix-lib (the common libraries)
- remix-debug aka current remix (the debugger frontend which can be integrated into other projects)
- remix-ide aka browser-solidity (the stand-alone IDE frontend)

@yann300 @chriseth what do you think?

chriseth:
Sounds good! Are you thinking about creating npm packages and / or repositories along this structure?

yann300:
yes that's cool!
what would put in remix-core/-lib? code that is currently in the remix repo but not ui stuff?

axic:
> what would put in remix-core/-lib? code that is currently in the remix repo but not ui stuff?

Yes and possible removing the duplicate items in browser-solidity (perhaps even moving some generic stuff from here).

axic:
> Are you thinking about creating npm packages

Yes, though remix-ide wouldn't need a published package.

> repositories along this structure

Would be nice ultimately.

axic:
@yann300 can you go ahead with renaming this repo to `remix-ide`? It is quite confusing that actual IDE issues are being kept opened on the remix repo.

yann300:
oh yes, actually I always thought it was not possible to rename a repo.. it is actually :)

axic:
It is simple and as long as nobody creates another repo named browser-solidity, it will symlink here. However, not sure that github pages works the same way :(

Perhaps you need to push a page version with a big popup to a new repo callde browser-solidity so that people can migrate their contracts.

axic:
This became a bit more complicated in the last week:
- renaming `yann300.github.com/remix` to `ethereum.github.io/remix-ide-alpha` and `remix-alpha.ethereum.org`
- renaming `browser-solidity` to `remix-ide`

Steps:
- [x] display modal dialogs for all the new addresses on the appropriate urls
- [ ] change build to push to `remix-ide-alpha` instead of `yann300`
- [ ] push a stable build to `browser-solidity/gh-pages`
- [ ] rename `browser-solidity` to `remix-ide`
- [ ] create a new repo called `browser-solidity` and push the previous `gh-pages`  branch
- [ ] update build scripts to `remix-ide`

axic:
Done!

PR #429, opened 2017-02-16
ADD file explorer to remix (WIP)

serapath:
This is `work in progress`. I will add more implementation, but opened it just so that it's there :yum:

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -66,9 +66,9 @@ var run = function () {
   }
 
   // Add files received from remote instance (i.e. another browser-solidity)
-  function loadFiles (files) {
-    for (var f in files) {
-      files.set(createNonClashingName(f), files[f].content)
+  function loadFiles (filesSet) {
+    for (var f in filesSet) {
+      files.set(createNonClashingName(f), filesSet[f].content)
     }
     switchToNextFile()
   }


src/app/editor.js
```diff
@@ -2,10 +2,32 @@
 
 var EventManager = require('../lib/eventManager')
 
+var csjs = require('csjs-inject')
 var ace = require('brace')
 var Range = ace.acequire('ace/range').Range
 require('../mode-solidity.js')
 
+var css = csjs`
```
axic: Is it possible to try keeping the CSS in one place?

src/app.js
```diff
@@ -25,6 +25,10 @@ var OffsetToLineColumnConverter = require('./lib/offsetToLineColumnConverter')
 
 var examples = require('./app/example-contracts')
 
+// components
+var fileExplorer = require('_file-explorer')
```
axic: Why did you chose to make this a "package" and not just a file?

axic:
Is it possible to split the saucelabs changes into more commits and perhaps moving them into their own PR?

chriseth:
Congratulations to your first pull request! I hope the onboarding up to now was not too rough...

src/app/editor.js
serapath: sure - but honestly - is this a good idea?
I feel, since I started to put the css related to a certain component into the place of that component, that things became a lot easier.

Having tons of global variables is a bad practice and that is what a css with tons of global selectors is. The global stylesheet will grow longer and longer and people will start adding hacks to it that interfere with each other until it becomes really really hard to change styling, especially layout related stuff.

That's my experience.

serapath: And I imagine, that an IDE will grow a very very complex stylesheet over time...

axic: We're actually trying to move onto LESS and that would help. See #203.

src/app.js
serapath: Oh, just a habit. I can make it a file.
I usually maintain a manually manage `src/node_modules` folder and prefix with underscore, so that i can require it without having to remember the path and whether i need to do `require('./modulename')` or `require('../modulename')`...  the underscore will clearly communicate that it is a project internal module, because stuff on `npm` can't contain `_` in their name.

It also makes it easy to later on publish internal modules as public ones on `npm` if they actually become something useful on their own :-)

serapath:
@chriseth yeah, it's ok. The saucelab tests are still not running for me and I'm trying together with @yann300 to figure out what the reason is. I already figured, that node version 5.11.0 caused one problem and now with upgrading to his version v6.2.2 it seems to throw one error less, but it's still not working :P

I am documenting this process in the readme for others

serapath:
@axic 
ok - yes, I'll factor out the saucelab related changes after i put all the pieces together and make it a seperate pull request, but before i do that i'd just like to make the saucelab stuff work first if that's ok :-)

src/app.js
axic: Note that `node_modules` is in the `.gitignore` file so it shouldn't even be possible to commit anything below that directory :)

package.json
```diff
@@ -5,46 +5,29 @@
...
-    "babel-plugin-transform-es2015-parameters": "^6.18.0",
-    "babel-plugin-transform-es2015-shorthand-properties": "^6.18.0",
-    "babel-plugin-transform-es2015-spread": "^6.8.0",
-    "babel-plugin-transform-es2015-sticky-regex": "^6.8.0",
-    "babel-plugin-transform-es2015-template-literals": "^6.8.0",
-    "babel-plugin-transform-es2015-unicode-regex": "^6.11.0",
-    "babel-plugin-transform-regenerator": "^6.16.1",
-    "babel-polyfill": "^6.22.0",
-    "babelify": "^7.3.0",
+    "babel-preset-es2015": "^6.22.0",
```
yann300: I think we should do all that stuff in another PR
also there were a reason why we didn't use preset but I can't remember now. That might be because of the `async` plugin.

src/app/editor.js
serapath: Really? But LESS is a separate language and doesn't shield you from css cascading.
That sounds a bit sad, but if that decision is fix i can put it back into the css file

src/app.js
serapath: i added `!src/node_modules`

package.json
serapath: oh ok... hm, it was kind of easier to add all this via `babelify` which enables the live reloading.

Because before i had to manually `rm -rf build`, `npm run build`, and then hit refresh... while now, when i save my file in my editor, the browser auto-refreshes which makes feedback cycles a lot faster.

I can factor that into a seperate pull request too. Would be nice to get the reason why it's not supposed to be just `babel-preset-es2015`, because the livereload is really handy and i can see how to make it work the old way, but maybe it slightly complicates things... not sure.

Btw. it doesnt throw an error for me after this refactoring step

src/app.js
yann300: Not sure if we should make those changes that are really related to the internal architecture of the app right now in this PR.
Actually I don't really like the idea of having this manually manage `node_module`  with prefixed name, it looks like mixing dependencies and internal module. But ok that is very nice to require a module without having to remember the full path :)
Do we need to have this `package.json` file in the file-explorer folder? what is the benefit of having it?
If the reason of doing that is only because of the easier use of `require`,we should find another way to require internal class in a easiest way.
So merging this could be fine for me (only as a temporary approach of requiring internal stuff), we should think on a better way (if there is).

axic: For the moment I'd hope to keep it just in `src/lib/`. Once interfaces are stable we could think about moving these to separate modules (please check #404 for a direction we were taking).

yann300: maybe this stuff https://www.npmjs.com/package/app-module-path .....
In a separate PR please :)

chriseth: Hm, I think that using directories to group modules makes stuff much more visible. This grouping is lost if you can just `require("_modulename")` from everywhere.

package.json
chriseth: Live reloading? tell me more :-)

src/app.js
serapath: 1. you can require("_modulename") from everywhere within the project's `./src` folder.

2. From my experience stuff like `app-module-path` breaks in the browser, because browserify doesn't know how to deal with this. In the past I wrote `https://www.npmjs.com/package/rootrelative` but that was a naive approach which of course - again - doesnt work in browsers :/
3. you do not need an actual folder and a package.json for simple modules. The simplest form would actually be just `./src/node_modules/_my-private-module.js` ... a next step (for more complex private modules would be `./src/node_modules/_my-private-module/index.js` which results in lots of open editor tabs with the name "index.js" (which sucks) ...a step up from that would be `./src/node_modules/_my-private-module/my-private-module.js` + a package.json to describe `"maine": "my-private-module.js"`.

The advantage is, that a project wide simple search & replace works, because the requirepath is a constant instead of a relative path, which makes it trivial to publish an internal module to npm if it turns out a cool module should be re-used across different projects :-)

package.json
serapath: 1. It run's a file watcher, that watches an entry file and everything that is `require('...')`ed for changes (e.g. on file save in your editor) ...and then re-runs browserify.
 
2. additionally running browserify with the live-reload feature will on-the-fly run an additional dev server on your local machine and add a little javascript to your build file, that - when loaded in the browser - will estabish a websocket connection to that additional dev server and listen for updates.

3. once the file is re-build with all updates (1.) then the additional server is informed and sends the updates via the web-socket connection to all clients to refresh the browser with the new stuff.


.... there is even more steps - for example browsersync - which will even sync current scroll position and other stuff and if you open the project in multiple browsers and scroll in one -the others will magically scroll too to stay in sync

...a next step is to integrate "hot module reloading" (HMR) which will maintain STATE between refreshes (so if you are in a debugging sessions) and the code updates, then the STATE will be the same, but with the new version of the code - but this takes a little bit MORE effort to integrate :-) ...but might be worth it.

One project that might be worth checking out is https://github.com/amokjs/amok (which plugs into the browser from "behind the scenes" https://www.youtube.com/watch?v=1onqNIPSSeo) but otherwise what will totally do is https://www.npmjs.com/package/browserify-hmr which i have used in the past

src/app.js
serapath: One additional advantage i forgot to mention is, that - by default - browserify transforms only work on local files (thus files included in the current project), but not on `require(...)`ed modules files.

But, if a required file is written in - for example - coffee script, or uses some other babel features or transforms that inline assets into the built version, then those transforms need to be executed even though the current project doesnt know about it.

To enable this, a dependency (e.g. a local internal module) might specify browserify transforms inside it's related `package.json` file which will then be executed by browserify for only this dependent module :-) ...which allows module authors to specify all the transform needed.

I find this quite a neat feature - and this is why i usually stub a `package.json` for my internal modules, so that i can add custom transforms - e.g. for example for visual components that depend on certain fonts/images/etc... assets that i then either inline or otherwise "grab" without the main project's need for the knowledge of dependency specific "asset paths".

I can recommend the npm modules `urify` and/or `urify-emitter` in that context :-)

src/app/editor.js
serapath: LESS, SASS, STYLUS and the likes are a seperate custom language not native to the web to add features to styling that could be added by a native web language (e.g. javascript).

One core feature that is enabled by javascript beside removing the need for transpilation is, that, if - for example - there would ever be a decision to make styling dependent on user input (e.g. fade color based on users mouse position or change styling based on user GPS or audio/webcam etc.. input, then there will be a "state explosion" of styling if that would be tried with defining different styling states, because the changes are continuous.

But if javascript would be employed here - it's easily possible to solve that problem, because javascript can change stuff at runtim, while normal css preprocessors can only work during build time.

I can recommend https://www.npmjs.com/package/csjs - in case the "LESS decision" is not set in stone. :P

serapath: Oh by the way - the name is flawed, because one esential feature of `csjs` is to end "global namespace pollution" and thus preventing styles from leaking and crashing styles of other components written by other people.

It also changes the "seperation of technology" (html,css,js) into "separation of concerns" by co-locating (html,css,js) of one component (which is anyway heavily co-dependent by nature) in the same file, which is awesome for authors who want to search for everything (styles, markup and behavior) related to a single e.g. ButtonComponent.

Without that, the author would need to search different files and scroll through a lot of unrelated code or would need to make heavy use of editor assisted search functionality.

PR #432, opened 2017-02-19
ADD browserify livereload

serapath:


chriseth:
Tests are failing :-(

serapath:
How can i reproduce?

When i locally run `npm run test` if says `tests 14` and `pass 14`.

ci/browser_tests.sh
```diff
@@ -4,8 +4,10 @@ set -e
 
 SC_VERSION="4.4.0"
 SAUCECONNECT_URL="https://saucelabs.com/downloads/sc-$SC_VERSION-linux.tar.gz"
-SAUCECONNECT_USERNAME="chriseth"
-SAUCECONNECT_ACCESSKEY="b781828a-9e9c-43d8-89d4-2fbb879595ca"
+SAUCECONNECT_USERNAME="fairydust.work"
```
yann300: can you use `chriseth` ?

nightwatch.js
```diff
@@ -16,8 +17,10 @@ module.exports = {
       'selenium_host': 'ondemand.saucelabs.com',
       'selenium_port': 80,
       'silent': true,
-      'username': 'chriseth',
-      'access_key': 'b781828a-9e9c-43d8-89d4-2fbb879595ca',
+      // 'username': 'chriseth',
+      'username': 'fairydust.work',
```
yann300: `chriseth`

package.json
```diff
@@ -41,30 +30,45 @@
...
+    "notify-error": "^1.2.0",
+    "npm-run-all": "^4.0.2",
+    "onchange": "^3.2.1",
+    "rimraf": "^2.6.1",
+    "selenium-standalone": "^6.0.1",
     "solc": "https://github.com/ethereum/solc-js",
     "standard": "^8.5.0",
     "swarmgw": "^0.2.0",
     "tape": "^4.5.1",
+    "uglify-js": "^2.8.16",
```
yann300: why do you need that? (seems not used in this PR)

package.json
```diff
@@ -41,30 +30,45 @@
     "babel-plugin-transform-es2015-unicode-regex": "^6.11.0",
     "babel-plugin-transform-regenerator": "^6.16.1",
     "babel-polyfill": "^6.22.0",
+
+    "babel-plugin-yo-yoify": "^0.3.3",
```
yann300: cool! so we don't need anymore the `yoyoify` module right?

serapath: no.

package.json
serapath: the bin is `uglifyjs` and it should be used. i'll check

ci/browser_tests.sh
serapath: oops

nightwatch.js
serapath: yes

package.json
serapath: the `"minify"` task is using it - which is used by the `"build"` task

serapath:
So, I'm actually now pretty happy with all the changes. Ideally `package.json#browserify` config and the babelify transform take care of the local code being transpiled to es5 and dependencies specify whatever transforms they need. 


One thing that's left to do ideally even though it works already is:
* remove the `"babel"` field from the `package.json`.

I'm actually not sure what's using it, but removing it breaks stuff.
**Maybe you have an idea how to debug that?**

I'm assuming the code breaks if it's not there, because the code of dependencies is "globally transformed" using the transforms specified in the `package.json#babel` so that some modules that have es2015 code inside them get transpiled to es5 so they dont break in certain browsers... but maybe it's something else.

package.json
```diff
@@ -41,30 +30,44 @@
...
     "browserify": "^13.0.0",
+    "browserify-reload": "^1.0.3",
     "csslint": "^1.0.2",
     "ethereum-remix": "https://github.com/ethereum/remix",
     "ethereumjs-abi": "https://github.com/ethereumjs/ethereumjs-abi",
-    "ethereumjs-block": "^1.2.2",
-    "ethereumjs-tx": "^1.1.1",
-    "ethereumjs-util": "^4.4.0",
-    "ethereumjs-vm": "^2.0.1",
+    "ethereumjs-block": "https://github.com/ethereumjs/ethereumjs-block",
```
chriseth: Please do not use direct github references to these unless we require an unreleased feature (remix is fine).

chriseth: This seems to be required for now.

serapath: yes for now, until new releases contain the changes.
The published versions are 2 months or 7 months old.
but cool - merged :-) ... yay

PR #442, opened 2017-02-23
Rename .browser-solidity.json to .remix.config

### LINKED ISSUE #440: Rename .browser-solidity.json to .remix.config
Or some more fitting name.

axic:
Fixes #440.

Needs to be agreed on the name and tested.


chriseth:
I like the renaming.

PR #449, opened 2017-03-07
File explorer (WIP)

serapath:


src/app.js
```diff
@@ -156,15 +159,39 @@ var run = function () {
   chromeCloudSync()
 
   // ----------------- editor ----------------------
-
   var editor = new Editor()
-
-  // ----------------- tabbed menu -------------------
+  // -------------- fileExplorer ------------------
+  // @NOTE HTML#ID creates window['<ID name>'] element as a global
```
chriseth: What does this comment refer to?

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
+var setval = require('setval')
```
chriseth: Please keep to the current style and put the file explorer into `src/app`.

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+    margin-right: 3%;
+  }
+  .activeMode {
+    display: flex;
+    justify-content: space-between;
+  }
+`
+module.exports = fileExplorer
+
+function fileExplorer (container, files, editor) {
```
chriseth: We are currently trying to disentangle the various modules in a way that modules do not know anything about other modules. Each module only registers for events created by other modules or gets an object with callback functions and `app.js` connects the modules with each other. In this case, the file explorer would register on events by the `files` module (it only gets the event machine of files, not the files object itself) and it would receive an object with callbacks from `app.js` of the form `{ renameFile: function(oldName, newName) { /* call the respective function of the "files" module */ }, renameDirectory: function(...) { ... }, deleteFile: function(...) { ... }`. This way, the tree view and its action logic are cleanly separated.

src/node_modules/_file-explorer.js
chriseth: Please don't use setval, it is too sparsely documented.

chriseth:
This creates a gigantic padding for me:
padding

chriseth:
An example of how to retrieve the absolute path without having to traverse the dom: https://github.com/ethereum/browser-solidity/commit/b9d9224e692bee8524a91b4aaf962d6db0940743

src/app.js
```diff
@@ -320,6 +347,8 @@ var run = function () {
 
     currentFile = file
 
+    files.event.trigger('fileFocus', [file])
```
chriseth: I think only a module itself should be able to trigger an event. I know this is already the case in app.js, but we need to change this. Please contain such uses to app.js, i.e. replace this by calling `switchToFile` inside app.js (provided as a callback from app.js).

package.json
```diff
@@ -87,5 +78,12 @@
       "soljson.js"
     ],
     "parser": "babel-eslint"
+  },
+  "dependencies": {
```
yann300: why moving stuff to `dependencies` ?

yann300:
There is still the tabs bar and I think we can remove it now.
it is confusing to have both. And it seems that we can rename folder in this file-explorer.

src/app.js
serapath: every browser automatically creates `window.nameofid` as a global reference to the DOM element with the id `nameofid`, for example `<div id="nameofid">hello world</div>`

It's a comment to explain where the `window.treeview` in line 166 comes from - i think not every developer might now about this behavior.

src/app.js
```diff
@@ -156,15 +159,39 @@ var run = function () {
...
 
   // ----------------- editor ----------------------
-
   var editor = new Editor()
-
-  // ----------------- tabbed menu -------------------
+  // -------------- fileExplorer ------------------
+  // @NOTE HTML#ID creates window['<ID name>'] element as a global
+  // var api = fileExplorer(window.treeview, files, editor)
+  fileExplorer(window.treeview, files, editor)
```
serapath: * I'm generally avoiding `new` and work only with factory functions.
* I'm not sure what `var fileExplorer` should contain - currently `fileExplorer(..)` returns an empty api object.
* The fileExplorer currently needs to know it's container so it can properly update itself, so i would still pass the same three arguments to fileExplorer

src/node_modules/_file-explorer.js
serapath: ok.
Just to maybe again give the reason for why a manually maintained nested `node_modules` directory is useful.
* It allows to do non-path-relative require calls so a search&replace is easier in case of refactoring
* It is easier to `npm publish` internaly modules to npm or elsewhere in case they grow into stuff that is useful across projects
* It enables to add a custom `package.json` which can contain a `"browserify"` field that includes all the transforms browserify will use (e.g. babel-transforms)

In general, running browserify transforms globally with the `-g` flag can break stuff, because it might try to run transforms on files that are not prepared. Each module should know the transform's it needs and list them in the package.json. Browserify will respect that and only execute only the relevant transforms on each module when it's traversing the tree to bundle everything.

I can anyway put it in `src/app` because for now there are no custom transforms for this module.

But the reason for currently using `babelify` and then on top of this also using `babel` is, because of trouble with global transforms.

serapath: So i copy & paste the code from setval to the project?

I still need the logic of turning an array of pathnames into a nested object and `setval` does only that. I can put it into `src/app/setval.js` maybe? Would that be good?

src/node_modules/_file-explorer.js
serapath: I can highly recommend a manually managed `src/node_modules/` folder which even allows to specify the relevant transforms if necessary. It really is awesome in making sure modules dont know anything about each other.

Otherwise - the file-explorer module the way it's currently written works a bit the other way around. It treats `files` as some kind of database singleton that might or might not sync with a backend and is passed everywhere.

I did that, because it was already very very hard to have any kind of overview in the `app.js` file, because basically everything is going on in the `app.js` file with no way of knowing - other than by convention - what is accessing which other things.

`app.js` feels like a very polluted global space.

The `files` object and accessing it's api is so essential to file-explorer, that i feel it basically means copy/pasting it into `app.js`. ...not sure what to do about it... but maybe, what about, for example:

```js
// inside file-explorer component
function removeSubtree (files, path) {
  var allPaths = Object.keys(files.list())
  var remove = allPaths.filter(function keep (p) { return ~p.indexOf(path) })
  remove.forEach(function (p) { api.files.emit('remove', p) }) // instead of `files.remove(p)`
}
// or
var filenames
api.files.emit('list', function (list) {
  filenames = list
}) // instead of `var filenames = files.list()`
```
and then in `app.js`
```js
var api = fileExplorer() // or `new FileExplorer()`
api.on('list', function (cb) { cb(files.list()) })
api.on('remove', function (path, cb) { cb(files.remove(p)) })
```
... and so on.

would that work?

serapath: so the `api` would have certain mainly "read" events, like `'list'` and certain more "write" events, like `'remove'`, even though removing might still return a boolean, thus the callback `cb`.
**The above would be for communication initiated by `fileExplorer`**

**For communication initiated by `app.js` maybe something like the following would work:**
```js
// app.js
files.event.register('fileRenamed', function (oldName, newName) {
  api.emit('fileRenamed', [oldName, newName], function response () { })
})
```
this pattern could inform the `fileExplorer` instance about stuff.
While `fileRenamed` does not need any response data, some other commands custom to the file-explorer might need a response - at least the response could inform about error or success

src/app.js
serapath: After not passing the `files` object anymore this will anyway change.

Do you like the proposed api from above?

---
regarding the padding from above - which browser do you use? I can't reproduce this.

package.json
serapath: oh sry. i thought modules required in a js file are supposed to be `dependencies` and tooling used during development (e.g. http-server, browserify, ...) belongs to dev-dependencies.

src/app.js
axic: I wonder if we would like to use jQuery, `window.getElementById` or `window.<id>`, because currently it seems we do use a mixture of these.

jQuery (or a more lightweight alternative) seems to be the safest way given we don't need to reinvent the wheel for supporting different browsers.

axic:
@serapath would it be possible to rebase or cherry-pick changes instead of merge commits? It is harder to see commits with merges.

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+
+  setTimeout(function () { // infere first active element from tab bar
+    var activePath = document.querySelector('.file.active').firstChild.innerText
+    var el = getElement(activePath)
+    focus.call(el,{})
+  },0)
+
+  var api = { }
+
+  return api
```
axic: Would it be possible reorganising the file so that the helpers aren't after the return statement? I know it is valid, but doesn't feel intuitive.

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+  function focus (event) {
+    event.cancelBubble = true
+    if (focusElement === this) return
+    if (focusElement) focusElement.classList.toggle(css.hasFocus)
+    focusElement = this
+    focusElement.classList.toggle(css.hasFocus)
+    var filepath = getPath(focusElement).file
+    if (filepath) files.event.trigger('fileFocus', [filepath])
+  }
+  function hover (event) {
```
axic: Can you leave spacing (empty line) between functions?

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+  function editModeOff (event) {
+    if (event.type === 'blur' || event.which === 27 || event.which === 13) {
+      if (event.which === 13) {
+        event.preventDefault()
+        save = true
+      } else if (textUnderEdit !== this.innerText) var save = confirm(`
+        Do you want to rename?
+      `)
+      if (save) {
```
axic: In general, can you leave a bit more spacing within functions?  Separating logical blocks with an empty line can help during reviews.

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+      this.classList.add(css.rename)
+      this.focus()
+    }
+  }
+  function editModeOff (event) {
+    if (event.type === 'blur' || event.which === 27 || event.which === 13) {
+      if (event.which === 13) {
+        event.preventDefault()
+        save = true
+      } else if (textUnderEdit !== this.innerText) var save = confirm(`
```
axic: For this simple question does it needs to be a backtick?

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+  files.event.register('fileRenamed', function (oldName, newName) {
+    var el = getElement(oldName)
+    if (el) {
+      oldName = oldName.split('/')
+      newName = newName.split('/')
+      var index = oldName.reduce(function (idx, key, i) {
+        return oldName[i] !== newName[i] ? i : idx
+      }, undefined)
+      var newKey  = newName[index]
+      var oldPath = oldName.slice(0,index+1).join('/')
```
axic: Please have spacing within parameters and arithmetics.

All this should validate with `standard`, I thought it would disallow `(0,index+1)`.

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+    if (~classes.indexOf('hasFocus') && !this.getAttribute('contenteditable')) {
+      textUnderEdit = this.innerText
+      this.setAttribute('contenteditable', true)
+      this.classList.add(css.rename)
+      this.focus()
+    }
+  }
+  function editModeOff (event) {
+    if (event.type === 'blur' || event.which === 27 || event.which === 13) {
+      if (event.which === 13) {
```
axic: Please try defining the variable before the first use. Eventually I think we'd want to move on to ES6 (e.g. `let`).

src/node_modules/_file-explorer.js
axic: Also the `var save = ...` should be a new line, and the statement should be enclosed in brackets. `standard` should have caught this.

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+        event.preventDefault()
+        save = true
+      } else if (textUnderEdit !== this.innerText) var save = confirm(`
+        Do you want to rename?
+      `)
+      if (save) {
+        var newPath = getPath(this.parentElement.parentElement.parentElement)
+        newPath = newPath.folder || newPath.file
+        var oldPath = newPath.split('/')
```
axic: This seems like a good candidate for a helper (replace file name in a path).

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+  function editModeOff (event) {
+    if (event.type === 'blur' || event.which === 27 || event.which === 13) {
+      if (event.which === 13) {
+        event.preventDefault()
+        save = true
+      } else if (textUnderEdit !== this.innerText) var save = confirm(`
+        Do you want to rename?
+      `)
+      if (save) {
+        var newPath = getPath(this.parentElement.parentElement.parentElement)
```
axic: Is there a way to make this a bit more explicit? Untangling the meaning requires one to review the DOM hierarchy.

test-browser/helpers/contracts.js
```diff
@@ -21,7 +21,7 @@ function testContracts (browser, contractCode, compiledContractNames, callback)
       .clearValue('#input textarea')
       .click('.newFile')
       .setValue('#input textarea', contractCode, function () {})
-      .waitForElementPresent('.contract .create', 5000, true, function () {
+      .waitForElementPresent('.contract .create', 15000, true, function () {
```
axic: Which part of the code slowed this process down?

src/node_modules/_file-explorer.js
```diff
@@ -0,0 +1,292 @@
...
+      }
+    }),false)
+    return true
+  }
+}
+/******************************************************************************
+  HELPER FUNCTIONS
+******************************************************************************/
+function adaptEnvironment (el, focus, hover) {
+  var node1 = el.parentElement.parentElement.parentElement
```
axic: More expressive naming (vs. `node1` and `node2`) might be useful here. Is `node` referring to a tree node here?

src/node_modules/_file-explorer.js
yann300: This looks like what we are currently doing:
https://github.com/ethereum/browser-solidity/blob/master/src/app.js#L573
-> `debugAPI` contains only what is needed by the debugger.
the same here for `rendererAPI`
https://github.com/ethereum/browser-solidity/blob/master/src/app.js#L642

Except that we currently have to pass also events. Using your stuff would allow to mix API stuff and event as even the API is managed by callback in your proposal.

Basically, 
What we do here in app.js
https://github.com/ethereum/browser-solidity/blob/master/src/app.js#L642 
to
https://github.com/ethereum/browser-solidity/blob/master/src/app.js#L607

would be replaced by
```

var api = fileExplorer() // or `new FileExplorer()`
api.on('list', function (args, response) { response(null, files.list()) })
api.on('remove', function (args, response) { response(null, files.remove(args.path)) })

files.event.register('fileRenamed', function (oldName, newName) {
  api.emit('fileRenamed', { oldName, newName }, function response (err, data) { })
})

// ...
```

So the idea is interesting,
Have to think about it more ....
I think in a way I like having to pass either event and/or specific API needed to the ctor.

In your case you would have this huge block in app.js
```
var api = fileExplorer() // or `new FileExplorer()`
api.on('list', function (args, response) { response(null, files.list()) })
api.on('remove', function (args, response) { response(null, files.remove(args.path)) })

files.event.register('fileRenamed', function (oldName, newName) {
  api.emit('fileRenamed', { oldName, newName }, function response (err, data) { })
})

```


That's is still messy in a way.

I think there's (a lot of) room for improvement in in the actual remix code, we need to think about it more.
For example your solution does not clean app.js much more.
and I think that using the constructor is a good thing:
I would prefer to have 
``fileExplorer(apiDef, event1, event2)`` than 
``var api = fileExplorer()``
and the mixed api/event implementation below.

so roughly, your solution has the same idea but enforce the use of callback through this 'proxy' API.

I would keep like it is now in the debugger/renderer ...
``fileExplorer(apiDef, event1, event2)``
 - Registering to event1 and event2 in the ctor.
 - As for me, you are free to do as you feel for the apiDef (if you want to manage it by emitting/registering that's fine). That could be a first approach.

and also we should I think do that (in another PR): moving debugAPI, rendererAPI into another file, that'll free up app.js.

src/node_modules/_file-explorer.js
chriseth: As I said, setval as it currently is needs way more documentation and especially readable variable names. To be honest, I have no idea what it actually does and every developer should either be aware of what happens in the codebase or is able to learn it quickly.

src/app.js
chriseth: I'm fine with using factory methods, but it should be visible from their name that they are factory methods.

src/node_modules/_file-explorer.js
chriseth: This is something I would like to discuss during the next meeting. Until then, please keep to the current style.

src/node_modules/_file-explorer.js
chriseth: I agree that `app.js` should remain as simple as possible. The idea would be that `app.js` is just there to connect the endpoints between the modules. Please also keep in mind that it should be possible to use remix (and potentially other modules) in a different way than they are used currently. For example, if the whole system would be used together with an existing IDE, the existing IDE would provide the editor and we just have to connect the modules differently.

package.json
chriseth: since the build step produces a single `.js` file, there should be no runtime dependencies.

chriseth:
@yann300 I don't think we can remove the tabs already, this would require further changes in app.js and the editor, I think. In general, I would like this change to be as small an incremental step as possible.

src/node_modules/_file-explorer.js
serapath: i can improve the documentation.
it basically turns:

```js
var data = { }
;['a/b/c', 'a/b/d', 'x/y/z'].forEach(function (path) {
  setval(data, path, '')
})
console.log(data) // into
/*
 { a: { b: { c: '', d: '' }, x: { y: { z: '' } } }
*/
````
setval by default takes '/' as a seperator but it can also take a custom one (like ".") and splits the path and sets them on a nested object.

The readme of the npm module tries to show this via lots of examples

File: src/node_modules/_file-explorer.js (hunk 7 above)
Comment (serapath) [inline src/node_modules/_file-explorer.js]: ok - sure i can reorganize it.

File: src/node_modules/_file-explorer.js (hunk 8 above)
Comment (serapath) [inline src/node_modules/_file-explorer.js]: yes

File: src/node_modules/_file-explorer.js (hunk 9 above)
Comment (serapath) [inline src/node_modules/_file-explorer.js]: ok

File: src/node_modules/_file-explorer.js (hunk 10 above)
Comment (serapath) [inline src/node_modules/_file-explorer.js]: standard is ok with how it's currently written - it's all standardjs compliant

...but no problem,

template literals can also be a string.

yes ok can do that to with the brackets...

File: src/node_modules/_file-explorer.js (hunk 11 above)
Comment (serapath) [inline src/node_modules/_file-explorer.js]: ok, i add spacing.
I'll check with standard.

File: src/node_modules/_file-explorer.js (hunk 12 above)
Comment (serapath) [inline src/node_modules/_file-explorer.js]: sure, i'll check and fix it

File: src/node_modules/_file-explorer.js (hunk 13 above)
Comment (serapath) [inline src/node_modules/_file-explorer.js]: ok, i'll make it into a helper :+1:

File: test-browser/helpers/contracts.js (hunk 15 above)
Comment (serapath) [inline test-browser/helpers/contracts.js]: i'm not sure.
I got random fails when running the tests... sometimes stuff passes, sometimes it doesnt.
Putting it to 15 sec. stopped the problem for me

File: src/node_modules/_file-explorer.js (hunk 16 above)
Comment (serapath) [inline src/node_modules/_file-explorer.js]: ok, i'll do better names. sry.
It's mainly because of the TreeView component from `remix` that i'm using and i need to patch it here and there and they don't have good names, but i can try to make up names and use at least their html tags

File: package.json (hunk 5 above)
Comment (serapath) [inline package.json]: ok, will do

File: src/app.js (hunk 6 above)
Comment (serapath) [inline src/app.js]: ok

File: package.json [hunk 17]
```diff
@@ -4,46 +4,34 @@
...
+    "test": "standard; babel src --out-dir babelify-src; node test/index.js",
+    "selenium": "selenium-standalone start",
     "browser-test": "nightwatch --config nightwatch.js --env local",
     "browser-test-remote-chrome": "nightwatch --config nightwatch.js --env chrome",
     "browser-test-remote-firefox": "nightwatch --config nightwatch.js --env default",
     "browser-test-remote-ie": "nightwatch --config nightwatch.js --env ie",
     "browser-test-remote-safari": "nightwatch --config nightwatch.js --env safari",
     "browser-test-remote-parallel": "nightwatch --config nightwatch.js --env safari,ie,default,chrome",
-    "build": "mkdir build; browserify src/index.js -g yo-yoify -o build/app.js -t [ babelify ]; babel --plugins=transform-es2015-template-literals build/app.js --out-file build/app.js",
-    "csslint": "csslint --ignore=order-alphabetical --errors='errors,duplicate-properties,empty-rules' --exclude-list='assets/css/font-awesome.min.css' assets/css/",
```
serapath: ok, i'll put it to alphabetic order then. didnt realize - tried to sort it according to the order its mentioned in the scripts

src/app.js
```diff
@@ -156,15 +157,74 @@ var run = function () {
...
+          li.parentElement.removeChild(li) // delete tab
+        }
+      })
+      return files.remove(path)
+    },
+    filesIsReadOnly: function (path) { return files.isReadOnly(path) },
+    filesRename: function (oldPath, newPath) { return files.rename(oldPath, newPath) }
+  }
+  var api = new FileExplorer('#treeview', fileExplorerAPI, files.event)
+  api.on('focus', function (path) {
```
yann300: Please use the internal event library for that

src/app.js
```diff
@@ -156,15 +157,74 @@ var run = function () {
...
+    [...window.files.querySelectorAll('.file .name')].forEach(function (span) {
```
yann300: where is set `window.files` ? is that necessary to set `files` as a global var?

yann300: `[...` ?

package.json
```diff
@@ -10,59 +10,53 @@
...
+    "jquery": "^3.1.1",
     "js-base64": "^2.1.9",
+    "mkdirp": "^0.5.1",
     "nightwatch": "^0.9.3",
+    "notify-error": "^1.2.0",
+    "npm-run-all": "^4.0.2",
+    "onchange": "^3.2.1",
     "solc": "https://github.com/ethereum/solc-js",
-    "standard": "^8.5.0",
+    "standard": "^8.6.0",
```
yann300: it is not necessary to change the version I think (same for jquery, yo-yo)
@axic @chriseth opininon on that ?

package.json
```diff
@@ -10,59 +10,53 @@
     "browser-test-remote-ie": "nightwatch --config nightwatch.js --env ie",
     "browser-test-remote-safari": "nightwatch --config nightwatch.js --env safari",
     "browser-test-remote-parallel": "nightwatch --config nightwatch.js --env safari,ie,default,chrome",
-    "build": "mkdir build; browserify src/index.js -g yo-yoify -o build/app.js -t [ babelify ]; babel --plugins=transform-es2015-template-literals build/app.js --out-file build/app.js",
+    "build": "mkdirp build; browserify src/index.js -t yo-yoify -t [ babelify --presets [ es2015 ] ] | babili > build/app.js",
```
axic: `mkdirp` probably is not a good idea, for example macOS doesn't have it.

package.json
```diff
@@ -10,59 +10,53 @@
...
     "csslint": "csslint --ignore=order-alphabetical --errors='errors,duplicate-properties,empty-rules' --exclude-list='assets/css/font-awesome.min.css' assets/css/",
-    "downloadsolc": "rm soljson.js; wget https://ethereum.github.io/solc-bin/soljson.js",
-    "lint": "standard",
+    "downloadsolc": "rm -f soljson.js; wget https://ethereum.github.io/solc-bin/soljson.js",
```
axic: Can you make a separate PR for this change? It looks useful.

package.json
```diff
@@ -10,59 +10,53 @@
...
+    "babel-preset-es2015": "^6.22.0",
     "babelify": "^7.3.0",
+    "babili": "0.0.11",
     "brace": "^0.8.0",
     "browserify": "^13.0.0",
+    "browserify-livereload": "^1.0.9",
+    "csjs-inject": "^1.0.1",
     "csslint": "^1.0.2",
-    "ethereum-remix": "https://github.com/ethereum/remix",
+    "ethereum-remix": "git+https://github.com/ethereum/remix.git",
```
axic: I have seen both in different packages, but what is the reason to prefer one or the other?

package.json
axic: jQuery 2.2.0 to 3.1.1 is a breaking change and some code would need to be updated. Changing standard from 8.5.0 to 8.6.0 is not a breaking change, but not sure if it is relevant to this PR.

src/app.js
axic: Apparently `window.files` here is just an alias to `files`. I'd prefer to use `files` since that is used in the rest in the file.

axic: That is an ES6 feature: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Spread_operator

package.json
```diff
@@ -10,59 +10,53 @@
...
+    "lint": "standard | notify-error",
+    "make-mock-compiler": "node ci/makeMockCompiler.js",
+    "onchange": "onchange build/app.js -- npm run lint",
+    "prepublish": "npm-run-all -ls downloadsolc make-mock-compiler build",
+    "selenium": "selenium-standalone start",
     "serve": "http-server .",
-    "test": "standard; babel src --out-dir babelify-src; node test/index.js"
+    "start": "npm-run-all -lpr serve watch onchange",
+    "test": "standard; babel src --out-dir babelify-src; node test/index.js",
+    "watch": "mkdirp build; watchify src/index.js -p [ browserify-livereload --host 127.0.0.1 --port 1337 ] -t [ babelify --presets [ es2015 ] --no-sourceMaps ] -dv -o build/app.js"
```
axic: This also uses `mkdirp`.

.gitignore
```diff
@@ -1,7 +1,10 @@
 .vscode
 build
 node_modules
+!src/node_modules
```
axic: Do we want to keep this here?

axic:
I would put a virtual file called *Summary* as the root entry of the treeview, which displays the list of the files with the size in bytes of each.

src/app.js
serapath: oh, no. `window.files` is automatically set by the browser, and is a reference to the DOM element with the id `#files` - so i'm not defining it, i'm using it

serapath: yes. is that ok? :-)

package.json
serapath: wasnt even aware. i can undo, but as long as stuff runs through we are always on the "newest" version of standardjs, maybe not too bad :-)

package.json
serapath: `mkdirp` is a cross platform npm cli module.
So some systems, like windows maybe? ...do not have mkdir.
But `mkdirp` is a module written in javascript that can be installed from npm and is supposed to work cross platform

package.json
serapath: i had some problem related to deleting `soljson.js` and the `-f` flag solved it.

package.json
serapath: no reason, i just did an `npm install ethereum/remix --save-dev` and that's what npm did automatically.

package.json
serapath: same as above. it's a command line interfcace tool that is cross platform and can be installed via npm, so `npm install` will actually install `mkdirp` so that it will be available cross platform

.gitignore
serapath: oh sry - i can remove it. i will.

package.json
axic: Oh I didn't see you've included the `mkdirp` package. Can you please create a separate PR which makes use of `mkdirp`? (You see it is hard to review if so many changes are interleaved).

src/app.js
```diff
@@ -156,15 +157,74 @@ var run = function () {
   chromeCloudSync()
 
   // ----------------- editor ----------------------
-
   var editor = new Editor()
 
-  // ----------------- tabbed menu -------------------
+  // -------------- FileExplorer -------------------
+  var sources = {
```
axic: I assume this will be removed before merging?

src/app.js
```diff
@@ -156,15 +157,74 @@ var run = function () {
...
+    'src/ug/finance.sol': false,
+    'app/solidity/mode.sol': true,
+    'app/ethereum/constitution.sol': true
+  }
+  Object.keys(sources).forEach(function (key) { files.set(key, sources[key]) })
+
+  var fileExplorerAPI = {
+    filesList: function () { return Object.keys(files.list()) },
+    filesRemove: function (path) {
+      [...window.files.querySelectorAll('.file .name')].forEach(function (span) {
```
axic: Can you put a note here that this removes tabs as well as files? Perhaps it would make sense pulling out the internal function of "close tab" and use it here as well as in close tab event.

src/app.js
```diff
@@ -156,15 +157,74 @@ var run = function () {
...
+        if (span.innerText === path) {
+          var li = span.parentElement
+          li.parentElement.removeChild(li) // delete tab
+        }
+      })
+      return files.remove(path)
+    },
+    filesIsReadOnly: function (path) { return files.isReadOnly(path) },
+    filesRename: function (oldPath, newPath) { return files.rename(oldPath, newPath) }
```
axic: In case of rename, the appropriate tab must be updated too.

src/app.js
axic: In that case I would prefer `document.querySelectorAll('#files .file .name')` (or using jQuery).

axic:
Can you please merge related commits and rebase the tree against master? Perhaps also use `files.listAsTree`?

Would be nice to see how this looks like currently.

chriseth:
@serapath please rebase

axic:
@serapath needs to be rebased again since livereload was merged

src/app/files.js
```diff
@@ -2,12 +2,14 @@
...
 
+var FILENAME = 'Untitled'
+
 function Files (storage) {
   var event = new EventManager()
   this.event = event
   var readonly = {}
 
-  this.exists = function (path) {
+  this.exists = function isFile (path) {
```
axic: Why are these needed?

src/app/files.js
```diff
@@ -107,6 +109,20 @@ function Files (storage) {
     return files
   }
 
+  // @TODO: use this API
+  this.add = function addFile (filepath, content) {
```
axic: Why does this needs to be added here? I would prefer to keep it separate or only add the current method `createNonClashingName`.

src/app.js
```diff
@@ -159,26 +159,86 @@ var run = function () {
...
-    if (!el.hasClass('active')) {
-      el.parent().find('li').removeClass('active')
-      $('#optionViews').attr('class', '').addClass(cls)
-      el.addClass('active')
+  window.editor = editor
+
+  // ---------------- FilePanel --------------------
+  // /////////// TEMPORARY STUFF - START //////////////////
+  // @TODO: remove this test setup stuff
+  var sources = {
```
axic: I suppose this will be removed before merging?

src/app/files.js
serapath: oh this helped in the debugger to show not lots of anonymous functions but named ones.

src/app.js
serapath: yes

serapath:
@chriseth @axic @yann300 
I think I added everything from the feedback you gave.
One thing that's missing for now, but is not forgotten is:
>I would put a virtual file called Summary as the root entry of the treeview, which displays the list of the files with the size in bytes of each

I would add such things after the merge and other refinement features after the merge

serapath:
If I try everything in firefox and chrome manually it works.
I'm not sure why the tests break and what to do about it. Does any of you understand what the error messages are saying and can give me some hint. Spent the whole night debugging, but no success.

I guess I would need more details than just those "Details" that i get from the logs above

src/app.js
```diff
@@ -159,26 +155,83 @@ var run = function () {
...
-    if (!match) return
-    var cls = match[0]
-    if (!el.hasClass('active')) {
-      el.parent().find('li').removeClass('active')
-      $('#optionViews').attr('class', '').addClass(cls)
-      el.addClass('active')
+  // ---------------- FilePanel --------------------
+  /****************************************************************************
+    @TODO's
+      1. I would put a virtual file called Summary as the root entry of the treeview, which displays the list of the files with the size in bytes of each
```
chriseth: Please move this into an issue.

src/app.js
chriseth: Please add a simple function to the files module that will trigger the event. Triggering events of other modules should not be possible.

chriseth: Also, do we need `editorSyncFile` above, won't it be called from the event procedure later?

src/app/file-explorer.js
```diff
@@ -0,0 +1,310 @@
...
+function fileExplorer (appAPI, files) {
+  var fileEvents = files.event
+  var tv = new Treeview({
+    extractData: function (value, tree, key) {
+      var newValue = {}
+      // var isReadOnly = false
+      var isFile = false
+      Object.keys(value).filter(function keep (x) {
+        if (x === '/content') isFile = true
+        // if (x === '/readOnly') isReadOnly = true
```
chriseth: read only files are not yet supported?

src/app/file-explorer.js
```diff
@@ -0,0 +1,310 @@
...
+        if (x[0] !== '/') return true
```
chriseth: Will this create an error if `x == ''`?

yann300:
- remove a file using the file-explorer
- then click anywhere on the file explorer, the file is added again.

yann300:
with firefox 45:
item in file-explorer are stuck to the right (remove the flex attributes of the activeMode css class) seems to fix the issue

yann300:
sometimes two items are selected

src/app/file-explorer.js
serapath: no they are supported by the `files` object, but the file explorer is not treating them.
I was assuming that a read-only-file might still be removed from the file explorer.
Don't know maybe i add it to issues

src/app/file-explorer.js
serapath: nope.  `var x = ''; console.log(x[0]) // => undefined`

serapath:
@chriseth what do you mean don't fire events of other modules?
I'm not sure I understand

serapath:
@yann300 
I can't reproduce the `removing behavior` you describe nor the `layout oddity` you describe.
The class `.activeMode` is added/removed to make the onmouseover delete button work.

It shouldnt cause that kind of layout bugs you describe, ....
Can you try again with the newest pushed version and clear the cache and "hard reload" the page?

src/app.js
serapath: i did https://github.com/ethereum/browser-solidity/issues/506

serapath:
I do not understand the failing tests.
If i manually make a new file `Untitled` and add the contract code:
```
pragma solidity ^0.4.0;
      contract TestContract { function f() returns (uint) { return 8; } }
```
and then check `document.querySelector('.output .returned')` or just `document.querySelector('.output')`, the former returns `null` and the latter a node that contains stuff like:
```html
<div class="output"><div class="result"><div class="udapp-close"></div><div class="gasUsed"><strong>Transaction cost:</strong> 91937 gas. <div class="gasUsed"><strong>Execution cost:</strong> 29477 gas.</div></div><div class="debugTx"><button title="Launch Debugger" class="debug"><i class="fa fa-bug"></i></button></div></div></div>
```

This does not really feel related to the file explorer i have to say.

Anyone a good idea?

yann300:
you should rebase cause an important UI changes has been pushed yesterday.

serapath:
did the rebase

chriseth:
Persisting the current file was a change that was not necessary here. Since it is there we can keep it, but please keep in mind to keep every change as minimal as possible. We want to merge this PR. Any unnecessary change can introduce new bugs and test failures.

serapath:
Yes I see, but the change is really minimal and is re-using really "generic" packages, like the `Files` and `Storage` module and otherwise replaces existing code in a kind of straight forward way.
I understood from out last weekly meeting, that `ui state` should or could potentially be persisted in the same/similar way than files, thus it made sense to me.

But I agree, no more changes and let's get this finally merged. :-)
After this i will always make small pull requests. It's kind of unfortunate that this one became a bit bigger and takes so long.

src/app/storage.js
```diff
@@ -49,7 +49,7 @@ function Storage () {
   safeKeys().forEach(function (name) {
     if (name.indexOf('sol-cache-file-', 0) === 0) {
       var content = window.localStorage.getItem(name)
-      window.localStorage.setItem(name.replace(/^sol-cache-file-/, 'sol:'), content)
+      window.localStorage.setItem(name.replace(/^sol-cache-file-/, prefix), content)
```
axic: This is broken I think. `sol-cache-file-*` should only be renamed to `sol:`, it doesn't make sense under `sol-ui:` or any other prefix.

src/app/storage.js
```diff
@@ -40,7 +40,7 @@ function Storage () {
   this.keys = function () {
     return safeKeys()
       // filter any names not including sol:
-      .filter(function (item) { return item.indexOf('sol:', 0) === 0 })
+      .filter(function (item) { return item.indexOf(prefix, 0) === 0 })
       // remove sol: from filename
       .map(function (item) { return item.replace(/^sol:/, '') })
```
axic: This is not removing the right prefix.

PR #451, opened 2017-03-09
Use jQuery for all screen elements [WIP]

axic:
I had this lying around for a while.

### DIFF
--- a/src/app/compiler.js
+++ b/src/app/compiler.js
@@ -1,5 +1,7 @@
 'use strict'
 
+var $ = require('jquery')
+
 var solc = require('solc/wrapper')
 var solcABI = require('solc/abi')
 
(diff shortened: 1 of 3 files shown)


yann300:
tests are failing

axic:
> tests are failing

It also depends on #452. We'll fix once that is merged.

src/app/compiler.js
```diff
@@ -153,10 +155,11 @@ function Compiler (handleImportCall) {
       compilationFinished({error: 'Compiler not yet loaded.'})
     })
 
-    var newScript = document.createElement('script')
```
chriseth: I remember there might have been some magic involved and I specifically did not use jquery here, but I might also be mistaken.

axic: A very similar piece of code is used here too: https://github.com/ethereum/browser-solidity/blob/master/src/app.js#L212

It seems to work for me on Chrome, though that doesn't mean it works everywhere.

chriseth: It seems that jquery removes script tags from the dom after they were executed. So this is the difference between the two methods, but I don't know if that matters. Can you test the code with firefox?
The iframe example cannot be compared to this, I think.

axic:
@yann300 @ninabreznik @serapath I'm not sure how relevant this is now,  but I think it would be nice to be consistent. Can you review, update or close?

axic:
This is obsolete now.

PR #453, opened 2017-03-09
Replace example by the one from the documentation.

chriseth:


src/app/example-contracts.js
```diff
@@ -1,68 +1,147 @@
...
 
-    address chairperson;
-    mapping(address => Voter) voters;
-    Proposal[] proposals;
+    address public chairperson;
 
-    /// Create a new ballot with $(_numProposals) different proposals.
-    function Ballot(uint8 _numProposals) {
+    // This declares a state variable that
+    // stores a `Voter` struct for each possible address.
```
axic: Since the whole thing is enclosed in backticks, using them here will break.

PR #458, opened 2017-03-14
editorWindowSize can get negative. RHS bar not usable anymore.

soad003:
I you pull your RHS bar smaller and go over the browser window bounds editorWindowSize gets negative. The only way to restore the bar and to make the IDE usable again is to change the local storage directly.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -462,7 +462,7 @@ var run = function () {
       $('#ghostbar').remove()
       $(document).unbind('mousemove')
       dragging = false
-      delta = (delta<50)?50:delta
+      delta = (delta < 50) ? 50 : delta
       setEditorSize(delta)
       config.set(EDITOR_WINDOW_SIZE, delta)
       reAdjust()
--- a/src/app.js
+++ b/src/app.js
@@ -462,6 +462,7 @@ var run = function () {
       $('#ghostbar').remove()
       $(document).unbind('mousemove')
       dragging = false
+      delta = (delta<50)?50:delta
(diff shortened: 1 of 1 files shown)


src/app.js
```diff
@@ -462,6 +462,7 @@ var run = function () {
       $('#ghostbar').remove()
       $(document).unbind('mousemove')
       dragging = false
+      delta = (delta < 50) ? 50 : delta
```
axic: Perhaps a `if (delta < 0) return` would be enough here?

soad003: Yes, would be enough, but probably less user friendly because his action is always reset to the prior state if he leaves the window while dragging. Apart form that he could drag very close to 0 making it hard to make the bar appear again. For hidding it there is a button anyways.

chriseth: Actually for me, the editor does not shrink arbitrarily anyway. Can we somehow find out what the correct bound is?

soad003: You mean we bound the RHS to a value where it is still usable?

soad003: I have just checked the rhs bar does need about 315px. If smaller not all tabs are shown.

chriseth: Is that value font-size-independent?

chriseth: Oh I might have confused this. I was talking about the LHS.

soad003: OK, what do you want me to change now?

chriseth: Sorry, I think this is fine for now :-)

PR #461, opened 2017-03-14
Refactor remote imports

axic:
Based on #381.


axic:
Need to rebase after #381 is merged.

PR #462, opened 2017-03-14
Implement files.listAsTree()

axic:
I did this last week out of curiosity. I think we could make it a clean API when placing into files.

### DIFF
--- a/src/app/files.js
+++ b/src/app/files.js
@@ -107,6 +107,47 @@ function Files (storage) {
     return files
   }
 
+  //
+  // Tree model for files
+  // {
+  //   'a': { }, // empty directory 'a'
+  //   'b': {
+  //     'c': {}, // empty directory 'b/c'
+  //     'd': { '/readonly': true, '/content': 'Hello World' } // files 'b/c/d'
+  //     'e': { '/readonly': false, '/path': 'b/c/d' } // symlink to 'b/c/d'
+  //     'f': { '/readonly': false, '/content': '<executable>', '/mode': 0755 }
+  //   }
+  // }
+  //
+  this.listAsTree = function () {
+    function hashmapize (obj, path, val) {
+      var nodes = path.split('/')
+      var i = 0
+
+      for (; i < nodes.length - 1; i++) {
+        var node = nodes[i]
+        if (obj[node] === undefined) {
+          obj[node] = {}
+        }
+        obj = obj[node]
+      }
+
+      obj[nodes[i]] = val
+    }
+
+    var tree = {}
+
+    var self = this
+    storage.keys().forEach(function (path) {
+      hashmapize(tree, path, {
+        '/readonly': self.isReadOnly(path),
+        '/content': self.get(path)
+      })
+    })
+
+    return tree
+  }
+
   // rename .browser-solidity.json to .remix.config
   if (this.exists('.browser-solidity.json')) {
     this.rename('.browser-solidity.json', '.remix.config')


src/app/files.js
```diff
@@ -107,6 +107,47 @@ function Files (storage) {
...
   }
 
+  //
+  // Tree model for files
+  // {
+  //   'a': { }, // empty directory 'a'
+  //   'b': {
+  //     'c': {}, // empty directory 'b/c'
+  //     'd': { '/readonly': true, '/content': 'Hello World' } // files 'b/c/d'
+  //     'e': { '/readonly': false, '/path': 'b/c/d' } // symlink to 'b/c/d'
```
axic: Note, it doesn't support symlinks yet, nor mode. I have them here as a future suggestion we could implement.

serapath: I was wondering if a simple straight forward nested object would maybe be enough.  
If it really matters whether the structure can support all or lots of stuff normal file systems support, then maybe it should follow some kind of "standard".  

I was thinking a bit ... and in case i'll use the result of `files.listAsTree()` i have to refactor the `file-explorer` anyway... and was checking `npm` a bit and found:
* **https://www.npmjs.com/package/fs-file-tree** (my favourite so far regarding "inspiration")
* https://www.npmjs.com/package/files-tree
* https://www.npmjs.com/package/treeify-paths
* https://www.npmjs.com/package/fsify
* https://www.npmjs.com/package/dir-to-json
* ... maybe there are more examples ...

---

So what about leaning on `require('fs').stats` method and do something like:

```js
// representation of a single node
// (given that "" (empty string) as a property key is valid json)
var exampleNode = {
  "name": {
    "stats": {
      "type": 'directory', // one of: ['file', 'directory', 'blockdevice' , 'characterdevice', 'symlink', 'FIFO', 'socket']
      // ... and a list of all the properties supported by https://nodejs.org/api/fs.html#fs_class_fs_stats
    },
    "": { /* .... */ }
  }
}
```
assuming the "meta data" will be contained in the [.stats property]( https://nodejs.org/api/fs.html#fs_class_fs_stats),  
an example file system object could look like the following:

```js
var fs = {
  "a" : {
    "stats": {
      "type": "directory"
      // ...
    },
    "": { }
  },
  "b": {
    "stats": {
      "type": "directory"
      // ...
    },
    "": {
      "c": {
        "stats": {
          "type": "directory"
          // ...
        },
        "": { }
      },
      "d": {
        "stats": {
          "type": "file",
          "mode": "r"
          // ...
        },
        "": "Hello World"
      },
      "e": {
        "stats": {
          "type": "symlink",
          "mode": "rw"
          // ...
        },
        "": ["b", "f"]
      },
      "f": {
        "stats": {
          "type": "file",
          "mode": "rwx"
          // ...
        },
        "": "...maybe javascript content or base64 encoded stuff or whatever ..."
        // "": function () { console.log('executable stuff') }
      }
    }
  }
}

function get (fs, path, query) {
  query = query ? 'stats' : ''
  var node = { '': path }
  do {
    node = node[''].reduce((f, p, i, path) => (i !== path.length - 1) ? f[p][''] : f[p], fs)
  } while (node.stats.type === 'symlink')
  return node[query]
}

// usage
function getContent (path) { return get(fs, path) }
function getStats (path) { return get(fs, path, 'stats') }

var path = 'b/e'.split('/')

var stats = getStats(path)
var content = getContent(path)

```
If an executable node would ACTUALLY be content wrapped in a callable function, that would be super awesome i think :P ...would be so cool to use. And if you set random garbage to be executable, it would just throw, but that's fine, because making random garbage on a real file system executable also does not work.

axic: My reason for having `/readonly` and `/content` is that the forward slash is invalid as a key within that structure. Reserving `type` as a key to indicate the type on any given level would mean `type` cannot be a directory or file name.

serapath: it's just that i felt `/readonly` and `/mode` feel arbitrary. I'm used to `fs.stats(...)` so it feels "natural" and an empty space as key is invalid as a filename too...

in case of `b`, the "content" would directly be `fs.b`, but in case of `b/c`, the content would be `fs.b['/content'].c` ... if c is a file, but `fs.b.c` if c is a directory.

Where would you store "meta information" about directories?

axic: > but in case of b/c, the content would be fs.b['/content'].c ... if c is a file, but fs.b.c if c is a directory.

I don't entirely follow. `/content` is only valid on a file which must be a leaf node.

> Where would you store "meta information" about directories?

When we get to that point, very likely I would store them with a different prefix in the storage API (e.g. `sol-meta:` and not `sol:`)

axic: This is from node's `fs` module you've linked:
```
{
  dev: 2114,
  ino: 48064969,
  mode: 33188,
  nlink: 1,
  uid: 85,
  gid: 100,
  rdev: 0,
  size: 527,
  blksize: 4096,
  blocks: 8,
  atime: Mon, 10 Oct 2011 23:24:11 GMT,
  mtime: Mon, 10 Oct 2011 23:24:11 GMT,
  ctime: Mon, 10 Oct 2011 23:24:11 GMT,
  birthtime: Mon, 10 Oct 2011 23:24:11 GMT
}
```

I could imagine to have this structure under `/meta` or `/stat`.

serapath:
```js
var fs1 = {
  b: {
    '/readonly': '...'
    c: {
      '/readonly': '...'
      // c is an empty directory?
    }
  }
}

var fs2 = {
  b: {
    '/readonly': '...'
    c: {
      '/readonly': '...'
      '/content': 'c is a file'
    }
  }
}

var fs2 = {
  b: {
    '/readonly': '...'
    c: {
      '/readonly': '...'
      '/path': 'a'
    }
  },
  a: {
    // empty directory
  }
}
```

* when updating the structure, it's important to keep `/readonly` and `/mode` in sync.
* when updating the structure, it's important to take care, that a node does not have `/content` and `/path`
* if the `/path` or `/content` is set, then a node can still have "subdirectories", but they are hidden, because the object is not interpreted as a directory.
* iterating over a node in case it's a directory requires filtering out all properties that start with `/`

axic: > when updating the structure, it's important to keep /readonly and /mode in sync.

We do not yet have `/mode`. If we implement it, we may opt to remove `/readonly`.

> when updating the structure, it's important to take care, that a node does not have /content and /path

Likewise, `/path` (symlink) is not supported yet, so we're free to design it to fit.

I don't think currently `/readonly` would be valid anywhere, but leaf content nodes.

axic: Oh I think it might be confusing, because I still intend to use the underlying `rename`, `remove`, etc. with full path. After every change, `listAsTree` should be called by consumer to get an up to date view.

Now we could optimise this to have an internal representation in `Files` to avoid recalculation, if needed.

The consumer should have some means to compare two trees, but that is outside of files.

serapath: Actually i start to like the way you do stuff :-)
Maybe it could be added ad `/props` the way you do it now, so the initial file tree from above would look like:

```js
var fs = {
  "a" : { // empty directory
    "/type": "directory",
    "/dev": 2114,
    "/ino": 48064969,
    "/mode": 33188,
    "/nlink": 1,
    "/uid": 85,
    "/gid": 100,
    "/rdev": 0,
    "/size": 527,
    "/blksize": 4096,
    "/blocks": 8,
    "/atime": "Mon, 10 Oct 2011 23:24:11 GMT",
    "/mtime": "Mon, 10 Oct 2011 23:24:11 GMT",
    "/ctime": "Mon, 10 Oct 2011 23:24:11 GMT",
    "/birthtime": "Mon, 10 Oct 2011 23:24:11 GMT"
  },
  "b": {
    "/type": "directory",
    // ... and more meta as above ...
    "c": {
      "/type": "directory",
      // ... and more meta as above ...
    },
    "d": {
      "/type": "file",
      // ... and more meta as above ...
      "": "Hello World"
    },
    "e": {
      "/type": "symlink",
      // ... and more as above ...
      "": ["b"]
    },
    "f": {
      "/type": "file",
      // ... and more meta as above ...
      "": "...maybe javascript content or base64 encoded stuff or whatever ..."
      // "": function () { console.log('executable stuff') }
    }
  }
}

```

serapath: # another iteration with `""` as dedicated "content property"
```js
Actually i start to like the way you do stuff :-)
Maybe it could be added ad `/props` the way you do it now, so the initial file tree from above would look like:

```js
var fs = {
  "a" : {
    "/type": "directory",
    "/dev": 2114,
    "/ino": 48064969,
    "/mode": 33188,
    "/nlink": 1,
    "/uid": 85,
    "/gid": 100,
    "/rdev": 0,
    "/size": 527,
    "/blksize": 4096,
    "/blocks": 8,
    "/atime": "Mon, 10 Oct 2011 23:24:11 GMT",
    "/mtime": "Mon, 10 Oct 2011 23:24:11 GMT",
    "/ctime": "Mon, 10 Oct 2011 23:24:11 GMT",
    "/birthtime": "Mon, 10 Oct 2011 23:24:11 GMT",
    "": { } // empty directory
  },
  "b": {
    "/type": "directory",
    // ... and more meta as above ...
    "": {
      "c": {
        "/type": "directory",
        // ... and more meta as above ...
        "": { } // empty directory
      },
      "d": {
        "/type": "file",
        // ... and more meta as above ...
        "": "Hello World"
      },
      "e": {
        "/type": "symlink",
        // ... and more as above ...
        "": ["b"]
      },
      "f": {
        "/type": "file",
        // ... and more meta as above ...
        "": "function () { console.log(' this could be a function or a base64 encoded string ') }"
      }
    }
  }
}

```

serapath: @axic
So this would be basically your proposal, with the slight difference, that instead of having a
* `/content` or `/path` or no property at all in case of a directory
* there would instead always be a `""` property

This would simplify lookup.

1. There would be no need to iterate over the object to filter out `/properties`
2. I can always return the `""` property's value and treat it according to the `/type` property instead of switching between `/content` or `/path` and eventually more properties or the above mentioned filtering.

chriseth:
I will merge this now, if it has to be changed later, it can be changed easily, it is just an internal interface.

PR #484, opened 2017-03-21
Static analysis, this on local calls.

soad003:
First additional static analysis and preparation for more modules.


src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -0,0 +1,156 @@
...
+  SEND: buildFunctionSignature([basicTypes.UINT], [basicTypes.BOOL], false),
+  CALL: buildFunctionSignature([], [basicTypes.BOOL], true),
+  DELEGATECALL: buildFunctionSignature([], [basicTypes.BOOL], false)
+}
+
+var lowLevelCallTypes = {
+  CALL: { ident: 'call', type: basicFunctionTypes.CALL },
+  CALLCODE: { ident: 'callcode', type: basicFunctionTypes.CALL },
+  DELEGATECALL: { ident: 'delegatecall', type: basicFunctionTypes.DELEGATECALL },
+  SEND: { ident: 'send', type: basicFunctionTypes.SEND }
```
axic: Should also include `transfer`.

src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+var specialVariables = {
+  BLOCKTIMESTAMP: { obj: 'block', member: 'timestamp', type: basicTypes.UINT },
+  BLOCKHASH: {
+    obj: 'block',
+    member: 'blockhash',
+    type: buildFunctionSignature([basicTypes.UINT], [basicTypes.BYTES32], false)
+  }
```
axic: Should this also include `tx`, `msg`, etc.?

src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+function isLowLevelCall (node) {
+  return isLLCall(node) ||
+          isLLCallcode(node) ||
+          isLLDelegatecall(node) ||
+          isLLSend(node)
+}
+
+function isLLSend (node) {
```
chriseth: Oh, I hope this will be much easier in the future...

src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -0,0 +1,156 @@
...
+          undefined, basicTypes.ADDRESS, lowLevelCallTypes.CALLCODE.ident)
+}
+
+function isLLDelegatecall (node) {
+  return isMemberAccess(node,
+          utils.escapeRegExp(lowLevelCallTypes.DELEGATECALL.type),
+          undefined, basicTypes.ADDRESS, lowLevelCallTypes.DELEGATECALL.ident)
+}
+
+function isMemberAccess (node, retType, accesser, accesserType, memberName) {
```
chriseth: Typo: Accessor

src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -0,0 +1,156 @@
...
+        nrOfChildren(node, 1) &&
+        name(node.children[0], accesser) &&
+        returnType(node.children[0], accesserType)
+}
+
+function nrOfChildren (node, nr) {
+  return (node && (nr === undefined || nr === null)) || (node && node.children && node.children.length === nr)
+}
+
+function returnType (node, typeRegex) {
```
chriseth: Is it return type or rather type of expression itself?

src/app/staticanalysis/modules/thisLocal.js
```diff
@@ -0,0 +1,30 @@
...
+
+thisLocal.prototype.visit = function (node) {
+  if (common.isThisLocalCall(node)) this.warningNodes.push(node)
+}
+
+thisLocal.prototype.report = function (compilationResults) {
+  this.warningNowNodes = []
+  return this.warningNodes.map(function (item, i) {
+    return {
+      warning: `use of "this" for local functions: never use this to call local functions, it only consumes more gas than normal local calls.`,
```
chriseth: Please capitalize the first word of a full sentence (I would even capitalize the first word of a sequence of words that is not a full sentence).

Perhaps use "functions in the same contract" instead of "local functions" for the second occurance.

src/app/staticanalysis/modules/staticAnalysisCommon.js
soad003: Well i don't know in this special case low level calls are all calls that do not throw exceptions on failure. Transfer does not do that. Apart form that, i just included stuff that i needed so far.

src/app/staticanalysis/modules/staticAnalysisCommon.js
soad003: Yes would be the place to include that if needed in some kind of analysis. I just implemented what's needed right now. And there is more to come.

src/app/staticanalysis/modules/staticAnalysisCommon.js
soad003: You are right it is the expression type. I will change that.

src/app/staticanalysis/modules/thisLocal.js
soad003: Ok i will do that!

soad003:
Requested changes done!

src/app/staticanalysis/modules/staticAnalysisCommon.js
axic: If you only want to check those which do not throw, then `transfer` is excluded.

PR #494, opened 2017-03-29
Rename and restructure tabs and content in righthand panel

ninabreznik:
So, first batch of changes on the righthand panel is done. 
* renaming tabs
* merging the content 
  * for example in RUN we now have **transaction + execution**, 
  * in ANALYSIS **verification and static analysis** etc.

NEXT TODOS:
1. what is left for now is to divide CONTRACT tab into 2 parts: static contract part and the part with buttons and legend (this part will go under RUN tab. But in order to do that, we will start building **components** for each of the tabs so we can untangle the CSS, HTML and JS and join them nicely into separate components)
2. After that we can also prepare a little documentation for each tab and save it [Solidity documentation on Github] (http://remix.readthedocs.io/en/latest/) and link to it from the tabs

### DIFF
--- a/assets/css/browser-solidity.css
+++ b/assets/css/browser-solidity.css
@@ -178,8 +178,8 @@ body {
 
 #header #options li {
     float: left;
-    padding: 0.3em 0.6em;
-    font-size: 1em;
+    padding: 0.7em 0.3em;
+    font-size: .9em;
     cursor: pointer;
     background-color: transparent;
     margin-right: 0.5em;
@@ -217,6 +217,10 @@ body {
 }
 
 #header #optionViews.envView #envView {
+    display: block;
+}
+
+#header #optionViews.compilationView #compilationView {
     display: block;
 }
 
@@ -275,6 +279,19 @@ body {
 }
 
 #publishView button {
+    background-color: #C6CFF7;
+    font-size: 12px;
+    padding: 0.25em;
+    margin-bottom: .5em;
+    color: black;
+    border:0 none;
+    border-radius: 3px;
+    width: 8em;
+    margin-right: 1em;
+    cursor: pointer;
+}
+
+#compilationView button {
     background-color: #C6CFF7;
     font-size: 12px;
     padding: 0.25em;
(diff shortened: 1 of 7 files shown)


ninabreznik:
screen shot 2017-03-29 at 8 52 20 pm
screen shot 2017-03-29 at 8 52 12 pm

yann300:
cool! that's definitely better without icons

yann300:
tab `Contract` when the block is minimized, everything is hidden so you cant reexpand it

yann300:
when the mode is changed (for instance from `JavaScript VM` to `Web3 provider`, the tab automatically swtch to the `Contract` tab. is that something wanted?

yann300:
in `Contract` tab, I would not display `Bytecode`, `Interface`, `Metadata location` , by default. it takes too much space.

yann300:
compilation error are now in `Run` tab, I am not sure it is visible enough.
If there is a compilation error, we should replace the content of the `Contract` tab by the list of errors

yann300:
tab `Analysis` , there should be a separation between formal verification and static analysis.
But might be better to do that in another PR.

yann300:
I really think we should merge the `Contract` and `Run` tab .
In order to shrink a little bit the `Run` tab so the content can fit better in the `Contract` tab, we could replace this 3 check box of the running mode by a dropdownlist - and that would be consistent with the rest of the content (Transaction  origin, ...)

yann300:
tests are failing: 
```
> browser-solidity@0.0.0 lint /home/travis/build/ethereum/browser-solidity
> standard
standard: Use JavaScript Standard Style (http://standardjs.com)
  /home/travis/build/ethereum/browser-solidity/src/app/renderer.js:216:7: 'detailsOpen' is assigned a value but never used.
```

ninabreznik:
@chriseth @LianaHus This pull request can be closed. I created a new one. Thank you!

yann300:
replaced by https://github.com/ethereum/browser-solidity/pull/496

PR #495, opened 2017-04-03
Rename and restructure tabs and content in righthand panel

ninabreznik:
Pull request 949 can be deleted. I messed up something because I wanted to use live reload so I forked @serapath's repo and deleted old one. Thought it will automagically update if I call repos and branches the same, but nope.

So, here is the first draft of right hand panel. 
* Icons are replaced, 
* we have a new tab for contract,
* content of the tabs is a bit bigger and stretched and design is slightly tightened

TODO:
* review the UI
* decide how to nicely redesign it for real

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,3 +1,5 @@
+<!DOCTYPE html>
+
 <html>
 	<head>
 		<meta charset="utf-8">


ninabreznik:
Here are snapshot of @yann300 's comments from previous pull request

screen shot 2017-04-04 at 3 34 33 am

screen shot 2017-04-04 at 3 34 42 am

screen shot 2017-04-04 at 3 34 50 am

PR #498, opened 2017-04-05
use solc version

yann300:


axic:
Why do we need to change this?

yann300:
just to use the same declaration as 
https://github.com/ethereum/remix/blob/master/package.json#L33
also with
`"solc": "https://github.com/ethereum/solc-js",`
soljson.js is not fetched

axic:
> just to use the same declaration as https://github.com/ethereum/remix/blob/master/package.json#L33

0.4.3 is quite outdated

> soljson.js is not fetched

browser-solidity doesn't need it from the npm package. Does remix needs it?

yann300:
@soad003 (static analysis) needs it for his test....

chriseth:
Sorry, I still don't understand. Why use `^0.4.3`? Why does it make a difference if we use npm or not? Because the prepublish step in solc-js downloads soljson? Why isn't it downloaded by install inside solc-js?

soad003:
@chriseth i am no expert but if you use github directly you have a problem with breaking changes in dependencies, whenever you change the repo (in this case solc-js) you have to immediately change all repos that depend on it (same problem as we had in the build toolchain change in ethereum-remix where the repo was already changed but the PR in browser-solidity was not yet merged) 

The problem with solc-js is that it depends upon soljson.js which is downloaded by the solc-js prepublish script. I guess if you deploy a version to NPM you run that first so soljson.js is included in the package.  If you refer to the npm dependency with soljson.js included everything is fine because it downloads soljson.js with the package so the code works. 

And yes soljson.js is actually downloaded in the solc-js project if you run install via the prepublish script. But the prepublish script is only run for the current project not for all dependent projects. That means if you refer to solc-js from browser-solidity and you run npm install in the browser-solidity folder only the prepublish script of browser-solidity is run not the ones of the dependencies (including solc-js) they are just downloaded as is (from either github or npm).  In the github repo is no soljson.js file so we have just downloaded a broken dependency. 

This article does a good job at explaining the npm prepublish step: https://blog.greenkeeper.io/what-is-npm-s-prepublish-and-why-is-it-so-confusing-a948373e6be1

I you don't want to use packaged versions of solc-js over npm (with the soljson.js file included) then another possibility would be to just commit the soljson.js file to the solc-js repo. The it would also be downloaded into the project and the dependency should work. 

I would argue for the npm version because it would make it far less painfull to develop the projects in a different pace at the moment it is pretty easy break browser-solidity be changed some repo linked over github. 

Hope this is not to confusing...

chriseth:
Ok, now I understand, thanks!
If we specify 0.4.3 here, it will still download the latest version, right? Should we still state a newer version?

axic:
> In the github repo is no soljson.js file so we have just downloaded a broken dependency.

We do not rely on the soljson.js from the solc-js package. It is downloaded separately from solc-bin.

axic:
Btw, I'm always for using versioned imports and using `^0.4.3` should be fine as that allows anything up to `0.5` (reminder: are the latest Solidity versions pushed on npm?), however I think the explanation above is misleading as it is still not clear what bug is this fixing?

soad003:
@axic the bug is that if you fetch solc-js over github there is not soljson.js file in the source downloaded (which is needed to run solc-js). This was not a problem so far because browser-solidity downloads soljson.js by itself upon loading. So up to now it wasn't a problem. I am currently developing some static analysis modules for browser-solidity and i use solc-js in my integration tests (compile example source with known problems and check if static analysis finds the problems as expected). My tests always fail with the error that there is no soljson.js file in my solc node dependency folder. So this is not a bug fix for an issue in the current code but preparation for my modules. Does this make sense?

axic:
I think you shouldn't use `require('solc')` but `require('solc/wrapper')` similarly to how it is used in browser-solidity in order to have the flexibility of choosing a compiler version - probably the very same selected in browser-solidity.

soad003:
@axic ok sounds good to me, sorry i am pretty new to all this. I am now using the soljson.js file that is downloaded upon `npm install`  in the browser-solidity project root via the solc/wrapper (so soljson.js can be updated easily and independent of other stuff). It is still not the same version as used in the browser because it downloads the soljson.js every time and the soljson.js i use is downloaded once the programmer installs the project or whenever he runs `npm run downloadsolc`. Still there is a problem using solc-js as dependency via github because it  does not download all files to run properly... but maybe i miss something here. But thanks for pointing out the wrapper thing. makes my life easier!

chriseth:
@axic please merge this if you are ok with it.

axic:
@soad003 you've said this is only need for integration testing? That should be done on Travis and that involves downloading the latest version from solc-bin.

During runtime in the browser it should use the same version as selected by the user if I'm not mistaken?

soad003:
@axic at the moment they run with the normal unit tests but technically they are integration tests. But i haven't found a separation between integration and unit tests in the project... But if travis loads the latest version of soljson.js and runs tests then the local version is not that much of a problem, you are right.

Anyways my problems are solved so far by using the wrapper and the local soljson.js version. I just wanted to point out a problem i ran into...

axic:
@soad003 is this still a problem? Are you using this patch locally?

soad003:
@axic no from my point of view problems are fixed by using 'require('solc/wrapper')'. And all the static analysis modules have been moved to the remix repo in the meantime.

Still the other issue we discussed here was that with direct github refs it is easy to break browser-solidity by introducing breaking changes in the solc-js repo (therefore the change in this pull). 
 
https://github.com/ethereum/remix/blob/master/remix-solidity/package.json
Remix uses versioned reference to the npm package instead.

yann300:
if we specify `^0.4.3` , it will not necessary get the latest version, we could specify `latest`

yann300:
closing solc dependency has been moved to the remix repo

PR #499, opened 2017-04-05
Change node version (Travis)

yann300:


yann300:
@soad003 will include that in its own PR

PR #508, opened 2017-04-11
Static Analysis: Assembly, Checks Effetct Interaction, Constant Functions without modifiers.

soad003:


chriseth:
I will try to remove the "foreign" commits.

chriseth:
Rebased, I hope I did not miss any changes.

src/app/staticanalysis/modules/checksEffectsInteraction.js
```diff
@@ -0,0 +1,83 @@
...
+var desc = 'Avoid potential reentrancy bugs'
+var categories = require('./categories')
+var common = require('./staticAnalysisCommon')
+var callGraph = require('./functionCallGraph')
+var AbstractAst = require('./abstractAstView')
+
+function checksEffectsInteraction () {
+  this.contracts = []
+
+  checksEffectsInteraction.prototype.visit = new AbstractAst().builder(
```
pirapira: Now I'm wondering which object is `checkEffectsInteraction.prototye`.  I'm also wondering if this assignment can create a new property in that object.

yann300: `checksEffectsInteraction.prototype.visit` will contain a function definition
https://github.com/ethereum/browser-solidity/pull/508/files/c3d1bb226ddb92301d2dd8f399e684c81528b774#diff-29e9831d84a60c99282b9a110c9651aaR14

src/app/staticanalysis/modules/checksEffectsInteraction.js
```diff
@@ -0,0 +1,83 @@
...
+    (node) => common.isInteraction(node) || common.isEffect(node) || common.isLocalCallGraphRelevantNode(node),
+    this.contracts
```
chriseth: Is it possible to return the list instead of passing an object in that is modified?

src/app/staticanalysis/modules/constantFunctions.js
```diff
@@ -0,0 +1,88 @@
...
+var desc = 'Check for potentially constant functions'
+var categories = require('./categories')
+var common = require('./staticAnalysisCommon')
+var callGraph = require('./functionCallGraph')
+var AbstractAst = require('./abstractAstView')
+
+function constantFunctions () {
+  this.contracts = []
+
+  constantFunctions.prototype.visit = new AbstractAst().builder(
```
soad003: ehm i had a weird issue after the source was bablified, i could not pass this.contracts because this was not defined. It worked without babel... this was more or less a quick and dirty fix.

src/app/staticanalysis/modules/constantFunctions.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+constantFunctions.prototype.report = function (compilationResults) {
+  var warnings = []
+  var hasModifiers = this.contracts.some((item) => item.modifiers.length > 0)
+
+  var cg = callGraph.buildGlobalFuncCallGraph(this.contracts)
+
+  this.contracts.forEach((contract) => {
+    if (!common.isFullyImplementedContract(contract.node)) return
```
soad003: Oh yes you are right. I havn't thought about partially implemented contracts...

src/app/staticanalysis/modules/checksEffectsInteraction.js
soad003: how would you return a list, from which function? visit has no return value it is ignored by the static analysis runner.

soad003:
@chriseth  thanks for cleaning up my commit mess. Do you have any tips to avoid this mess the next time?

chriseth:
@soad003 I don't know how you ended up with that, so sorry, no tips ;-)

src/app/staticanalysis/modules/checksEffectsInteraction.js
chriseth: Oh I see, did not fully follow how that was used :-)

soad003:
@chriseth What is the current status with this PR? I am not in a hurry, just interested.

chriseth:
@soad003 thanks for asking - I'm still digesting, but I want to merge this as soon as possible. Did you find a way to fix the prototype assignment issue? And could you improve the check for contracts that have some unimplemented functions?

soad003:
@chriseth i haven't searched for a better solution, concerning prototypes. And yes i can improve the handling of contracts with some unimplemented functions. I also have some more modules but i thought it would be better to first wait until this PR is merged and apply the new stuff on a fresh base (to avoid PR with chaotic commit histories)

yann300:
perhaps using es6 class instead of prototype....

src/app/staticanalysis/modules/abstractAstView.js
```diff
@@ -0,0 +1,93 @@
...
+  this.currentFunctionIndex = null
+  this.currentModifierIndex = null
+  this.isFunctionNotModifier = false
+}
+
+abstractAstView.prototype.builder = function (relevantNodeFilter, contractsOut) {
+  this.contracts = contractsOut
+  return function (node) {
+    if (common.isContractDefinition(node)) {
+      setCurrentContract(this, {
```
chriseth: Better not use `this` here, we do not know how the function will be called.

soad003: Oh, yeah you mean like insert a `var that = this` and make a closure over that in the returned function?

src/app/staticanalysis/modules/functionCallGraph.js
```diff
@@ -0,0 +1,82 @@
...
+    var getNodeIdent = (node) => { return common.getFullQualifiedFunctionCallIdent(contract.node, node) }
+    var getFunDefIdent = (funcDef) => { return common.getFullQuallyfiedFuncDefinitionIdent(contract.node, funcDef.node, funcDef.parameters) }
+
+    callGraph[common.getContractName(contract.node)] = { contract: contract, functions: buildLocalFuncCallGraphInternal(contract.functions, filterNodes, getNodeIdent, getFunDefIdent) }
+  })
+
+  return callGraph
+}
+
+function analyseCallGraph (cg, funcName, context, nodeCheck) {
```
chriseth: I think this also deserves some documentation.

src/app/staticanalysis/modules/functionCallGraph.js
```diff
@@ -0,0 +1,82 @@
...
+
+  return combinator(current.node.relevantNodes.reduce((acc, val) => combinator(acc, nodeCheck(val, context)), false),
+                        current.calls.reduce((acc, val) => combinator(acc, analyseCallGraphInternal(cg, val, context, combinator, nodeCheck, visited)), false))
+}
+
+function resolveCallGraphSymbol (cg, funcName) {
+  return resolveCallGraphSymbolInternal(cg, funcName, false)
+}
+
+function resolveCallGraphSymbolInternal (cg, funcName, silent) {
```
chriseth: Does this take overloads into account?

src/app/staticanalysis/modules/constantFunctions.js
```diff
@@ -0,0 +1,88 @@
...
+function getStateVariables (contract, func) {
+  return contract.stateVariables.concat(func.localVariables.filter(common.isStorageVariableDeclaration))
+}
+
+function checkIfShouldBeConstant (startFuncName, context) {
+  return !callGraph.analyseCallGraph(context.cg, startFuncName, context, isConstBreaker)
+}
+
+function isConstBreaker (node, context) {
+  return common.isWriteOnStateVariable(node, context.stateVariables) ||
```
soad003: send is included, i will try to include transfer aswell (forgot about that). What does f.value(x)() mean is this a local call that sends money to itself? And i mean if it is payable then it can't be const right because it changes the balance (the compiler throws an error that this is not possible). 
About new was not sure because it does not change the state of the current contract but the blockchain. If one would invoke new and safe the returned address to a storage field than it is a const breaker because of the write on state variable condition. So the only case not covered are contracts that create other contracts without saving the address of the new contract ... how should we handle that?

src/app/staticanalysis/modules/functionCallGraph.js
soad003: yes all functions are handled via their full qualified name e.g. contractName.functionName(parameter1Type, parameter2Type)...

src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -17,7 +28,12 @@ var basicTypes = {
 
 var basicRegex = {
   CONTRACTTYPE: '^contract ',
-  FUNCTIONTYPE: '^function \\('
+  FUNCTIONTYPE: '^function \\(',
+  EXTERNALFUNCTIONTYPE: '^function \\(.*\\).* external',
+  CONSTANTFUNCTIONTYPE: '^function \\(.*\\).* constant',
+  REFTYPE: '( storage )|(mapping\\()|(\\[\\])',
```
soad003: I include the spaces because somebody could name fields storage e.g. the type would be `struct KingOfTheEtherThrone.storage storage ref` without the spaces it would match the fields name. I don't have this problem with mappings because nobody can use parentheses in variable names.

src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -26,6 +42,17 @@ var basicFunctionTypes = {
...
 
+var builtinFunctions = {
+  'keccak256()': true,
+  'sha3()': true,
+  'sha256()': true,
+  'ripemd160()': true,
+  'ecrecover(bytes32,uint8,bytes32,bytes32)': true,
+  'addmod(uint256,uint256,uint256)': true,
+  'mulmod(uint256,uint256,uint256)': true,
+  'selfdestruct(address)': true
```
soad003: At the time i wrote this the compiler did not support revert or assert. But i will add it. Whats the complete signature? revert() i guess?

src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -42,24 +69,237 @@ var specialVariables = {
...
+
+// #################### Complex Getters
+
+function getFunctionCallType (func) {
+  if (!(isExternalDirectCall(func) || isThisLocalCall(func) || isSuperLocalCall(func) || isLocalCall(func) || isLibraryCall(func))) throw new Error('staticAnalysisCommon.js: not function call Node')
+  if (isExternalDirectCall(func) || isThisLocalCall(func) || isSuperLocalCall(func) || isLibraryCall(func)) return func.attributes.type
+  return findFirstSubNodeLTR(func, exactMatch(nodeTypes.IDENTIFIER)).attributes.type
+}
+
+function getEffectedVariableName (effectNode) {
```
soad003: you mean every getter? or all the functions exported?

src/app/staticanalysis/modules/abstractAstView.js
```diff
@@ -0,0 +1,93 @@
...
+
+function abstractAstView () {
+  this.contracts = null
+  this.currentContractIndex = null
+  this.currentFunctionIndex = null
+  this.currentModifierIndex = null
+  this.isFunctionNotModifier = false
+}
+
+abstractAstView.prototype.builder = function (relevantNodeFilter, contractsOut) {
```
soad003: OK i will comment a bit ;) , Yes those are the nodes that are relevant for your analysis. For example if you are only interessted in variable declarations within functions you can filter here to make the analysis code easier

soad003:
@chriseth I am done with the changes you requested. Do you have any suggestion how to include them in this PR without messing up the history again? if i just push to my branch what happens to this PR where you rewrote the git history?

pirapira:
@soad003 This week @chriseth is on vacation.  One safe way to build a clean history is to create a new branch from `develop` and cherry-pick relevant commits one by one.  There might be more reasonable ways, depending on the shape of the Git history on your machine.

chriseth:
@soad003 if you are afraid of destroying the history, you can push to a different branch and I can cherry-pick your new commits.

src/app/staticanalysis/modules/staticAnalysisCommon.js
chriseth: Yes, `revert()`, `require(bool)` and `assert(bool)`.

src/app/staticanalysis/modules/constantFunctions.js
chriseth: `f.value(x)()` is not possible, but `c.f.value(x)()` is - is that handled? But you are probably right about payable and const.
Creating a new contract should not be a constant operation, because it modifies the state of the blockchain (even if it does not modify the current contract's state). constant functions also cannot call non-constant functions on other contracts (not only because they could themselves again call non-constant functions on the current contract).

src/app/staticanalysis/modules/abstractAstView.js
chriseth: yes

src/app/staticanalysis/modules/constantFunctions.js
soad003: Yes value and gas modifiers are also handled, this would count as external direct call, which is a constbreaker. Concerning new i will add that immediately ;)

soad003:
@chriseth ok i pushed to staticanalysis_integration, the two last commits are relevant (24c642d60ed1a596d7a742a5a6bf5d5737521671, 0c3b527b987633b4ef59311673e8e17a5e50964c). Sorry for the effort, i have to learn to deal with more complex git scenarios asap.

src/app/staticanalysis/modules/checksEffectsInteraction.js
chriseth: Can you try to move this out now?

src/app/staticanalysis/modules/constantFunctions.js
chriseth: Also here, can you try to move it out?

src/app/staticanalysis/modules/functionCallGraph.js
```diff
@@ -16,6 +16,30 @@ function buildLocalFuncCallGraphInternal (functions, nodeFilter, extractNodeIden
...
 
+/**
+ * Builds a function call graph for the current contracts.
+ * Example Contract call graph:
+ *
+ * {
+ *  "KingOfTheEtherThrone": {
+ *    "contracts": {...},                                        // Contract node as defined in abstractAstView.js
+ *    "functions": {
+ *      "KingOfTheEtherThrone.claimThrone(string memory)": {    // function in KingOfEtherThrone
```
chriseth: Please note that contract names are not unique, they are only unique if you prefix them with their absolute source unit path. So if this is a correct example of a fully qualified function name, those will not be unique.

src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -50,7 +51,9 @@ var builtinFunctions = {
   'ecrecover(bytes32,uint8,bytes32,bytes32)': true,
   'addmod(uint256,uint256,uint256)': true,
   'mulmod(uint256,uint256,uint256)': true,
-  'selfdestruct(address)': true
+  'selfdestruct(address)': true,
+  'revert()': true,
+  'assert()': true
```
chriseth: `assert` takes a `bool`.

chriseth:
I added your commits and rebased. There were conflicts in the descriptions of the modules, I resolved them and hope the tests will pass.
Thanks for the comments in the code, I think it will be much better maintainable now!
I hope I can finish the review this week.

chriseth:
Oh wow, that was fast! I'm through! :-)
So I would say the only things left are trying to properly fix the prototype hack, adding file name prefixes and the `new` operator and then this is good to go!

soad003:
I am glad to hear that!  new is already included! I will look for a good solution for to prototype hack today, and i will add the file name prefixes!

src/app/staticanalysis/modules/functionCallGraph.js
soad003: Are you sure about that? I just tried to define two contracts with the same name... got an Error that the identifier is already declared. Can you give me an example of where this would work\make sense?

soad003: I also tried it via import. I noticed that in `compilationResult.sources` the files are not ordered in source order (imports first, then the current file). At the moment the static analysis modules rely on that if you inherit from a contract it is defined before some other contract inherits form it (if you only use one file this is actually enforced by the compiler). Is there an easy way to change the order in the source array such it reflects source order?

soad003: @chriseth: have you seen my comment?

chriseth: Oh I'm sorry, I did not see the comment.
Every source file is a scope. You cannot have two contracts of the same name in the same scope, but you can have them in different scopes. Example:

file1:

`contract c{}`

file 2:

`import "file1" as x; contract c{}`

The import is only required for remix so that it also pulls the other file into the compiler. If you only use `import "file1";`, the import will create an identifier clash in the scope of file2, but if you add `as x`, this can be avoided.

The sources are not in source order because there is no real source order and there is no "current file". What the compiler does in this case is that it tries to order the sources according to imports, but that also regularly fails. Can you do such an analysis in your part, too?

soad003: Thanks for the reply. Ok, I see. Concerning the source code order. At the moment i only need the order of contracts because if some contract inherits from another i copy the state variables from the base contract. I can quick fix it by postpone the logic to run after i built all contract objects. The whole thing with import aliases is probably a bit more complicated and i don't have all the info in the AST at the moment, this is the AST node returned by the compiler for the import `{"attributes":{"file":"file1"},"id":3,"name":"ImportDirective","src":"0:20:1"}` the alias is missing. That means i can not resolve stuff like `contract c is x.c {}` because the alias is not available in the AST. For now i could produce warnings if i spot two contracts with the same name, that analysis does not support that at the moment(should be a rare case anyways...), if the AST contains all the info we can think of implementing this properly but i think it would take some time. What do you think?

chriseth: Yes, this sounds good. The new ast even contains pointers from each identifier to the correct declaration, so you do not even need to resolve that yourself.

soad003: @chriseth: commit (10ed4f381ba13ad5fadbb58660c8163c53833bba)

chriseth:
@soad003 so in the end you decided to rather warn if there are multiple contracts with the same name instead of adding a prefix? This is also fine, I just want to know :-)
Is there anything left to do here?

soad003:
@chriseth the prefix would have been much more work. And without the possibility of properly handle the import aliases it does not make a difference in both cases i am fine as long as no to contracts have the same name. And in both cases i can't handle the situation when two contracts have the same name in one case because i have two contracts with the same name in the other because i can not properly resolve the alias.

No I nothing left here as far as i can tell. 
Open in general are the modifier support and import aliases
I have 4 more modules i want to push concerning:
-blockhash
-timestamp
-lowlevelcalls
-similar variable names (as described in https://github.com/ethereum/browser-solidity/issues/109#issuecomment-295380808)

But I think it is easier to make a seperate PR.

PR #517, opened 2017-04-13
Do not autocompile when switching tab, if we are in a debug session

yann300:


chriseth:
Could we instead make all files read-only when the debugger is active?

yann300:
that's another thing no? we anyway need to disable compilation if debugger is active.

chriseth:
ok!

PR #519, opened 2017-04-18
Do not alert user if config set from the url param

yann300:


src/app/execution-context.js
```diff
@@ -87,32 +86,24 @@ function ExecutionContext () {
...
-
-  if (web3.providers && web3.currentProvider instanceof web3.providers.IpcProvider) {
-    $web3endpoint.val('ipc')
+    executionContextChange(context, endPointUrl)
   }
 
-  function executionContextChange (context) {
-    if (context === 'web3' && !confirm('Are you sure you want to connect to a local ethereum node?')) {
+  function executionContextChange (context, endPointUrl) {
+    if (context === 'web3' && !confirm('Are you sure you want to connect to an external ethereum node?')) {
```
chriseth: External sounds like it will not be on the same computer. Perhaps just say `to an Ethereum node`?

yann300: done

PR #520, opened 2017-04-19
Remove dependencies entry

yann300:


yann300:
`babel-plugin-transform-object-assign` is used by remix

PR #524, opened 2017-04-19
File explorer cleanup

chriseth:
This fixes a bug so that readonly files are actually visible in the file explorer. It also improves the sync between the file tabs and the tree view, but we are not 100% there yet.

I added some TODOs which should be addressed in future PRs.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -63,19 +63,12 @@ window.addEventListener('message', function (ev) {
   }
 }, false)
 
-/*
-  trigger tabChanged
-*/
 var run = function () {
   var self = this
   this.event = new EventManager()
   var fileStorage = new Storage('sol:')
   var files = new Files(fileStorage)
   var config = new Config(fileStorage)
-  var uiStorage = new Storage('sol-ui:')
-  var ui = new Files(uiStorage)
-
-  ui.set('currentFile', '')
 
   // return all the files, except the temporary/readonly ones
   function packageFiles () {
@@ -184,56 +177,57 @@ var run = function () {
   var editor = new Editor(document.getElementById('input'))
 
   // ---------------- FilePanel --------------------
+  // TODO: All FilePanel related CSS should move into file-panel.js
+  // app.js provides file-panel.js with a css selector or a DOM element
+  // and file-panel.js adds its elements (including css), see "Editor" above
   var css = csjs`
     .filepanel    {
       display     : flex;
       width       : 200px;
     }
   `
-  var filepanel = document.querySelector('#filepanel')
-  filepanel.className = css.filepanel
+  var filepanelContainer = document.querySelector('#filepanel')
+  filepanelContainer.className = css.filepanel
   var FilePanelAPI = {
     createName: createNonClashingName,
     switchToFile: switchToFile,
     event: this.event
   }
-  var el = new FilePanel(FilePanelAPI, files)
-  filepanel.appendChild(el)
-  var api = el.api
+  var filePanel = new FilePanel(FilePanelAPI, files)
+  // TODO this should happen inside file-panel.js
+  filepanelContainer.appendChild(filePanel)
 
-  api.register('ui', function changeLayout (data) {
+  filePanel.events.register('ui-hidden', function changeLayout (isHidden) {
     var value
-    if (data.type === 'minimize') {
+    if (isHidden) {
       value = -parseInt(window['filepanel'].style.width)
       value = (isNaN(value) ? -window['filepanel'].getBoundingClientRect().width : value)
       window['filepanel'].style.position = 'absolute'
       window['filepanel'].style.left = (value - 5) + 'px'
       window['filepanel'].style.width = -value + 'px'
       window['tabs-bar'].style.left = '45px'
-    } else if (data.type === 'maximize') {
+    } else {
       value = -parseInt(window['filepanel'].style.left) + 'px'
       window['filepanel'].style.position = 'static'
       window['filepanel'].style.width = value
       window['filepanel'].style.left = ''
       window['tabs-bar'].style.left = value
-    } else {
-      window['filepanel'].style.width = data.width + 'px'
-      window['tabs-bar'].style.left = data.width + 'px'
     }
   })
-  api.register('focus', function (path) {
-    [...window.files.querySelectorAll('.file .name')].forEach(function (span) {
-      if (span.innerText === path) switchToFile(path)
-    })
+  filePanel.events.register('ui-resize', function changeLayout (width) {
+    window['filepanel'].style.width = width + 'px'
+    window['tabs-bar'].style.left = width + 'px'
   })
   files.event.register('fileRenamed', function (oldName, newName) {
+    // TODO please never use 'window' when it is possible to use a variable
+    // that references the DOM node
     [...window.files.querySelectorAll('.file .name')].forEach(function (span) {
       if (span.innerText === oldName) span.innerText = newName
     })
   })
   files.event.register('fileRemoved', function (path) {
-    if (path === ui.get('currentFile')) {
-      ui.set('currentFile', '')
+    if (path === config.get('currentFile')) {
+      config.set('currentFile', '')
       switchToNextFile()
     }
     editor.discard(path)
@@ -338,7 +332,7 @@ var run = function () {
         if (!files.rename(originalName, newName)) {
           alert('Error while renaming file')
         } else {
-          ui.set('currentFile', '')
+          config.set('currentFile', '')
           switchToFile(newName)
           editor.discard(originalName)
         }
@@ -368,7 +362,7 @@ var run = function () {
   function switchToFile (file) {
     editorSyncFile()
 
-    ui.set('currentFile', file)
+    config.set('currentFile', file)
 
     if (files.isReadOnly(file)) {
       editor.openReadOnly(file, files.get(file))
@@ -399,10 +393,10 @@ var run = function () {
       $filesEl.append($('<li class="file"><span class="name">' + name + '</span><span class="remove"><i class="fa fa-close"></i></span></li>'))
     }
 
-    var currentFileOpen = !!ui.get('currentFile')
+    var currentFileOpen = !!config.get('currentFile')
 
     if (currentFileOpen) {
-      var active = $('#files .file').filter(function () { return $(this).find('.name').text() === ui.get('currentFile') })
+      var active = $('#files .file').filter(function () { return $(this).find('.name').text() === config.get('currentFile') })
       active.addClass('active')
     }
     $('#input').toggle(currentFileOpen)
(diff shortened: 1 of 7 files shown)


axic:
Can you split the two commits into two PRs? The first commit looks ok to be merged as is. The second commit looks way more complex :)

ninabreznik:
@chriseth What needs to be done here? 
To checkout fileExplorerCleanup branch and make the last 2 commits into 2 PR?
And to make tests pass?

What is fileExplorerCleanup1 branch? Do we need it or is everything here?

chriseth:
@ninabreznik The first commit of these two has been merged, so it should be clearer now. We should have a talk, please tell me when you have time.

src/app/file-panel.js
```diff
@@ -120,10 +122,15 @@ function filepanel (appAPI, files) {
     this.classList.toggle(css.isVisible)
     this.children[0].classList.toggle('fa-angle-double-right')
     this.children[0].classList.toggle('fa-angle-double-left')
-    api.trigger('ui', [{ type: isHidden ? 'minimize' : 'maximize' }])
+    events.trigger('ui-hidden', [isHidden])
   }
 
   function uploadFile (event) {
+    // TODO This should not go to the file explorer.
```
chriseth: Sorry, this is wrong, please disregard.

src/app/file-panel.js
```diff
@@ -120,10 +122,15 @@ function filepanel (appAPI, files) {
...
+    // The file explorer is merely a view on the current state of
+    // the files module. Please ask the user here if they want to overwrite
+    // a file and then just use `files.add`. The file explorer will
+    // pick that up via the 'fileAdded' event from the files module.
     ;[...this.files].forEach(fileExplorer.api.addFile)
```
chriseth: Please add the `uploadFile` function as an anonymous function directly to the `input` DOM node, so it is much clearer what `this` is. Furthermore, do not call `fileExplorer.api.addFile` but rather `files.addFile` directly, and then the treeView will just pick it up through the fileadded event.

src/app.js
```diff
@@ -184,56 +177,57 @@ var run = function () {
   var editor = new Editor(document.getElementById('input'))
 
   // ---------------- FilePanel --------------------
+  // TODO: All FilePanel related CSS should move into file-panel.js
+  // app.js provides file-panel.js with a css selector or a DOM element
+  // and file-panel.js adds its elements (including css), see "Editor" above
```
serapath: The `.filepanel` class below should actually be named `.filepanel-container` or something.
It get's assigned to the `filepanelContainer` in line `190` below.

So basically, the css here is not related to the "file-panel.js" or anything, but it's more related to the code that uses the file-panel and wants to position it or put it in the place it wants to use it. So i don't think it belongs into the "filepanel" itself.

In theory, a different codebase could re-use the file panel and append it to a container that puts it in a different place, so it would use different css than the one below.

src/app.js
```diff
@@ -184,56 +177,57 @@ var run = function () {
...
   var css = csjs`
     .filepanel    {
       display     : flex;
       width       : 200px;
     }
   `
-  var filepanel = document.querySelector('#filepanel')
-  filepanel.className = css.filepanel
+  var filepanelContainer = document.querySelector('#filepanel')
+  filepanelContainer.className = css.filepanel
```
serapath: it the above change is realized and the "@TODO" goes away, then probably in this line, we would do:
`filepanelContainer.className = css.filepanelContainer`
or
`filepanelContainer.className = css['filepanel-container']`

src/app.js
```diff
@@ -184,56 +177,57 @@ var run = function () {
...
     createName: createNonClashingName,
     switchToFile: switchToFile,
     event: this.event
   }
-  var el = new FilePanel(FilePanelAPI, files)
-  filepanel.appendChild(el)
-  var api = el.api
+  var filePanel = new FilePanel(FilePanelAPI, files)
+  // TODO this should happen inside file-panel.js
+  filepanelContainer.appendChild(filePanel)
```
serapath: I actually like that :-) So would the api change to:
var apiFP = new FilePanel(container, FilePanelAPI, files) // container is a DOM node or css selector

src/app.js
```diff
@@ -184,56 +177,57 @@ var run = function () {
...
 
-  api.register('ui', function changeLayout (data) {
+  filePanel.events.register('ui-hidden', function changeLayout (isHidden) {
```
serapath: So instead of having a `ui` event and then having the event or "payload" with the structure like for example:
* `{ type: 'minimize', value: true,  ... }`
* `{ type: 'maximize', value: true,  ... }`
* `{ type: 'show', value: true,  ... }`
* `{ type: 'hide', value: true,  ... }`
* ...

instead there would be a seperate event for each of those, so event `ui` and `{ type: 'hide', value: true,  ... }`
becomes event `ui-hidden` and `true`, right?

If that's the case, then ok, i'm fine with the change. It's probably a matter of personal taste. I usually like to send messages around that have a "type" field.

src/app.js
```diff
@@ -184,56 +177,57 @@ var run = function () {
...
-    [...window.files.querySelectorAll('.file .name')].forEach(function (span) {
-      if (span.innerText === path) switchToFile(path)
-    })
+  filePanel.events.register('ui-resize', function changeLayout (width) {
+    window['filepanel'].style.width = width + 'px'
+    window['tabs-bar'].style.left = width + 'px'
   })
   files.event.register('fileRenamed', function (oldName, newName) {
+    // TODO please never use 'window' when it is possible to use a variable
+    // that references the DOM node
```
serapath: I understand.
Basically this comes from the very bad habit of using ID's in html elements.
Every time somebody uses an ID for an html element, like for example `<div id="foobar">...</div>`, the **global variable** `window.foobar` will automatically be created and the DOM element with that ID "foobar" will be assigned.

So I did not bother to create a lof of variables, like for example:
`var filesTabBar = document.querySelector('#files')` because there is already the automatically created global variable `window.files` that references the "tab bar".

I would highly recommend to slowly refactor the code base away from any uses of "ID"s in html tags :-)

src/app/file-explorer.js
```diff
@@ -20,12 +20,9 @@ var css = csjs`
     background-color  : white;
   }
   .remove             {
-    align-self        : center;
-    padding-left      : 10px;
+    float             : right;
```
serapath: I highly recommend to **ALWAYS** use `display: flex`.
These days it is supported by all browsers http://caniuse.com/#search=flex (also click: Usage relative)

It's an absolutely superior method to do any kind of layout in the browser. It is reliable and always works.
It takes some time to pick up and a good source is this:
* https://css-tricks.com/snippets/css/a-guide-to-flexbox/
* (for a playful introduction) http://flexboxfroggy.com/

Maybe the changed styling works - if it does, sure why not merge it. I dont mind, but personally i would try to stick with flexbox. If it is supposed to fix some weird styling behavior, maybe it was introduced by [this commit](https://github.com/ethereum/remix/commit/a503dc2c412f92fb262946a548e150f199953c22) to the remix repository?

src/app/file-explorer.js
```diff
@@ -70,9 +67,9 @@ function fileExplorer (appAPI, files) {
   })
 
   var deleteButton = yo`
-    <span class=${css.remove} onclick=${deletePath}>
+    <div class=${css.remove} onclick=${deletePath}>
```
serapath: a `span` has the same layout properties as a `div` with `dislpay: inline-block;` or `display: inline;` or `display: inline-flex;`.
Maybe this change is/was also styling related :-)

My personal recommendation would be to try to stick to flexbox (until in the future the grid layout becomes widely available)
* http://gridbyexample.com/examples/
* https://css-tricks.com/snippets/css/complete-guide-grid/

src/app/file-panel.js
```diff
@@ -106,11 +106,13 @@ function filepanel (appAPI, files) {
...
 
-  var api = new EventManager()
+  var events = new EventManager()
   var element = template()
-  element.api = api
-  fileExplorer.api.register('focus', function (path) {
-    api.trigger('focus', [path])
+  // TODO please do not add custom javascript objects, which have no
+  // relation to the DOM to DOM nodes
+  element.events = events
```
serapath: I understand.
The element represent's the instance and so the events object is kind of related to the API of that instance.

The contsructor returns one element, and that was until now the DOM node, but because the constructur will from now on take a selector or container element to which to attach to, it is instead possible to return the api object and not the element object from the constructor (or let's say from the factory function) that created the "file-panel" :-)

src/app/file-panel.js
```diff
@@ -120,10 +122,14 @@ function filepanel (appAPI, files) {
...
     this.children[0].classList.toggle('fa-angle-double-left')
-    api.trigger('ui', [{ type: isHidden ? 'minimize' : 'maximize' }])
+    events.trigger('ui-hidden', [isHidden])
   }
 
   function uploadFile (event) {
+    // TODO The file explorer is merely a view on the current state of
+    // the files module. Please ask the user here if they want to overwrite
+    // a file and then just use `files.add`. The file explorer will
+    // pick that up via the 'fileAdded' event from the files module.
```
serapath: Ok. I think this code comes from a time when there was no direct access to the `files` object yet

src/app.js
yann300: yes that's that

src/app/file-explorer.js
yann300: using `align-self: center` is not really appropriate cause we don't want to center it but to push it on the right..
yes it was introduce by this commit, but the commit should stay i think.
about using `flex`, as I understood it, this is a good way to distribute space between component in various ways, but does it apply to our case? cause here we just want to push the remove icon close to the right border

src/app/file-explorer.js
yann300: yes rolledback to span

src/app/file-explorer.js
serapath: in flex you have 2 axis. One is for vertical, the other for horizontal. The layout among one axis can be changed with the property `justify-content` and the other with the `align-items` ..but which axis is which depends on the flex layout, which can be `flex-direction: row` (default) or `flex-direction: column;`
The axis (vertical/horizontal) will switch based on that

...now if you want to layout elements among the axis, you have many options... to put them in the begining or end there are the properties, e.g.: `align-items: flex-start;` or `align-items: flex-end;`  ...and the same for `justify-content: flex-start;` or `justify-content: flex-end;` ...but it depends on row/column whether that affects the x- or y-axis :-)

yann300: In that case, we don't want to `align-items: start` or `end`, we just want to put the `remove` icon at the end. can  `flex`  able to handle that?

serapath: yes. i'll do that

PR #525, opened 2017-04-19
Fix readonly files in file explorer and sync to tabs.

chriseth:


src/app.js
```diff
@@ -224,7 +209,15 @@ var run = function () {
     })
   })
   files.event.register('fileRemoved', function (path) {
-    if (path === ui.get('currentFile')) ui.set('currentFile', '')
+    editor.discard(path)
```
axic: I think `discard` must be after `switchToNextFile` otherwise the editor cannot discard it.

src/app.js
```diff
@@ -394,10 +383,6 @@ var run = function () {
     }
     $('#input').toggle(currentFileOpen)
     $('#output').toggle(currentFileOpen)
-
-    $filesEl.animate({ left: Math.max((0 - activeFilePos() + (FILE_SCROLL_DELTA / 2)), 0) + 'px' }, 'slow', function () {
```
axic: 👍

src/app.js
```diff
@@ -342,10 +335,6 @@ var run = function () {
     if (confirm('Are you sure you want to remove: ' + name + ' from local storage?')) {
       if (!files.remove(name)) {
         alert('Error while removing file')
-      } else {
-        ui.set('currentFile', '')
-        switchToNextFile()
-        editor.discard(name)
```
axic: This can be removed because `files.remove()` will trigger `fileRemoved`.

src/app/file-explorer.js
```diff
@@ -76,11 +76,14 @@ function fileExplorer (appAPI, files) {
     </span>
   `
 
-  appUI.register('currentFile', fileFocus)
+  appUI.register('fileChanged', (changedFiles) => {
+    if (changedFiles[0] == 'currentFile') {
+      fileFocus(files.get('currentfile'))
```
axic: Does this mean that `currentfile` is a file in the regular files API? Also why is one camelcase and the other not?

src/app/file-explorer.js
```diff
@@ -253,8 +254,6 @@ function fileExplorer (appAPI, files) {
     el.className = css.fileexplorer
     element.parentElement.replaceChild(el, element)
     element = el
-    fileFocus(filepath)
-    appAPI.switchToFile(filepath)
```
axic: Will this  `switchToFile` done by `fileFocus` or app catches the `fileAdded` event and does the switch itself?

src/app/files.js
```diff
@@ -138,12 +138,7 @@ function Files (storage) {
     var tree = {}
 
     var self = this
-    storage.keys().forEach(function (path) {
-      // NOTE: as a temporary measure do not show the config file
-      if (path === '.remix.config') {
-        return
-      }
-
+    Object.keys(this.list()).forEach(function (path) {
```
axic: I guess adding a comment here could be useful to say that since it uses `list()`, the `.remix.config` is removed form the list.

src/app.js
chriseth: Yes, that's the idea. It IS removed, right?

src/app/file-explorer.js
chriseth: I think the old behaviour was wrong. Why should I switch to a file just because it was added? The example use-case is that we import a file from github - we should not automatically switch to it. We should only switch when we use the add-button to create a new file, but that should be hanlded by the add button, not by the tree view.

axic: It was retained from the original code. I'm not keen on having it. It makes sense when a new file is created with the "New file" button, however not so much when it is imported.

src/app.js
axic: Yes, just confirmed it to myself :)

src/app/file-explorer.js
```diff
@@ -76,11 +76,14 @@ function fileExplorer (appAPI, files) {
     </span>
   `
 
-  appUI.register('currentFile', fileFocus)
+  appUI.register('fileChanged', (changedFiles) => {
+    if (changedFiles[0] === 'currentFile') {
+      fileFocus(files.get('currentFile'))
```
axic: This probably should change to `appUI.get()`?

src/app.js
```diff
@@ -224,7 +209,15 @@ var run = function () {
     })
   })
```
axic: I guess `refreshTabs()` is needed here too?

PR #532, opened 2017-04-24
Loading spinner

ninabreznik:
* Rearange tabs - contract first tab
* Change default tab to contract
* Add loaderSpinner to contract and settings tab

?? Since changing a compiler version is super connected to the contract, maybe we should think about how to elegantly add **change version** to contract tab, maybe above could be settings (all that is there already + current version/change version) and below the contract(s). 
Don't know exactly how, but it's hard for new users to understand that in the settings you change the compiler version and that actually changes the version in all your contracts... Would make sense to put this together somehow...

### DIFF
--- a/index.html
+++ b/index.html
@@ -62,7 +62,7 @@
 						<li class="publishView" title="Publish" >Files</li>
 						<li class="debugView" title="Debugger">Debugger</li>
 						<li class="staticanalysisView" title="Static Analysis">Analysis</li>
-						<li id="helpButton"><a href="https://solidity.readthedocs.org" target="_blank" title="Open Documentation">Docs</a></li>
+						<li id="helpButton"><a href="https://remix.readthedocs.org" target="_blank" title="Open Documentation">Docs</a></li>
 					</ul>
 				</div>
 				<div id="optionViews" class="settingsView">
--- a/index.html
+++ b/index.html
@@ -58,10 +58,10 @@
 					<img id="solIcon" title="Solidity realtime compiler and runtime" src="assets/img/remix_logo_512x512.svg" alt="Solidity realtime compiler and runtime">
 					<ul id="options">
 						<li class="envView" title="Environment">Contract</li>
+						<li class="settingsView" title="Settings">Settings</li>
 						<li class="publishView" title="Publish" >Files</li>
 						<li class="debugView" title="Debugger">Debugger</li>
 						<li class="staticanalysisView" title="Static Analysis">Analysis</li>
-						<li class="settingsView" title="Settings">Settings</li>
 						<li id="helpButton"><a href="https://solidity.readthedocs.org" target="_blank" title="Open Documentation">Docs</a></li>
 					</ul>
 				</div>
--- a/index.html
+++ b/index.html
@@ -57,7 +57,7 @@
 				<div id="menu">
 					<img id="solIcon" title="Solidity realtime compiler and runtime" src="assets/img/remix_logo_512x512.svg" alt="Solidity realtime compiler and runtime">
 					<ul id="options">
-						<li class="envView active" title="Environment">Contract</li>
+						<li class="envView" title="Environment">Contract</li>
 						<li class="publishView" title="Publish" >Files</li>
 						<li class="debugView" title="Debugger">Debugger</li>
 						<li class="staticanalysisView" title="Static Analysis">Analysis</li>
--- a/index.html
+++ b/index.html
@@ -57,11 +57,11 @@
 				<div id="menu">
 					<img id="solIcon" title="Solidity realtime compiler and runtime" src="assets/img/remix_logo_512x512.svg" alt="Solidity realtime compiler and runtime">
 					<ul id="options">
-						<li class="envView" title="Environment">Contract</li>
+						<li class="envView active" title="Environment">Contract</li>
 						<li class="publishView" title="Publish" >Files</li>
 						<li class="debugView" title="Debugger">Debugger</li>
 						<li class="staticanalysisView" title="Static Analysis">Analysis</li>
-						<li class="settingsView active" title="Settings">Settings</li>
+						<li class="settingsView" title="Settings">Settings</li>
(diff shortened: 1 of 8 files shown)


src/app/tabbed-menu.js
```diff
@@ -0,0 +1,48 @@
...
+  compiler.event.register('loadingCompiler', function compilationStarted () {
+    var contractTab = document.querySelector('.envView')
+    if (!contractTab.children.length) {
+      contractTab.appendChild(loadingSpinner(function cb (finish) {
+        // compiler.event.register('compilationFinished', function () {
+        compiler.event.register('compilerLoaded', finish)
+      }))
+    }
+  })
+  compiler.event.register('loadingCompiler', function loadingCompiler () {
```
yann300: no need to register twice no?

src/app/tabbed-menu.js
```diff
@@ -0,0 +1,48 @@
...
+  selectTab($('#options .envView'))
+
+  // add event listeners for loading spinner
+  // compiler.event.register('compilationStarted', function compilationStarted () {
+  compiler.event.register('loadingCompiler', function compilationStarted () {
+    var contractTab = document.querySelector('.envView')
+    if (!contractTab.children.length) {
+      contractTab.appendChild(loadingSpinner(function cb (finish) {
+        // compiler.event.register('compilationFinished', function () {
+        compiler.event.register('compilerLoaded', finish)
```
yann300: better to register to `compilerLoaded` outside of the `loadingCompiler` registration

src/app/tabbed-menu.js
```diff
@@ -0,0 +1,48 @@
...
+      }))
+    }
+  })
+  compiler.event.register('loadingCompiler', function loadingCompiler () {
+    var settingsTab = document.querySelector('.settingsView')
+    if (!settingsTab.children.length) {
+      settingsTab.appendChild(loadingSpinner(function cb (finish) {
+        compiler.event.register('compilerLoaded', finish)
```
yann300: same as above

src/app/tabbed-menu.js
```diff
@@ -0,0 +1,48 @@
...
+    var $el = $(this)
+    selectTab($el)
+  })
+
+  // initialize tabbed menu
+  selectTab($('#options .envView'))
+
+  // add event listeners for loading spinner
+  // compiler.event.register('compilationStarted', function compilationStarted () {
+  compiler.event.register('loadingCompiler', function compilationStarted () {
```
yann300: or maybe you mean `compilationStarted` here?

ninabreznik:
@yann300 Fixed! Pls check

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - stable
 script:
-  - npm run lint && npm run csslint && npm run test && npm run downloadsolc && npm run make-mock-compiler && npm run build
+  - npm run lint && npm run test && npm run downloadsolc && npm run make-mock-compiler && npm run build
```
yann300: please rebase

src/app.js
```diff
@@ -277,24 +286,7 @@ var run = function () {
     }).appendTo('body')
   })
 
-  // ---------------- tabbed menu ------------------
-  $('#options li').click(function (ev) {
-    var $el = $(this)
-    selectTab($el)
-  })
-
-  var selectTab = function (el) {
```
yann300: that's cool to remove `selectTab` from here but it is used. Look at the function `startdebugging` in app.js. we could pass the debugger event to ´tabbedMenu´. and listen on `startDebugging` event
.

src/app/tabbed-menu.js
```diff
@@ -0,0 +1,58 @@
...
+
+  $('#options li').click(function (ev) {
+    var $el = $(this)
+    selectTab($el)
+  })
+
+  // initialize tabbed menu
+  selectTab($('#options .envView'))
+
+  function cb (finish) {
```
yann300: this is not used i think

index.html
```diff
@@ -57,12 +57,12 @@
...
 					<ul id="options">
-						<li class="settingsView active" title="Settings">Settings</li>
-						<li class="publishView" title="Publish" >Files</li>
 						<li class="envView" title="Environment">Contract</li>
+						<li class="settingsView" title="Settings">Settings</li>
+						<li class="publishView" title="Publish" >Files</li>
 						<li class="debugView" title="Debugger">Debugger</li>
 						<li class="staticanalysisView" title="Static Analysis">Analysis</li>
-						<li id="helpButton"><a href="https://remix.readthedocs.org" target="_blank" title="Open Documentation">Docs</a></li>
+						<li id="helpButton"><a href="https://solidity.readthedocs.org" target="_blank" title="Open Documentation">Docs</a></li>
```
yann300: that should be remix.readthedocs.org

src/app/tabbed-menu.js
ninabreznik: It is in use... line 23 sets which tab is a default tab (it's not necessary the first one). It was already in the code, so I just changed it to the be a contract tab 

And finish is used to finish spinner after you select new solidity version.

src/app.js
ninabreznik: Hm, I actually can't remember removing this one... Would you maybe do this part, since you understand the debugging a bit better, so we don't mess up something here.

yann300: sure

ninabreznik:
Moved logic to tabbed-menu, loadingSpinner is now only a visual component that can be used on other spaces.

IMPORTANT! We spent at least 2h debugging this, trying out all sorts of stuff and then @serapath figured out what is causing it: a 'bug'? in eventManager that assigns an empty object for each registration if there object is not specified. And the point is that when you want to unregister and create a function, it creates a new empty object and then compares old and new empty object and doesn't unregister, because they are not the same...

yann300:
hmm thanks for that!

yann300:
replaced by https://github.com/ethereum/browser-solidity/issues/553

PR #533, opened 2017-04-26
Style guide

ninabreznik:
When I started styling and moving around elements, I figured out this will take ages if we don't start creating components.

Therefore, we now:
- created **style-sheet** to unify design/colors/size of buttons, input fields etc.
- moved tab views in separate files: contract-tab.js, settings-tab.js, debugger-tab.js, analysis-tab.js, files-tab.js
- applied basic unified styling on each tab component, using this style-sheet

From now on, we will have a bit easier job restyling stuff in right-hand panel 
BUT
we still need to transition all the styling to components + the logic from app.js etc.

It will be a process, but we have a base now.

Please, check how it feels, if it's ok for first merge and then we continue.

Probably we will also need to solve the tests. Will see now, what Mr. Travis says :)


.travis.yml
```diff
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - stable
 script:
-  - npm run lint && npm run csslint && npm run test && npm run downloadsolc && npm run make-mock-compiler && npm run build
+  - npm run lint && npm run test && npm run downloadsolc && npm run make-mock-compiler && npm run build
```
axic: Why remove csslint?

axic:
This PR seems to be rather big with seemingly unrelated changes. Any chance splitting it up?

.travis.yml
chriseth: It is not really helpful, it reports errors, but we ignore them.

test-browser/tests/compiling.js
```diff
@@ -30,7 +30,11 @@ function runTests (browser) {
   contractHelper.testContracts(browser, sources.sources.Untitled, ['Untitled:TestContract'], function () {
     browser.click('.create .constructor .call')
       .waitForElementPresent('.instance .call[title="f"]')
-      .click('.instance .call[title="f"]')
+    if (browser.options.desiredCapabilities.browserName === 'internet explorer') {
```
chriseth: The tests seem to be disabled above completely. Please choose either :-)

src/universal-dapp.js
```diff
@@ -336,7 +421,7 @@ UniversalDApp.prototype.getInstanceInterface = function (contract, address, $tar
       var eventFilter = self.web3.eth.contract(abi).at(address).allEvents()
       eventFilter.watch(parseLogs)
     }
-    $instance.append($title)
+    $instance.get(0).appendChild(title)
```
chriseth: Is it possible to give this inner child a name or an id so that we do not have to change this line if we change the html layout in the future?

.travis.yml
axic: We've only ignored the less severe warnings. And that the ignore list should be shrunk too as we fix those less severe warnings, just nobody bothered.

It did help us catch a ton of severely invalid CSS when I've added it.

src/universal-dapp.js
ninabreznik: @chriseth This part is just jQuery refactoring
.append is a jQuery thingy and .appendChild works on dom elements. And therefore I change $instance into $instance.get(0) to get a dom element so I can use .appendChild on it. 

It's a slow process of removing jQuery, but I would rather do it bit by bit, otherwise we need few more days to refactor the whole thing now. I would suggest we leave it like this. I could do var instance = yo`<div>something</div>`, but then we use this $instance lots of times and this would cause a chain reaction, so... maybe we could skip it for now.

.travis.yml
yann300: When csslint fails it breaks `downloadsolc` and `make-mock-compiler `... this is why tests were failing more often recently. need to investigate why.
We, again, still need to work on the tests infra.. but that would be better to get this merged before

yann300:
yeah @axic sorry for that :) . that pr also been used to track the failing tests.

PR #538, opened 2017-05-09
[WIP] Shortcut

yann300:
- fix a bug in autocompilation
 - disable autocompile by default
 - warn the user if compilation needs to be run
 - integrate `hotkeys-fr` ( F5 will run a compilation )

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "execr": "^1.0.1",
     "exorcist": "^0.4.0",
     "fast-async": "^6.1.2",
+    "hotkeys-js": "^2.0.1",
     "http-server": "0.9.0",
     "jquery": "^2.2.0",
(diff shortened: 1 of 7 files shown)


ninabreznik:
@yann300 So auto-compile is unmarked and on compile button you added a shortcut? I had to press Fn+F5, just F5 doesn't work on Mac... You also mention that it 'warns users'? I didn't get any alert or other type of warning. What did you mean by that?

yann300:
we need to find a better shortcut... in Xcode what is the shortcut to start a compilation?
There should be an alert in the contract tab ( a warning incon ). This icon is only shown when you change the content of the editor

yann300:
disable autocompile will not be implemented (instead we warn).
this code could be re use when we want to add shorcut.

PR #556, opened 2017-05-23
fix gas Limit

yann300:
That PR lower the gas limit calculation (from the block gas limit). 
In case the next blocks have their gas limit lowered down.


yann300:
fixes https://github.com/ethereum/remix/issues/506

PR #557, opened 2017-05-23
fix open issues from pull#524

serapath:
fix all open issues from https://github.com/ethereum/browser-solidity/pull/524/files
the one remaining is styling which will be handled in another pull request

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -159,6 +159,7 @@ var run = function () {
       width       : 200px;
     }
   `
+
   var filepanelContainer = document.querySelector('#filepanel')
   filepanelContainer.className = css['filepanel-container']
   var FilePanelAPI = {
@@ -655,8 +656,25 @@ var run = function () {
       var match = handler.match.exec(url)
       if (match) {
         found = true
-
-        $('#output').append($('<div/>').append($('<pre/>').text('Loading ' + url + ' ...')))
+        var css = csjs`
+          .pre {
+            padding: 8px 15px;
+            background: #f8f8f8;
+            border-radius: 5px;
+            border: 1px solid #e5e5e5;
+            overflow-x: auto;
+            font-family: Monaco, Bitstream Vera Sans Mono, Lucida Console, Terminal, monospace;
+            color: #333;
+            font-size: 12px;
+            margin: 0 0 20px;
+          }
+          @media print, screen and (max-width: 720px) {
+            pre {
+              word-wrap: normal;
+            }
+          }
+        `
+        $('#output').append($('<div/>').append($('<pre class="' + css.pre + '" />').text('Loading ' + url + ' ...')))
         handler.handler(match, function (err, content) {
           if (err) {
             cb('Unable to import "' + url + '": ' + err)
--- a/assets/css/styles.css
+++ b/assets/css/styles.css
@@ -11,7 +11,7 @@ h1, h2, h3, h4, h5, h6 {
   margin: 20px 0 10px;
 }
 
-p, ul, ol, table, pre, dl {
+p, ul, ol, table, dl {
   margin: 0 0 20px;
 }
 
@@ -60,18 +60,10 @@ blockquote {
   font-style: italic;
 }
 
-code, pre {
+code {
   font-family: Monaco, Bitstream Vera Sans Mono, Lucida Console, Terminal, monospace;
   color: #333;
   font-size: 12px;
-}
-
-pre {
-  padding: 8px 15px;
-  background: #f8f8f8;
-  border-radius: 5px;
-  border: 1px solid #e5e5e5;
-  overflow-x: auto;
 }
 
 table {
@@ -193,13 +185,6 @@ small {
   font-size: 11px;
 }
 
-hr {
-  border: 0;
-  background: #e5e5e5;
-  height: 1px;
-  margin: 0 0 20px;
-}
-
 footer {
   width: 270px;
   float: left;
@@ -255,7 +240,7 @@ footer {
     position: static;
   }
 
-  pre, code {
+  code {
     word-wrap: normal;
   }
 }
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "clipboard-copy": "^1.2.0",
     "csjs-inject": "^1.0.1",
     "csslint": "^1.0.2",
+    "dom-console": "^5.1.1",
     "ethereum-remix": "https://github.com/ethereum/remix",
     "ethereumjs-abi": "https://github.com/ethereumjs/ethereumjs-abi",
     "ethereumjs-block": "https://github.com/ethereumjs/ethereumjs-block",
--- a/src/app/renderer.js
+++ b/src/app/renderer.js
@@ -36,6 +36,22 @@ var css = csjs`
   .toggle:hover {
     opacity: .8;
   }
+  .pre {
+    padding: 8px 15px;
+    background: #f8f8f8;
+    border-radius: 5px;
+    border: 1px solid #e5e5e5;
+    overflow-x: auto;
+    font-family: Monaco, Bitstream Vera Sans Mono, Lucida Console, Terminal, monospace;
+    color: #333;
+    font-size: 12px;
+    margin: 0 0 20px;
+  }
+  @media print, screen and (max-width: 720px) {
+    pre {
+      word-wrap: normal;
+    }
+  }
 `
 // ----------------------------------------------
 
@@ -77,9 +93,9 @@ Renderer.prototype.error = function (message, container, options) {
   }
   var $pre
   if (opt.isHTML) {
-    $pre = $(opt.useSpan ? '<span />' : '<pre />').html(message)
+    $pre = $(opt.useSpan ? '<span />' : '<pre class="' + css.pre + '" />').html(message)
   } else {
-    $pre = $(opt.useSpan ? '<span />' : '<pre />').text(message)
+    $pre = $(opt.useSpan ? '<span />' : '<pre class="' + css.pre + '" />').text(message)
   }
   var $error = $('<div class="sol ' + opt.type + '"><div class="close"><i class="fa fa-close"></i></div></div>').prepend($pre)
   if (container === undefined) {
@@ -147,7 +163,7 @@ Renderer.prototype.contracts = function (data, source) {
   var preRow = function (description, data) {
     return tableRowItems(
       $('<span/>').text(description),
-      $('<pre/>').text(data)
+      $('<pre class="' + css.pre + '" />').text(data)
     )
   }
 
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -280,13 +280,11 @@ UniversalDApp.prototype.getContractByName = function (contractName) {
 }
 
 UniversalDApp.prototype.getCreateInterface = function ($container, contract) {
+  function remove () { self.$el.remove() }
   var self = this
   var createInterface = yo`<div class="create"></div>`
   if (self.options.removable) {
     var close = yo`<div class="udapp-close" onclick=${remove}></div>`
-    function remove () {
-      self.$el.remove()
-    }
     createInterface.appendChild(close)
   }
 
@@ -347,11 +345,11 @@ UniversalDApp.prototype.getInstanceInterface = function (contract, address, $tar
   var createInterface = yo`<div class="createContract"></div>`
 
   var appendFunctions = function (address, $el) {
+    function remove () { $instance.remove() }
     if ($el) $el = $el.get(0)
     var $instance = $(`<div class="instance ${cssInstance.instance}"/>`)
     if (self.options.removable_instances) {
       var close = yo`<div class="udapp-close" onclick=${remove}></div>`
-      function remove () { $instance.remove() }
       $instance.get(0).appendChild(close)
     }
     var context = self.executionContext.isVM() ? 'memory' : 'blockchain'


src/app/file-explorer.js
```diff
@@ -88,18 +88,6 @@ function fileExplorer (appAPI, files) {
 
   var events = new EventManager()
```
chriseth: Can you create these two variables closer to where they are used?

src/app/file-explorer.js
```diff
@@ -251,9 +239,9 @@ function fileExplorer (appAPI, files) {
     element = el
   }
 
-  element.events = events
-  element.api = api
-  return element
+  api.events = events
+  container.appendChild(element)
+  return api
```
chriseth: I think it would be more consistent if this function would be used as constructor and then returns an object that has a property `api`.

src/app/file-panel.js
```diff
@@ -76,8 +76,12 @@ var limit = 60
 var canUpload = window.File || window.FileReader || window.FileList || window.Blob
 var ghostbar = yo`<div class=${css.ghostbar}></div>`
 
-function filepanel (appAPI, files) {
-  var fileExplorer = new FileExplorer(appAPI, files)
+function filepanel (container, appAPI, files) {
+  if (!container) return
```
chriseth: Shouldn't an exception be thrown here?

src/app/file-panel.js
```diff
@@ -108,14 +112,16 @@ function filepanel (appAPI, files) {
...
   fileExplorer.events.register('focus', function (path) {
     appAPI.switchToFile(path)
   })
 
-  return element
+  var api = { events: events }
+
+  container.appendChild(element)
+
+  return api
```
chriseth: Here: https://github.com/ethereum/browser-solidity/pull/557/files#diff-bd9c9dcd314f2d7df52935b3a6a4d504R198 you use `filepanel` as the constructor of an object, but you return something explicitly in this line. Am I looking at two different things? If yes, they should not have the same name. If no, please just use something like `this.events = events` here and do not return explicitly.

src/app/file-explorer.js
serapath: Hmm, ...
The object, with the property `api` existed before, but the object was a DOM node.
I wonder how exactly would we want to use it, because that's what we will do all over the place later...

```js
var foobarAPI = { baz: function bar (a,b,c) { console.log(a,b,c) } }
var foobarEvents = files.events // for example
var foobar = new Foobar('.foobarContainer', foobarAPI, foobarEvents)

// the above - as far as i get it - is currently our standard of designing "components", right?
// ... now what is `foobar` ?

// 1. `foobar` was a DOM element with the properties:
var foobar = yo`<div></div>` // but returned from the constructor above of course
foobar.events
foobar.api
foobar.api.callAFoobarMethod()
foobar.events.register('m3h', function () { console.log('meow') }

// 2. then we switched to a model, where a container node is passed into the constructor
var foobar = {
  events: new EventManager(),
  callAFoobarMethod: function () {},
  callAnotherMethod: function () {}
}

// 3. i understand your above proposal as:
var foobar = {
  events: new EventManager(),
  api: {
    callAFoobarMethod: function () {},
    callAnotherMethod: function () {}
  }
}

// 4. Can I add another option that maybe would spontaneously be my favourite option? :P
var foobar = new EventEmitter() // which - by convention - restricts api methods to be called:
// foobar.register
// foobar.unregister
// foobar.trigger
foobar.callAFoobarMethod = function () {}
foobar.callAnotherMethod = function () {}
// personally i would even prefer a system where the `.trigger` method is seperated from the "event emitter", thus one object where callbacks can be registered/unregistered and one on which events can be triggered... the one where events can be triggered should probably stay private to an instance, while the register/unregister method will be public on the instance returned from the constructor call.
```

**What's the reasoning why would any of those options from above be better/worse than others?**

I can see a lot of value in the last option (4.) and i am relatively indifferent to options (1. - 3.)

Whatever the decision would be, it should probably be documented somewhere as a convention for modules/components in general

src/app/file-panel.js
serapath: oh, hm... probably :P ok i will fix that :P

src/app/file-panel.js
serapath: oh yeah, you are right. that actually doesnt work anymore. interesting. i'll think about it and adapt the pull request. good catch, thx.

serapath: ah ok - i think i get it.
* I normally always use the style showed in `FooEasy` below.
* If i need to get fancy - which i usually dont - but maybe I need to create tons ob instances in a short time and it's the performance bottleneck of my application, then i might optimize by utilizing javascripts prototype chain as shown in version `Foo3`.

**but...**
...if you for some reasons - maybe familiarity or the likes prefer to introduce some extra javascript keywords with special meaning on top of the HTML5 cross browser implemented and ES6 standard property `__proto__` ... let's say you would like to add: `new` and `this` and `prototype` property or maybe even `class` and `super` and the `constructor` property ... i guess the sky is the limit - to essentially do the same as what you can do with the `FooFast` style

... you might follow the examples in `FooTraditional` listed below instead.

They all work the same if you call them with `new`, but the explicit return also works if you forget the `new` operator. The ones with the explicit return works without all that magic introduced by special keywords of the language (like `this` or `new` and the knowledge of how `prototype`s in javascript work). ...just direct access via `__proto__` needs to be known if you need performance optimization.
I perceive it less heavy to reason about, because i don't have to guess what `this` is when i encounter it, e.g
* http://2ality.com/2014/05/this.html

or maybe just choose the correct behavior from the examples below :-)

```js
var instance

function FooEasy () {
  return { foo: 'bar' }
}
// normal use
console.log(instance = FooEasy()) // => 'bar'
// unusual use
console.log(instance instanceof FooEasy) // => false
console.log(instance = new FooEasy()) // => 'bar'
console.log(instance instanceof FooEasy) // => false


function FooFast () {
  return { foo: 'bar', __proto__: {/*put methods here*/} } // in case you really need the prototype chain
}
console.log(instance = FooFast()) // => 'bar'
console.log(instance instanceof FooFast) // => false
console.log(instance = new FooFast()) // => 'bar'
console.log(instance instanceof FooFast) // => false


function FooTraditional () {
  this.foo = 'bar'
}
console.log(instance = FooTraditional()) // => undefined
console.log(instance instanceof FooTraditional) // => false
console.log(instance = new FooTraditional()) // => 'bar'
console.log(instance instanceof FooTraditional) // => true

function FooTraditionalEmulated () {
  if (this instanceof Foo4)  return { foo: 'bar', __proto__: Foo4.prototype } // was called with 'new'
}
console.log(instance = FooTraditionalEmulated()) // => undefined
console.log(instance instanceof FooTraditionalEmulated) // => false
console.log(instance = new FooTraditionalEmulated()) // => 'bar'
console.log(instance instanceof FooTraditionalEmulated) // => true
```

serapath: And because I kind of feel this is related to the issue in the previous versions of this pull request regarding the "API" of modules/component, here again the excerpt from the outdated comment above:

---
Hmm, ...
The object, with the property `api` existed before, but the object was a DOM node.
I wonder how exactly would we want to use it, because that's what we will do all over the place later...

```js
var foobarAPI = { baz: function bar (a,b,c) { console.log(a,b,c) } }
var foobarEvents = files.events // for example
var foobar = new Foobar('.foobarContainer', foobarAPI, foobarEvents)

// the above - as far as i get it - is currently our standard of designing "components", right?
// ... now what is `foobar` ?

// 1. `foobar` was a DOM element with the properties:
var foobar = yo`<div></div>` // but returned from the constructor above of course
foobar.events
foobar.api
foobar.api.callAFoobarMethod()
foobar.events.register('m3h', function () { console.log('meow') }

// 2. then we switched to a model, where a container node is passed into the constructor
var foobar = {
  events: new EventManager(),
  callAFoobarMethod: function () {},
  callAnotherMethod: function () {}
}

// 3. i understand your above proposal as:
var foobar = {
  events: new EventManager(),
  api: {
    callAFoobarMethod: function () {},
    callAnotherMethod: function () {}
  }
}

// 4. Can I add another option that maybe would spontaneously be my favourite option? :P
var foobar = new EventEmitter() // which - by convention - restricts api methods to be called:
// foobar.register
// foobar.unregister
// foobar.trigger
foobar.callAFoobarMethod = function () {}
foobar.callAnotherMethod = function () {}
// personally i would even prefer a system where the `.trigger` method is seperated from the "event emitter", thus one object where callbacks can be registered/unregistered and one on which events can be triggered... the one where events can be triggered should probably stay private to an instance, while the register/unregister method will be public on the instance returned from the constructor call.
```

**What's the reasoning why would any of those options from above be better/worse than others?**

I can see a lot of value in the last option (4.) and i am relatively indifferent to options (1. - 3.)

Whatever the decision would be, it should probably be documented somewhere as a convention for modules/components in general

src/app/file-explorer.js
yann300: for `3.` not really:

```
// 3. i understand your above proposal as:
var foobar = {
  events: new EventManager(),
  api: {
    callAFoobarMethod: function () {},
    callAnotherMethod: function () {}
  }
}
```

The api object contains all the stuff that the module require from the outside.
And the way it is instanciated is: 
for instance for the Debugger:

var debug = new debug("#domid", api, executionContext.event, edtior.event)
the debug component needs this `api` to retrieve stuff from the outside, and need to register to some events. In that case he'll register to the `editor event` and `execution context event`
e.g https://github.com/ethereum/browser-solidity/blob/master/src/app.js#L650

Note that the code of the dom component (yoyo) is completely separated from the execution logic.

src/app/file-panel.js
chriseth: We should be consistent. If you use `new FooBar()`, then I would expect `FooBar` to be a constructor function that does not return explicitly. Please keep it simple, we do not need optimization here. My perception is that `new` without explicit return (and potentially prototypes but not necessarily) is the standard way. It is at least the standard way in this codebase.

src/app.js
```diff
@@ -182,9 +182,6 @@ var run = function () {
   var editor = new Editor(document.getElementById('input'))
 
   // ---------------- FilePanel --------------------
-  // TODO: All FilePanel related CSS should move into file-panel.js
-  // app.js provides file-panel.js with a css selector or a DOM element
-  // and file-panel.js adds its elements (including css), see "Editor" above
   var css = csjs`
     .filepanel-container    {
```
yann300: You use this `css` on a static container. why not directly put the class there: 
https://github.com/ethereum/browser-solidity/blob/master/index.html#L49 ?

src/app/file-explorer.js
```diff
@@ -250,10 +235,11 @@ function fileExplorer (appAPI, files) {
...
   }
-
-  element.events = events
-  element.api = api
-  return element
+  
+  var events = new EventManager()
+  var api = { events: events }
+  container.appendChild(element)
+  return api
```
yann300: why do you return `api` ?
could you put 
`this.events = new EventManager()`
`container.appendChild(element)`
at the beginning of the file?

serapath: ok, maybe. that's why i wrote the long comment above.

In JavaScript, the `new` operator and `this` is considered bad style and should be avoided.
But it exists for traditional reasons.
Here is the longer explanation: https://github.com/ethereum/browser-solidity/pull/557#discussion_r118080066

...anyway, i can do it and add it as you wish.

src/app.js
serapath: Regarding css.
I recommend to read:
* https://medium.com/seek-blog/a-unified-styling-language-d0c208de2660

Despite that - the separation of css/html/js in separate files is considered - not separation of concerns, but separation of technology. The markup of a component (e.g. a <div class="btn">...) and the styling of it (e.g. .btn { ... }) and the javascript (e.g. document.querySelector('.btn').onclick = ...) will always be highly coupled anyway.

The traditional way to deal with this is to have 3 files (css, html and js)
All very long.
You scroll through many lines of css to find the `.btn { ... }` related css
You scroll through many lines of html to find the `<div class="btn">...`
You scroll through many lines of js to find the `...querySelector('.btn').addEventListner...`

Modern practice is to make self contained components, that include all the css/js/html necessary for one e.g. button component.

So if a change needs to be done to the button, one file or one place contains all button related code html, css and javascript.

It reduces the cognitive load significantly.
Also it helps A LOT with the isolation of components. `csjs` suffixes classnames, so that the styling does not clash with the styling of other components - which can introduce very complicated layout bugs.

I would strongly recommend to not use global variables in css (=global css selectors) and follow the best practice of all programming language that usually consider global variables to be a bad thing.

src/app/file-panel.js
serapath: The `new` keyword doesn't care whether there is an explicit return or not actually. That's part of the spec. I personally prefer explicit over implicit.

I will change it to `this.events = events` but can personally not recommend it.
In JavaScript I consider classes as a legacy that is less flexible than more modern approaches, but if this style is preferred i will stick to it.

Here some links:
* https://twitter.com/peitor/status/845178317233807360
* https://medium.com/javascript-scene/common-misconceptions-about-inheritance-in-javascript-d5d9bab29b0a
* https://medium.com/javascript-scene/javascript-factory-functions-vs-constructor-functions-vs-classes-2f22ceddf33e
* http://wiki.c2.com/?ArgumentsAgainstOop

serapath: i think version **Number 4** is the most traditional version that existed already in ECMA Script 3.

---

#### Usage (my personal recommendation without `this`)
```js
var foobarAPI = {}
var foobarEvents = new EventManager()
var opts = { }
var foobar = FooBar('.container', foobarAPI, foobarEvents, opts)
// but
// var foobar = new FooBar('.container', foobarAPI, foobarEvents, opts) // works exactly the same
```

##### Definition  (my personal recommendation would be something like)
```js
var css = csjs`.foobar { font-size: 20px; }`
function FooBar (container, appAPI, appEvents, opts) {
  if (typeof container === 'string') container = document.querySelector(container)
  if (!container) throw new Error('no container given')
  /* ... */
  var el = bel`<h1 class=${css.foobar}> hello ${opts.name || 'world'} </h1>`
  container.appendChild(el)
  /* ... */
  var api = new EventEmitter()
  api.name = opts.name || 'anoymous'
  api.__proto__ = methods
  return api
}
var methods = {
  bar: function () {},
  baz: function () {}
}
```

PR #563, opened 2017-05-26
Restyle compilation warning

ninabreznik:
screen shot 2017-05-26 at 9 07 51 pm

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -799,7 +799,7 @@ var run = function () {
     config.set('autoCompile', autoCompile)
   })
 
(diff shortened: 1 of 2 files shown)


yann300:
if the compilation is took less than 1000ms, we should hide the warning. now it just remove the content of the error div. I'll push changes if that's fine

ninabreznik:
@yann300 The problem with the span is when you resize, it rearranges the whole checkbox area. Try to make it smaller. That's the reason I though about rather making a new div and put it in the new line...

PR #564, opened 2017-05-30
Remove reduntant scrollbars

ninabreznik:
Right hand panel had many scrollbars. I removed them, but since I have iOS I can't see 100% if all are gone, so please check it out.


ninabreznik:
@yann300 I rebased it now...

PR #570, opened 2017-05-31
readd gas estimation

yann300:


yann300:
web3.eth.gasestimation seems to work again

PR #591, opened 2017-06-10
Clean up of text in Remix UI

greggdourgarian:


yann300:
Thanks for this PR!

src/app/contract-tab.js
```diff
@@ -87,21 +87,18 @@ function contractTab () {
...
       <div class="${css.crow}">
-        <div class="${css.col1_1}">Transaction gas limit</div>
+        <div class="${css.col1_1}">Gas limit</div>
         <input type="number" class="${css.col2}" id="gasLimit" value="3000000">
       </div>
       <div class="${css.crow} hide">
       <div class="${css.col1_1}">Gas Price</div>
         <input type="number" class="${css.col2}" id="gasPrice" value="0">
       </div>
-      <div class="${css.crow}">
```
yann300: why deleting this field?

src/app/staticanalysis/modules/txOrigin.js
```diff
@@ -1,4 +1,4 @@
-var name = 'tx.origin: '
+var name = 'Origin: '
```
yann300: `transaction origin` ?

greggdourgarian: Hi Yann,  Yes, transaction origin, which is how it is described after the lead "Origin: ".   This change is cleaner and leaves no ambiguity about the field being tx.origin.

yann300: hi, I meant that I would just name it `Transaction origin` instead of `Origin`

greggdourgarian:
Hi Yann

Yes, sure.    That works!

________________________________
________________________________

PR #592, opened 2017-06-11
Refactoring: right hand panel

ninabreznik:
- created a righthand panel module
- refactor tabs modules more into direction of 'internal component API standard'

### DIFF
--- a/index.html
+++ b/index.html
@@ -52,25 +52,6 @@
 			<div id="dragbar"></div>
 		</div>
 
-		<div id="righthand-panel">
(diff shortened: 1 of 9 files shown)


yann300:
I will post here some tiny stuff we need to fix (I will try to push some):
 - there's too much margin between tab title and the content (maybe that a cross browser issue).
 - cursor when hovering tab title does not display a pointer anymore.
 - should rename the param `API` to `api` (using uppercase looks like constant)

PR #604, opened 2017-06-22
Compile and run tabs

ninabreznik:
Work in progress 
https://github.com/ethereum/browser-solidity/issues/593


ninabreznik:
Comments from 20.7.

Like:
- compile tab and grey buttons and removed blue from the contract dropdown :+1: 

TODOS:
- in run tab - at address creates instances and create throws some callback error  
- details button (modal dialog has the blue line from right-hand-panel. How to get rid of it? + let's put close button top right (fa-fa-close fontawesome icon) and also copy to clipboard to the title area - copy from instance title))
- add cursor-pointer and onhover opacity 0.8 to copy-to-clipboard icon

yann300:
let's do the modal dialog ui changes in the next PR

PR #606, opened 2017-06-28
Refactor browser-solidity

serapath:
1. use https://github.com/ethereum/browser-solidity/issues/605
2. remove global css and global html
3. restructure layout for the "dom console" feature

### DIFF
--- a/src/app/editor.js
+++ b/src/app/editor.js
@@ -8,22 +8,12 @@ var Range = ace.acequire('ace/range').Range
 require('../mode-solidity.js')
 
 var css = csjs`
-  .editor-container   {
-    display           : flex;
-    position          : absolute;
-    top               : 2.5em;
-    left              : 0;
-    right             : 0;
-    bottom            : 0;
-    min-width         : 20vw;
-  }
   .ace-editor         {
     top               : 4px;
     font-size         : 1.1em;
     width             : 100%;
   }
 `
-document.querySelector('#editor-container').className = css['editor-container']
 
 function Editor (editorElement) {
   var editor = ace.edit(editorElement)
--- a/index.html
+++ b/index.html
@@ -37,21 +37,6 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no">
 	</head>
 	<body>
-
-		<div id="editor">
-			<div id="tabs-bar">
-				<div class="scroller scroller-left"><i class="fa fa-chevron-left "></i></div>
-				<div class="scroller scroller-right"><i class="fa fa-chevron-right "></i></div>
-				<ul id="files" class="nav nav-tabs"></ul>
-			</div>
-			<span class="toggleRHP" title="Toggle right hand panel"><i class="fa fa-angle-double-right"></i></span>
-			<div id="editor-container">
-				<div id="filepanel"></div>
-				<div id="input"></div>
-			</div>
-			<div id="dragbar"></div>
-		</div>
-
 		<script src="build/app.js"></script>
 		
 		<div id="modaldialog" class="modal">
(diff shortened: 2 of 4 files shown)


serapath:
Not sure what the failed check is about. It's related to the new stuff introduced by remixd.

I guess it's about `document.querySelector...` trying to access something to display a modal or append a modal that doesnt yet exist in the DOM. It's a quite tricky rebase, because the `files` object changed in many places and it's now using this `filesProviders` object.

Not too bad, but I already moved a lot of code out of `app.js` and I need a bit more to update all the code that has been touched.

yann300:
test should be fixed now

yann300:
For the DOM relayout not sure this is already in the PR..
as I see the code I guess `fileexplorer_4tpWTH` should also be moved in a new `lefthand-panel' that yet to be added. am I right?

serapath:
yes, you are right

src/app/editor.js
```diff
@@ -23,7 +23,9 @@ var css = csjs`
     width             : 100%;
   }
 `
-document.querySelector('#editor-container').className = css['editor-container']
+setTimeout(_ => {
```
yann300: why this timeout?

serapath: because the refactoring is not done yet. this is what i'm working on right now and it has a lot of implications and quite a lot of code changes.

Currently, the app tend's toward "spagetthi code" in the sense that, no matter where, there are `document.querySelectors` all over the place grabbing random stuff in the DOM and changing it.

The first step of trying to get rid of it was to clear `index.html` from stuff, but now, all kinds of components can't grab nodes with `document.querySelector`, so delaying there actions with `setTimeout` will make them still work - for now.

First goal for now is to shuffle lots of code into it's own components or add them to existing ones, removing code from `app.js` with the goal of removing the `setTimeout`

yann300: ah I see. 
I think we should really get rid of the use of `setTimeout` in this way even if this temporary. 
in `index.js`, 
we could have
```
document.body.appendChild(app.render())
app.init()
```
Refactoring by moving code from app.js to each component is for sure something to do but in a long run.
I would for now stick PRs to make room for the dom-console and not refactor the entire app.js .

serapath: yes that's what this refactoring is all about.
The markup is currently set up in a way that makes it hard to add the console.
That's why i'm doing the refactoring.

The lines you wrote above is exactly what i want to make happen, even though i personally think the `.init()` function is what should happen in the `App(...)` constructor

yann300: sorry I was not precise. I meant that the code that need dom elements to be already rendered (basically the code under your `setTimeout`) could be in an `init` function that you call after the `render` function. In that way you get rid of the `setTimeout`

serapath: i see. that makes sense.
I will update the pull request today, with the refactoring work in it.
It doesn't need the `.init` method anymore, but i understand how this would solve the issue.

serapath:
Hey guys,

This pull request was quite a lot of work. I already talked to @yann300 and after checking how this pull request can be made smaller, I figured it's not easy and would be quite some more work and `app.js` would stay big and more technical debt would be added. I personally would strongly prefer and be a lot more motivated to get feedback about how to improve this pull request to satisfy everyone so that it can be merged than going back to before.

**Pull Request Goal**
Prepare markup for "dom-console" + make `app.js` small by refactoring according to our [code conventions](https://github.com/ethereum/browser-solidity/issues/605#issue-238352529)

**Change Reasons:**
All changes are more or less related to:
1. moving `js/html/css` out of global scope/files & `app.js` to it's own or existing module(s) it belongs to
2. or refactoring from `container.appendChild(...)` style to `instance.render()` style

*Here are the links for the 2 screencasts in which i explain the pull request.*
* [Screencas1](https://drive.google.com/file/d/0Bz-JDJ8tQeKWX0EzQU03Ukp4M1U/view) (compare old `app.js` file to the new one) **[~4min]**
* [Screencast2](https://drive.google.com/file/d/0Bz-JDJ8tQeKWN2c1cm1HSG1aeWc/view) (go through the pull request in detail explaining all changes to the files) **[~8min]**

PR #609, opened 2017-06-29
adding gotoLine to the remix scroller

ryestew:


src/app/editor.js
```diff
@@ -69,6 +69,12 @@ function Editor (editorElement) {
     editor.session.setBreakpoint(row, css)
   }
 
+  this.setPosition = function (line, column, noClip) {
+    editor.moveCursorTo(line, column)
```
chriseth: Is it possible to scroll into view without moving the cursor?

PR #612, opened 2017-07-03
Transaction Listener

yann300:
That module record transactions coming from the network and try to resolve them against the solidity compilation result.
Should be used by the dom-console.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -29,7 +29,7 @@ var FilePanel = require('./app/file-panel')
 var EditorPanel = require('./app/editor-panel')
 var RighthandPanel = require('./app/righthand-panel')
 var examples = require('./app/example-contracts')
-var Txlistener = require('./app/txListener')
+// var Txlistener = require('./app/txListener')
 
 var css = csjs`
   html { box-sizing: border-box; }
@@ -849,7 +849,8 @@ function run () {
   node.insertBefore(staticanalysis.render(), node.childNodes[0])
 
   // ----------------- Tx listener -----------------
-
+  // Commented for now. will be used later.
+  /*
   var txlistener = new Txlistener({
     api: {
       web3: function () { return executionContext.web3() },
@@ -885,6 +886,7 @@ function run () {
       resolvedTransaction: resolvedTransaction
     })
   })
+  */
 
   // ----------------- autoCompile -----------------
   var autoCompile = document.querySelector('#autoCompile').checked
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -889,6 +890,47 @@ function run () {
   var node = document.getElementById('staticanalysisView')
   node.insertBefore(staticanalysis.render(), node.childNodes[0])
 
+  // ----------------- Tx listener -----------------
+
+  var txlistener = new Txlistener({
+    appAPI: {
```
serapath: For all modules I started to call it just `api` or `events`, like:
```js
var txlistener = new Txlistener({
  api: { /* .... */ },
  events: { app: self.event, udapp: udap.event /* , .... */ }
})
```
To me this looks cleaner and communicates the same message :-)

src/app/txListener.js
```diff
@@ -0,0 +1,267 @@
...
+  * poll web3 each 2s if web3
+  * listen on transaction executed event if VM
+  * attention: blocks returned by the event `newBlock` have slightly different json properties whether web3 or the VM is used
+  * trigger 'newBlock'
+  *
+  */
+class TxListener {
+  constructor (opt) {
+    this.event = new EventManager()
+    this._appAPI = opt.appAPI
```
serapath: I also change the signature and constructore like this:
```js
class TxListener {
  constructor (opts = {}) { // in case the constructor is called with no arguments, because they are all optional
    this.event = new EventManager()
    this._api = opts.api
  }
  /* .... */
}
```

serapath: i assume that the way the code looks like now - it's broken, because the `this` might not refer to what `self` is refering to, which should probably be the `this` of the `constructor`

src/app/txListener.js
```diff
@@ -0,0 +1,267 @@
...
+          })
+        })
+      }
+    })
+  }
+
+  /**
+    * reset recorded transactions
+    */
+  init () {
```
serapath: The `init` method looks like all of this could be already happen in the `constructor`.
So maybe it's not necessary to have the `init` method?

src/app/txListener.js
serapath: **aha, no `this` would work i guess**
When the handler is an arrow function, `this` is set to the surrounding `this` i think i remember... :/ ..ok, then it's probably fine. Just a bit confusing to my eyes, but that's just old habits i think

src/app.js
yann300: ok

src/app/txListener.js
yann300: `init()` is also used in `startlistening()`

PR #613, opened 2017-07-05
FIX file explorer layout related styling bug

serapath:


yann300:
tested locally. tests are failing due to a bug in solcjs (error formatting)

PR #614, opened 2017-07-05
re-layout [WIP]

yann300:
This will basically relayout the position of left / center / right panel. So we can add the dom-console.

still in progress : 

TODO : 
 - Still some ui glitch.
 - The editor is not showing up on maximum of the his height.
 - resize functions has been removed (to be replaced by the resize algo of https://github.com/ethereum/browser-solidity/pull/606)


yann300:
@serapath takes over in another PR

PR #617, opened 2017-07-11
Support for SMT queries.

chriseth:
Uses https://github.com/ethereum/remixd/pull/1


axic:
@chriseth did we decide on a different interface in the end obsoleting this one?

axic:
Yep, it was replaced by #873.

PR #630, opened 2017-07-18
REFACTOR (3) add rightpanel & centerpanel to layout

serapath:


src/app.js
```diff
@@ -66,6 +73,13 @@ class App {
     var self = this
     if (self._view.el) return self._view.el
     /***************************************************************************/
+    self._view.rightpanel = yo`<div></div>`
+    self._view.centerpanel = yo`
+      <div id="editor-container" class=${css['editor-container']}>
+        <div id="filepanel" class=${css['filepanel-container']}></div>
```
yann300: that should be in the left-panel no?

src/app.js
```diff
@@ -694,7 +707,9 @@ function run () {
     app: self.event,
     udapp: udapp.event
   }
-  var righthandPanel = new RighthandPanel(document.body, rhpAPI, rhpEvents, {}) // eslint-disable-line
+  var righthandPanel = new RighthandPanel(rhpAPI, rhpEvents, {}) // eslint-disable-line
+  self._view.rightpanel.replaceWith(righthandPanel.render())
```
yann300: ... I would rather make the `rightPanel` the container of righthandpanel instead of replacing

src/app/righthand-panel.js
```diff
@@ -118,7 +121,9 @@ function RighthandPanel (container, appAPI, events, opts) {
   })
 
   if (appAPI.config.exists(EDITOR_WINDOW_SIZE)) {
-    self._api.setEditorSize(appAPI.config.get(EDITOR_WINDOW_SIZE))
+    setTimeout(function () {
```
yann300: why this timeout? is this still because the editor is not yet in the DOM?

serapath: yes

serapath: i think the whole layout and collapsing and drag resizing system is a little bit broken and needs and overhaul. There seems to be a lot of code (way more then necessary) in often half random positions to make it work.

In the big refactor pull request i already did one more serious refactoring, but I think it can even be better than that. So maybe we could open an issue for that to make all of this more nice in the future

src/app.js
serapath: oh yes, this comes with the next pull request

serapath: just didn't want to stuff too much stuff into one pull request

src/app/righthand-panel.js
yann300: and this will be removed in the next PR right?
from line 83 to the end.

src/app.js
yann300: Please just make `self._view.rightpanel` a container that supports resizing (or will support in the next PR) and `RightHandpanel` a module that just fit it in his container (and thus follow the size of the container).

src/app.js
yann300: ok cool

src/app.js
serapath: yes, this is also just temporary - because changing this will cause a lot of lines to change :/

serapath: because adding `RighthandPanel` in the `.render()` function and not in the `.init()` call also requires ALL the dependencies of RighthandPanel to be already initialized ... this has a lot of ripple effects, so i would rather slowly go there to not again end up with a huge pull request :/

src/app/righthand-panel.js
serapath: yes this is all temporary changes that will be removed with more refactoring.
The problem is the `app.init()` function, which is called after `app.render()` in `index.js` and enables to not use `setTimeout`.

Anyway, there should neither be a `setTimeout` nor `app.init()` function, but fixing this everywhere has a lot of "ripple effects" so i would rather slowly refactor all of this and add here and there "temporary hacks" that make a pull request work, but prevent that a single pull request to become very large.

PR #632, opened 2017-07-18
REFACTOR (5) put centerpanel's editor and tabs into editorpanel component

serapath:


src/app.js
```diff
@@ -728,17 +687,17 @@ function run () {
...
+  self._components.righthandpanel.event.register('resize', delta => self._adjustLayout('right', delta))
 
   // ----------------- editor resize ---------------
 
   function onResize () {
     editor.resize(document.querySelector('#editorWrap').checked)
-    reAdjust()
   }
-  window.onresize = onResize
+  self._onResize = window.onresize = onResize
```
yann300: is `_onResize` still used ?

src/app.js
```diff
@@ -43,44 +44,89 @@ var css = csjs`
...
+          show: true
+        }
+      }
+    }
+    // ----------------- editor ----------------------------
+    self._components.editor = new Editor({}) // @TODO: put into editorpanel
+    // ----------------- editor panel ----------------------
+    self._components.editorpanel = new EditorPanel({
+      api: { editor: self._components.editor },
+      data: { _layout: self.data._layout }
```
yann300: `data` seems not to be used in `EditorPanel` . perhaps just remove ?

src/app.js
```diff
@@ -43,44 +44,89 @@ var css = csjs`
...
     var self = this
+    self._api = {}
+    var fileStorage = new Storage('sol:')
+    self._api.config = new Config(fileStorage)
+    self._api.filesProviders = {}
+    self._api.filesProviders['browser'] = new Browserfiles(fileStorage)
+    self._api.filesProviders['localhost'] = new SharedFolder(new Remixd())
     self._view = {}
+    self._components = {}
+    self.data = {
```
yann300: what kind of container'll be `data` ?
could we just have  `self.layout = ... ` ? `data` is a bit too general

src/app/editor-panel.js
```diff
@@ -0,0 +1,241 @@
...
+class EditorPanel {
+  constructor (opts = {}) {
+    var self = this
+    self.data = {
+      _FILE_SCROLL_DELTA: 200
+    }
+    self._view = {}
+    self._api = { editor: opts.api.editor }
+    self.event = new EventManager()
+    // var events = opts.events
```
yann300: remove commented line ?

src/app.js
```diff
@@ -728,17 +687,17 @@ function run () {
     app: self.event,
     udapp: udapp.event
   }
-  var righthandPanel = new RighthandPanel(rhpAPI, rhpEvents, {}) // eslint-disable-line
-  self._view.rightpanel.appendChild(righthandPanel.render())
-  righthandPanel.init()
+  self._components.righthandpanel = new RighthandPanel(rhpAPI, rhpEvents, {})
```
yann300: why keeping `self._components.righthandpanel` ? because it is not used (and `filePanel` which have the same role is not kept in `self._component`) . 
Also in the future, would be nice to rename `righthandpanel` to maybe `tabsPanel` (to more consistent with `filePanel`)

yann300:
tests are failing (will take a look)

yann300:
there is a refactor to do in the next PRs:
the resizing logic is duplicated between `right hand panel` and `file panel`
https://github.com/ethereum/browser-solidity/pull/632/files#diff-24c104a490d010af7e543c8d8e4ef4caR119
https://github.com/ethereum/browser-solidity/pull/632/files#diff-fa6201e51c5f6698de02976b5f957545R197

src/app.js
serapath: I'm not sure.... looks like it is not.
I'll remove it then

src/app.js
serapath: i see. yeah, i tried a few versions how to best make the "component dance" to synchronize layout when users decide to drag and resize panels.
I'll remove it too

src/app.js
serapath: I want it to contain all state. `data` is meant to always be serializable to JSON.

The other properties I'm using are:
`self._view` which is kind of a cache for currently rendered DOM elements the component uses
`self._components` which references all instantiated children components the component uses
`self._api` & `self._events` which contains methods and events passed in by the parent
and then, the component itself exports an API to the instantiating parent.
This api includes `self.event` where the parent can register and it might include more methods...

Probably `self.data` should be `self._data` ..but for now i was just **underscoring** certain props on `self.data` to indicate those as private (component internal) state

src/app/editor-panel.js
serapath: sure

src/app.js
serapath: Oh, that's a mistake :-) `filepanel` should also be `self._components.filepanel`

I agree with renaming the righthandpanel to tabspanel or maybe even in the spirit of browsers `devtoolspanel` :-)

I just try to standardize the codebase more, so that no matter where any existing or new programmer will go, he will find familiar patterns so it's easier to expect whats going on and where things are.

Maybe the **issue** where we documented "best practices" should become a markdown file in a github repository so it's possible to make PULL REQUESTS to our best practices.

PR #645, opened 2017-07-31
Event Decoder

yann300:
Just a little refactoring of the current way to decode event.
That will be used by the console

### DIFF
--- a/src/app/eventsDecoder.js
+++ b/src/app/eventsDecoder.js
@@ -1,5 +1,4 @@
 'use strict'
-var $ = require('jquery')
 var ethJSABI = require('ethereumjs-abi')
 
 /**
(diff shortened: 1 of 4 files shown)


src/app/eventsDecoder.js
```diff
@@ -0,0 +1,82 @@
...
+    if (!contract || !receipt.logs) {
+      return cb()
+    }
+    this._decodeEvents(tx, receipt.logs, contract, contracts, cb)
+  }
+
+  _eventABI (contractName, compiledContracts) {
+    var contractabi = JSON.parse(compiledContracts[contractName].interface)
+    var eventABI = {}
+    $.each(contractabi, function (i, funABI) {
```
serapath: is `contractabi` an array?  ...and is this the only place where `jquery` is used?
if so ... you could get rid of the dependency by replacing it with a native
```js
contractabi.forEach(function (funABI, i) {
  // ... do stuff ...
})

PR #654, opened 2017-08-05
Add best-practices.md

ninabreznik:
Added best practices into this document that we can then update in the future and because we will use pull requests we can always roll back to the history of the document.


serapath:
LGTM

@yann300 

can this be merged? I'd like to make a change proposal to include `jsdoc` comments, but i'll wait until it's merged i think.

yann300:
made some changes:
 - replaced self.data by self.state
 - explained a bit what properties/conventions are for

PR #655, opened 2017-08-06
ADD terminal (first draft)

serapath:


src/app/terminal.js
```diff
@@ -0,0 +1,301 @@
...
+    return entries
+  }
+  scroll2bottom () {
+    var self = this
+    setTimeout(function () {
+      self._view.term.scrollTop = self._view.term.scrollHeight
+    }, 0)
+  }
+  _serialize (obj) { return JSON.stringify(obj, null, 2) }
+  _shell (text) { // default shell
```
yann300: is this for the support of `web3` ?

src/app/terminal.js
```diff
@@ -0,0 +1,301 @@
...
+      // console.log(event.returnValue)
+      // console.log('event.returnValue')
+      // console.log(event.type)
+      // console.log('event.type')
+      // console.log('-----')
+      // console.log(event)
+    }
+  }
+  log (obj) {
+    var self = this
```
yann300: `obj` could also be a node element. 

And could you document those function? like
```
/**
 * what the function does
 *
 * @param {String} txHash    - hash of the transaction
 */
 ```

File: src/app/terminal.js [hunk 3]
```diff
@@ -0,0 +1,301 @@
...
+        ${entries.map(str => document.createTextNode(`${str}\n`))}
+      </div>
+    `
+    self._view.log.appendChild(block)
+    self.scroll2bottom()
+    return entries
+  }
+  scroll2bottom () {
+    var self = this
+    setTimeout(function () {
```
yann300: why `setTimeout` ?

yann300:
cool!
A few comments about the ui (but that could be done in a subsequent PR)
- we should use the same rendering as for the chrome devtool console
- the horizontal drag bar is too large (that because of the collapse button). an idea: setting the bar height as minimun as possible and put the collapse button in the upper-right corner of the console.
- when resizing, the drag bar takes the entire browser width...

src/app/terminal.js
yann300: we really need this to be addressed before merging. cause the `txlistener` needs to log a node element (containing the action to start debugging) and not only text.

src/app/terminal.js
serapath: oh no, this is basically for interpreting the input a user writes into the terminal.
It's not yet added.
Yes, it will "know" about the `web3` object

src/app/terminal.js
serapath: yes, `obj` can also be a node element.

Yes I guess i need comments to document the function.
Usually i put everything into it's own module and document it in the readme, but here we kind of have a "mono repo", so i'll try to get into the habit of documenting it jsdocs style

src/app/terminal.js
yann300: could we add to this PR a basic version where when you click enter it will copy the content of the input to the console / and clear the input ?

src/app/terminal.js
yann300: I see but as I see the code, `obj` can't be a node element

src/app/terminal.js
serapath: it's an issue of terminal input getting initialized while the terminal is not yet in the DOM, so scroll2bottom uses a size of 0 to scroll2bottom ...which does nothing :/

src/app/terminal.js
serapath: yes

src/app/terminal.js
serapath: not yet

PR #670, opened 2017-08-08
reorganize project folders

yann300:


serapath:
I guess i would try to find folder names according to what could become it's own repository in the future.
And i would even add a `README.md` file into each of those folders describing the module.

Then once a folder becomes mature enough, i would actually make a seperate repository from it and `npm install` it back into `browser-solidity`.

PR #673, opened 2017-08-09
TxLogger

yann300:
Module that just listen on new tx and log them.


yann300:
basically if the transaction is sent to a known contract, this function is called: 
`renderResolvedTransaction` 
This should display smt like:
`from: 0x445a...123c, to: <contract_name> (0x445a...123c), value: <value>, function:<function_name(string, string)>, x logs - [Debug it]`

If the contract is not known, then `renderTransaction` is called
This should display smt like:
`from: 0x445a...123c, to: 0x445a...123c, , value: <value>, data: <data> - [Debug it]`

serapath:
looks good to me. most of the code i dont really understand what it's doing, but generally i looks fine to me :-)

PR #679, opened 2017-08-10
ADD terminal

serapath:


src/app/panels/terminal.js
```diff
@@ -0,0 +1,420 @@
...
+      range.collapse(toEnd)
+
+      var sel = window.getSelection()
+      sel.removeAllRanges()
+      sel.addRange(range)
+
+      editable.focus()
+    }
+  }
+  _log (mode) {
```
yann300: could you describe a bit what this function is doing?
cause I don't think that `if (types[idx] === 'element') val = jsbeautify.html(val)` is needed. is `element` refers to dom element?

src/app/panels/terminal.js
```diff
@@ -0,0 +1,420 @@
...
+    var self = this
+    var modes = { log: true, info: true, error: true }
+    if (modes[mode]) {
+      return function logger () {
+        var args = [].slice.call(arguments)
+        self.data.session.push(args)
+        var types = args.map(type)
+        var values = javascriptserialize.apply(null, args).map(function (x, i) {
+          return (typeof args[i] === 'string') ? args[i] : x
```
yann300: should it be `return (typeof type[i] === 'string') ? args[i] : x` ?

src/app/panels/terminal.js
```diff
@@ -0,0 +1,420 @@
...
+  }
+  _log (mode) {
+    var self = this
+    var modes = { log: true, info: true, error: true }
+    if (modes[mode]) {
+      return function logger () {
+        var args = [].slice.call(arguments)
+        self.data.session.push(args)
+        var types = args.map(type)
+        var values = javascriptserialize.apply(null, args).map(function (x, i) {
```
yann300: if this is a dom element, we don't want to serialize but just to append the dom node.

src/app/panels/terminal.js
serapath: hmm, interesting. yes I think you might be right xD wow... good catch. you crazy guy actually read and try to understand the code? :P haha

Was I drunk? I actually have no clue what this should do. It makes no sense to me now.
Like `typeof args[i] === 'string'` should always be **true** because that's the whole point of `javascriptserialize` :p 

I copied it from my module and adapted it, see here: https://github.com/serapath/dom-console/blob/master/domconsole.js#L195

Thx, for spotting it, i'll just remove it.

PR #683, opened 2017-08-14
Fix no contract instance text

yann300:


yann300:
The 'no create instances' label was a bit broken

PR #692, opened 2017-08-15
warn user if deployed contract doest not correspond to compiled one

yann300:


yann300:
still to do: manage the case where the contract has been added using "at address"

PR #696, opened 2017-08-17
Fix sort ordering logic

MidnightLightning:
Having two separate `sort` calls doesn't actually preserve the ordering of the first sort (by name) so only the second (constant vs. not) was being sorted on properly. This fix first checks to see if the two items have different `constant` values, and only if their `constant` values are the same does it then sort by name.

### DIFF
--- a/src/app/execution/txHelper.js
+++ b/src/app/execution/txHelper.js
@@ -32,6 +32,9 @@ module.exports = {
   },
 
   sortAbiFunction: function (contractabi) {
+    // Sorts the list of ABI entries. Constant functions will appear first,
+    // followed by non-constant functions. Within those t wo groupings, functions
+    // will be sorted by their names.
     return contractabi.sort(function (a, b) {
       if (a.constant === true && b.constant !== true) {
         return 1
--- a/src/app/execution/txHelper.js
+++ b/src/app/execution/txHelper.js
@@ -32,16 +32,17 @@ module.exports = {
   },
 
   sortAbiFunction: function (contractabi) {
-    var abi = contractabi.sort(function (a, b) {
-      if (a.name > b.name) {
-        return -1
-      } else {
-        return 1
+    return contractabi.sort(function (a, b) {
+      if (a.constant === true && b.constant !== true) {
+        return 1;
+      } else if (b.constant === true && a.constant !== true) {
+        return -1;
       }
-    }).sort(function (a, b) {
-      if (a.constant === true) {
-        return -1
-      } else {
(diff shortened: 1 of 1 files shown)


yann300:
thanks for that! we will push it.
a little thing: `a.name` can be undefined (if constructor) so it need to be tested.

src/app/execution/txHelper.js
```diff
@@ -28,20 +28,20 @@ module.exports = {
...
-    }).sort(function (a, b) {
-      if (a.constant === true) {
+      } else if (b.constant === true && a.constant !== true) {
         return -1
-      } else {
+      }
+      // If we reach here, either a and b are both constant or both not; sort by name then
+      // special case for fallback and constructor      
+      if (a.type === 'function') {
+        return a.name.localeCompare(b.name)
```
axic: Hm, this shouldn't be locale specific I think. Also identifiers can be ASCII only.

MidnightLightning: `localeCompare()` is the only native string-comparison function in Javascript; would you rather multiple separate calls to check greater, less, and equal?

axic: Since this is only used for display purposes it should be fine, though I'd add a comment.

If this would be critical piece of code, either:
- a manual implementation
- or setting the locale to ASCII English upfront

would be needed, because certain languages consider di/trigraphs with different sorting order.

axic:
@MidnightLightning are you interested in finishing this? Would be a good feature to merge.

MidnightLightning:
@axic Thanks for the reminder; updated now!

yann300:
@MidnightLightning could you rebase?

MidnightLightning:
@yann300 Rebased now.

MidnightLightning:
@yann300 Coding standard errors fixed now.

yann300:
great! thanks

MidnightLightning:
@yann300 Any further changes needed on this one? GitHub's still indicating you as a reviewer have changes requested.

src/app/execution/txHelper.js
```diff
@@ -32,20 +32,20 @@ module.exports = {
   },
 
   sortAbiFunction: function (contractabi) {
```
axic: Can you please add a comment above here explaining the order? e.g. constructor, fallback, constant functions, non-constant functions?

MidnightLightning: @axic Okay, added now.

axic:
@yann300 anything left on this?

yann300:
moved, thanks @MidnightLightning

PR #699, opened 2017-08-17
Resolve truffle like import

yann300:


foxundermoon:
why not merge to master?

yann300:
I think there's a better way to handle it... I can switch to that one soon , I will create an issue

itstehkman:
Hi @yann300, is this something that is already supported / will be supported soon? I've gotten this feedback about Remix is that it is hard to resolve contracts that come from npm install, such as openzeppelin-solidity. This small PR would be a big win for productivity!

yann300:
this is already supported @itstehkman 
please try installing and run https://github.com/ethereum/remixd#share-a-folder

PR #705, opened 2017-08-18
Remove legend, add info about type (call/payable/not payable, restyle…

ninabreznik:
… pending)

### DIFF
--- a/src/universal-dapp.js
+++ b/src/universal-dapp.js
@@ -334,7 +334,7 @@ UniversalDApp.prototype.getCallButton = function (args) {
 
   if (lookupOnly) {
     $contractProperty.addClass('constant')
-    call()
+    button.attr('title', (title + " - call)"))
   }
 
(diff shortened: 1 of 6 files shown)


ninabreznik:
@yann300 Can you please check if call/payable/not payable display correctly?

src/universal-dapp.js
```diff
@@ -334,7 +334,7 @@ UniversalDApp.prototype.getCallButton = function (args) {
 
   if (lookupOnly) {
     $contractProperty.addClass('constant')
-    call()
```
yann300: you can't remove this . this function get the return value of a call just after the instance is added

src/universal-dapp.js
```diff
@@ -343,6 +343,11 @@ UniversalDApp.prototype.getCallButton = function (args) {
 
   if (args.funABI.payable === true) {
     $contractProperty.addClass('payable')
```
yann300: you also need to manage the case where `args.funABI.constant === true`

PR #720, opened 2017-08-25
Also treat view and pure functions as constant.

chriseth:


src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -442,7 +442,11 @@ function isStateVariable (name, stateVariables) {
  * @return {bool}
  */
 function isConstantFunction (node) {
-  return isFunctionDefinition(node) && node.attributes.constant === true
+  return isFunctionDefinition(node) && (
+    node.attributes.constant === true ||
+    node.attributes.stateMutability === 'view' ||
+    node.attributes.stateMutability === 'pure'
```
axic: I wonder why this is needed since `constant` is always set. Though for forward compatibility it should be done this way.

src/app/staticanalysis/modules/staticAnalysisCommon.js
```diff
@@ -32,7 +32,7 @@ var basicRegex = {
   CONTRACTTYPE: '^contract ',
   FUNCTIONTYPE: '^function \\(',
   EXTERNALFUNCTIONTYPE: '^function \\(.*\\).* external',
-  CONSTANTFUNCTIONTYPE: '^function \\(.*\\).* constant',
+  CONSTANTFUNCTIONTYPE: '^function \\(.*\\).* \\(constant|view|pure\\)',
```
axic: What is this regex doing exactly? Trying to match it based on specifiers? Isn't the AST enough?

yann300: it check the `Identifier` node (parent of `Function Call` . type attribute contains `function () view`

axic: But should it use purely the AST for that?

src/app/staticanalysis/modules/staticAnalysisCommon.js
chriseth: That is the intention.

src/app/staticanalysis/modules/staticAnalysisCommon.js
chriseth: For a function call, it takes a look at the type of the function that is called.

chriseth: This regex is not applied to the source but to the type string.

PR #726, opened 2017-08-25
Fix the problem with name that created a bug

ninabreznik:
Error in the naming of file: changed from helper.js to helpers.js


yann300:
done by https://github.com/ethereum/browser-solidity/pull/727

PR #738, opened 2017-08-30
Fix scroll on file explorer

yann300:


serapath:
LGTM (looks good to me)

ninabreznik:
I checked, have to say, I wouldn't implement it. 

If you want to see the full name, you click on the file and you see up the full name.

Otherwise, when you have a long name in one file, this scrollbar stays on the screen all the time and it looks weird, to be honest.

screen shot 2017-09-01 at 12 53 36 am

ninabreznik:
And normally it looks like this

screen shot 2017-09-01 at 12 54 17 am

And with a long name, you see it in the tab

screen shot 2017-09-01 at 12 55 17 am

ninabreznik:
I see that for deleting is important that you can scroll to the end, but it doesn't work in this test I made, for example.

screen shot 2017-09-01 at 1 14 33 am

I suggest, we shorten the name in the file explorer to x char and full name is visible in the tab and in the file explorer you can see the delete icon, because we shorten the name

PR #743, opened 2017-08-30
shortened strings and some text fixes for title of each key in details

ninabreznik:


yann300:
just need to check whether `metadata` is can be parsed. In case of an abstract contract, it is empty `""`

PR #746, opened 2017-08-31
add dropdown menu

serapath:


src/app/ui/dropdown.js
```diff
@@ -0,0 +1,141 @@
...
+*/
+class Dropdown {
+  constructor (opts = {}) {
+    var self = this
+    self.event = new EventManager()
+    self.data = {
+      _options: opts.options || ['script'],
+      selected: opts.defaults || []
+    }
+    if (self.data._options.indexOf('script') !== -1) {
```
yann300: This part is specific to one particular use of a dropdown . can this be removed?

serapath: yes sure. it should be removed i guess

PR #749, opened 2017-08-31
Update editor panel

serapath:


src/app/panels/editor-panel.js
```diff
@@ -221,7 +221,7 @@ class EditorPanel {
   }
   log () {
     var self = this
-    self._components.terminal.log.apply(self._components.terminal, arguments)
+    self._components.terminal._output.apply(self._components.terminal, arguments)
```
yann300: `_output` is not a function of terminal at this point?

serapath: true
...so maybe dont merge this one yet, but after the terminal. i'm preparing it right now

serapath: i added on https://github.com/ethereum/browser-solidity/pull/750 to contain both methods
`terminal.log` forwording to `terminal._output` ...so that https://github.com/ethereum/browser-solidity/pull/750 can be merged while this one here uses still **.log(...)** but when this one here is merged to use **._output(...)** it will not break.

After the merge i will update the terminal and remove support for `terminal.log(...)`

PR #753, opened 2017-08-31
add terminal menu bar with clear/filter/...

serapath:


src/app/panels/terminal.js
```diff
@@ -461,10 +509,6 @@ class Terminal {
       self._output.apply(self, args)
     }
   }
-  log () {
```
yann300: can we keep it for now? `TxLogger` is using that function

serapath: oh i see.
i'm anyway restructuring again. I'll update this PR soon.
I'm optimizing the log structure, so that it makes the filtering easy.
While developing, it easily happens, that my terminal fills with 10k or 100k ...i even had already 1million transaction messages... and that make it slow, so i have to tune this a bit, especially when you want to go and filter them

PR #771, opened 2017-09-06
Terminal Menu Bar -  UI

yann300:
- small ui changes
 - 3 options : only remix transactions / all transactions / script (only remix + all are linked)


serapath:
interesting.
yes - looks good to me

PR #793, opened 2017-09-14
Modal prompt confirm

ryestew:
I'm calling modalDialogCustom.prompt from app.js 
I've got a callback function now as part of the parameters.  Is that enough?

### DIFF
--- a/src/app/files/file-explorer.js
+++ b/src/app/files/file-explorer.js
@@ -1,4 +1,4 @@
-/* global FileReader, confirm */
+/* global FileReader */
 var yo = require('yo-yo')
 var csjs = require('csjs-inject')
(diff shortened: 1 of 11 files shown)


src/app/files/file-explorer.js
```diff
@@ -144,9 +144,7 @@ function fileExplorer (appAPI, files) {
     }
 
     var name = files.type + '/' + file.name
-    // this next line looks wrong
-    console.log('look at this - (!files.exists(name) || files.exists(name))')
-    if (!files.exists(name) || files.exists(name)) {
+    if (!files.exists(name)) {
```
ryestew: this is what I have locally

PR #802, opened 2017-09-20
Record Transactions

yann300:


src/app/tabs/run-tab.js
```diff
@@ -243,6 +244,59 @@ function updateAccountBalances (container, appAPI) {
 }
 
 /* ------------------------------------------------
+    RECORDER
```
yann300: Is that possible to put this in a separate file?

serapath: It's ...probably possible.
The `recorder.js` is already seperate.
The `run-tab.js` has buttons on it that use the recorder.

This logic is only about the buttons and where they are positioned and what they do.

I initially started to give the `recorder.js` a `.render()` method, but then felt, the design might very much depend on the panel and how to position the buttons inside the panel.

If you really think it should be put into a seperate file? should i reconsider the `.render()` or something else?

**implication:**
It would mean, i also need to pass or require `modelDialog` and `copy2clipboard` to recorder.

serapath: so to answer your question: **no** (i don't have a good idea how to do that)

*(but yes, you can always make a separate file and put stuff in - if you have a good idea then just let me know)*

src/recorder.js
```diff
@@ -0,0 +1,53 @@
...
+class Recorder {
+  constructor (opts = {}) {
+    var self = this
+    self._api = opts.api
+    self.event = new EventManager()
+    self.data = { journal: [], _pending: {} }
+    opts.events.executioncontext.register('contextChanged', function () {
+      self.clearAll()
+    })
+    var counter = 1
```
yann300: Why not starting from 0? cause the first account will <account - 2>... see line 18

src/universal-dapp.js
```diff
@@ -537,11 +537,13 @@ UniversalDApp.prototype.runTx = function (args, cb) {
     },
     // run transaction
     function (callback) {
+      var stamp = Date.now()
```
axic: Why not `timestamp`?

src/recorder.js
```diff
@@ -0,0 +1,53 @@
...
+    self._addressCache = {}
+    opts.events.udapp.register('initiatingTransaction', (stamp, tx) => {
+      var { from, to, value, gas, data } = tx
+      var record = { value, gas, data }
+      record.from = self._addressCache[from] || (self._addressCache[from] = `<account -${(++counter)}>`)
+      if (to === null) self.data._pending[stamp] = record
+      else record.to = self._addressCache[to] || (self._addressCache[to] = `<account -${(++counter)}>`)
+      self.append(stamp, record)
+    })
+    opts.events.udapp.register('transactionExecuted', args => {
```
yann300: this event has more than one arg:
`transactionExecuted', [error, args.from, args.to, args.data, false, result, stamp]`

src/recorder.js
```diff
@@ -0,0 +1,53 @@
...
+    opts.events.executioncontext.register('contextChanged', function () {
+      self.clearAll()
+    })
+    var counter = 1
+    self._addressCache = {}
+    opts.events.udapp.register('initiatingTransaction', (stamp, tx) => {
+      var { from, to, value, gas, data } = tx
+      var record = { value, gas, data }
+      record.from = self._addressCache[from] || (self._addressCache[from] = `<account -${(++counter)}>`)
+      if (to === null) self.data._pending[stamp] = record
```
yann300: I think `_pending` should be named `_pendingCreation`

src/recorder.js
```diff
@@ -0,0 +1,53 @@
...
+      else record.to = self._addressCache[to] || (self._addressCache[to] = `<account -${(++counter)}>`)
+      self.append(stamp, record)
+    })
+    opts.events.udapp.register('transactionExecuted', args => {
+      var err = args[0]
+      if (err) console.error(err)
+      var stamp = args[6]
+      var record = self._pending[stamp]
```
yann300: we should make more obvious that we are dealing with creation transaction here

src/universal-dapp.js
serapath: sure i'll update it. Did I even write it?
I actually called it `stamp` everywhere, because I thought you named it `stamp` in "universal-dapp" :P

src/app/ui/modal-dialog-custom.js
```diff
@@ -11,10 +11,13 @@ module.exports = {
   alert: function (text) {
     modal('', yo`<div>${text}</div>`, null, { label: null })
   },
-  prompt: function (title, text, inputValue, ok, cancel) {
+  prompt: function ({ title, text, inputValue, multiline }, ok, cancel) {
```
yann300: doing that you break other calls to `prompt` ... could you just make it another function in here please ?

src/universal-dapp.js
```diff
@@ -461,103 +460,103 @@ UniversalDApp.prototype.pendingTransactions = function () {
...
-      } else {
-        self.getAccounts(function (err, ret) {
-          if (err) {
-            return callback(err)
-          }
+UniversalDApp.prototype.replayTx = function (args, cb) {
+  var self = this
+  self.getAccounts(function (err, accounts = []) {
+    if (err) console.error(err)
+    if (args.to && args.to[0] === '<') args.to = accounts[args.to.split('>')[0].slice(11)]
```
yann300: I think that part of the job should be done in the recorder

src/universal-dapp.js
```diff
@@ -461,103 +460,103 @@ UniversalDApp.prototype.pendingTransactions = function () {
...
+}
 
-          callback()
-        })
-      } else {
-        self.getAccounts(function (err, ret) {
-          if (err) {
-            return callback(err)
-          }
+UniversalDApp.prototype.replayTx = function (args, cb) {
```
yann300: This function should do the same as `runTx` except that the pipeline should be
[queryGasLimit, runTransaction]
I m not sure the function name fit as well...

src/app/tabs/run-tab.js
```diff
@@ -239,6 +240,84 @@ function updateAccountBalances (container, appAPI) {
...
+      //     "bloom": { "bitvector": { "type": "Buffer", "data": [0, /* ... */ 0, 0] } },
+      //     "amountSpent": "5318"
+      //   },
+      //   "transactionHash": "0x84f68f96944a47b27af4b4ed1986637aa1bc05fd7a6f5cb1d6a53f68058276d8"
+      // }
+      /*
+        at each callback call, if the transaction succeed and if this is a creation transaction, we should call
+
+        runtab.addInstance( ... ) // which basically does:
+        instanceContainer.appendChild(appAPI.udapp().renderInstance(contract, address, selectContractNames.value))
```
yann300: at this point if we have a contract creation we should add a new instance

src/universal-dapp.js
serapath: How do you mean that?

serapath: I thought we will even give the user an option to set custom addresses.
So we don't want to retrieve information like `<account - 1>` from the recorder?

src/universal-dapp.js
serapath: So `replayTx` is not a good function name?

Otherwise, the function does exactly the same as `runxTx` but has the pipeline:
`[queryGasLimit, runTransaction]`, no?

What is otherwise different?

src/app/tabs/run-tab.js
serapath: ok. but i guess the "op" bug needs to be fixed to test this

src/universal-dapp.js
yann300: the recorder takes care of replacing real address to <account - n> , so i guess this is also his job to convert the other way around (from <account - n> to real address).
I think the `universal-dapp` does not have to know about anything related to these translations.
the udapp just takes a tx that is quasi filled with all the needed param (except gasLimit).

For setting custom addresses yes we can do that as well in the recorder or another modules yet to be coded, but not in udapp.

src/universal-dapp.js
yann300: "What is otherwise different?" : resolving of `<account -n>` .

from udapp perspektive, this is not really `replay`, not `re` at least. udapp just run transactions.

serapath: The button in the UI is called `re-run transactions` so i think the name should be the same.
Users will anyway not see it, so it's just for us. Is that good? :-)

src/universal-dapp.js
serapath: I kind of agree, but then... look at the code (from `run-tab.js`)
```js
// ...
/* ------------------------------------------------
    RECORDER
------------------------------------------------ */
function makeRecorder (appAPI, appEvents) {
  var udapp = appAPI.udapp()
  var recorder = new Recorder({
    events: {
      udapp: appEvents.udapp,
      executioncontext: executionContext.event
    },
    api: {
      getAccounts (cb) { udapp.getAccounts(cb) }
    }
  })

// ...

var runButton = yo`<button class=${css.runTxs}>re-run transactions</button>`

// ....

  runButton.onclick = () => {
    var opts = { title: `Enter Transactions`, text: `Paste the array of transaction you want to replay here`, inputValue: '', multiline: true }
    modalDialogCustom.promptMulti(opts, function confirm (json = '[]') {
      try {
        var txArray = JSON.parse(json)
      } catch (e) {
        modalDialogCustom.alert('Invalid JSON, please try again')
      }
      if (txArray.length) {
        txArray.forEach(tx => udapp.rerunTx(tx.record, CALLBACK))
      }
    }, function cancel () { })
  }
```

serapath: I think this somehow needs to be re-architected.

1. `run-tab.js` uses `udapp` and a `recorder`, ... which depends on `udapp` as a "passed in/injected" dependency.

2. now `run-tab.js` has of course the buttons to copy and re-run transactions and the event listeners use `udapp.rerunTx` - which currently makes sense, because we are re-using A LOT of code from `UniversalDApp.prototype.runTx` ... because `.replayTx(...)` or now `.rerunTx(...)` is very similar.

**now what you are saying is**
The `recorder` in runtab.js, which already get's `udapp events` injected...
...would need to be accessible inside of `udapp` for the `.rerunTx` method?
Or i would move that method to recorder instead, but it heavily re-uses udapp.runTx stuff...

Somehow everything is using everything else and it feels to me essentially like spagetthi code.

I could probably think about it and come up with some kind of a little bit bigger refactoring to untangle this, but maybe you have a good suggestion how to deal with this situation.

yann300: I am not saying that the recorder should be accessible inside udapp.
I am saying that `the udapp just takes a tx that is quasi filled with all the needed param (except gasLimit).`

The process of converting `<account - n>` to real addresses should not be done inside `rerunTx` but should be done just before calling `rerunTx` and should be done inside `recorder`.
Actually the process of translating addresses should perhaps be done in an another module, but keeping the all the code in the `recorder` is fine for now.

btw the name `rerunTx` is not really appropriate. Because from the point of view of `udapp`, we just *run* a transaction (well I can't right now figure out a name, but will think on that).

src/universal-dapp.js
```diff
@@ -461,103 +460,103 @@ UniversalDApp.prototype.pendingTransactions = function () {
...
-      } else {
-        self.getAccounts(function (err, ret) {
-          if (err) {
-            return callback(err)
-          }
+UniversalDApp.prototype.rerunTx = function (args, cb) {
+  var self = this
+  self.getAccounts(function (err, accounts = []) {
+    if (err) console.error(err)
+    if (args.to && args.to[0] === '<') args.to = accounts[args.to.split('>')[0].slice(11)]
```
yann300: This part should really be in `recorder.js` , and called line 315 run-tab.js

src/app/tabs/run-tab.js
```diff
@@ -167,26 +164,45 @@ var css = csjs`
...
 module.exports = runTab
 
-var instanceContainer = yo`<div class="${css.instanceContainer}"></div>`
 var noInstancesText = yo`<div class="${css.noInstancesText}">0 contract Instances</div>`
 
-var pendingTxsText = yo`<div class="${css.pendingTxsText}"></div>`
-var pendingTxsContainer = yo`<div class="${css.pendingTxsContainer}">${pendingTxsText}</div>`
-
 function runTab (container, appAPI, appEvents, opts) {
+  var self = {
```
yann300: why `self` is that does not reference `this` ?

yann300: That's confusing as `self` is generally used to reference `this`

src/app/tabs/run-tab.js
```diff
@@ -167,26 +164,45 @@ var css = csjs`
...
+    _view: {},
+    addInstance: addInstance
+  }
+  var udapp = appAPI.udapp()
+
+  var instanceContainer = yo`<div class="${css.instanceContainer}"></div>`
```
yann300: do you really need to move that bunch of code ? (from line 176 to line 183)

src/app/tabs/run-tab.js
```diff
@@ -167,26 +164,45 @@ var css = csjs`
...
+    ${contractDropdown(self, appAPI, appEvents, instanceContainer)}
     ${pendingTxsContainer}
     ${instanceContainer}
   </div>
   `
   container.appendChild(el)
 
+  function addInstance (sourcename) {
+    var contract = appAPI.getContract(sourcename)
+    var address = self._view.atAddressButtonInput.value
```
yann300: that why you get `0x...0x` in the contract instance. here you should get the created contract address...

PR #808, opened 2017-09-21
Fix analysis (fn call at contract scope)

yann300:
avoid throwing when analysis code where a function call is made at the contract scope (e.g not inside a function)


chriseth:
looks good!

PR #819, opened 2017-09-25
Dynamic highlight

yann300:


src/app.js
```diff
@@ -383,6 +384,25 @@ function run () {
...
 
+  // ---------------- ContextualListener -----------------------
+  this._components.contextualListener = new ContextualListener({
+    getCursorPosition: () => {
+      return this._components.editor.getCursorPosition()
+    },
+    getCompilationResult: () => {
+      return compiler.lastCompilationResult
+    },
+    warnExpression: (position) => {
```
chriseth: Why is it called "warn"?

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,93 @@
...
+      this._warnExpressions(api.getCursorPosition(), api.getCompilationResult())
+    }, 1000)
+  }
+
+  _warnExpressions (cursorPosition, compilationResult) {
+    if (this.currentPosition === cursorPosition) return
+    this._stopWarning()
+    this.currentPosition = cursorPosition
+    if (compilationResult && compilationResult.data && compilationResult.source) {
+      var nodes = this.sourceMappingDecoder.nodesAtPosition(null, cursorPosition, compilationResult.data.sources[compilationResult.source.target])
```
chriseth: I'm getting `this.sourceMappingDecoder.nodesAtPosition is not a function`

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,93 @@
...
+      if (nodes && nodes.length && nodes[nodes.length - 1]) {
+        this._warnExpression(nodes[nodes.length - 1], compilationResult)
```
chriseth: `nodes.length - 1` is the one with the "smallest" source location, i.e. the deepest entry in the AST?

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,93 @@
...
+    if (compilationResult && compilationResult.sources) {
+      var self = this
+      var callback = {}
+      callback['*'] = function (node) {
+        if (node && node.attributes && node.attributes.referencedDeclaration) {
+          if (!self._index['ReferencedDeclarations'][node.attributes.referencedDeclaration]) {
+            self._index['ReferencedDeclarations'][node.attributes.referencedDeclaration] = []
+          }
+          self._index['ReferencedDeclarations'][node.attributes.referencedDeclaration].push(node)
+          self._index['Expressions'].push(node)
```
chriseth: Why is it called "expressions"?

src/app/editor/contextualListener.js
yann300: yes

src/app.js
yann300: Means that the module wants to warn app.js that this position need a particular treatment. cause we might extend this function to do something else than just highlight.

src/app.js
```diff
@@ -383,6 +384,25 @@ function run () {
...
+      position = offsetToLineColumnConverter.offsetToLineColumn(position, position.file, compiler.lastCompilationResult)
+      return editor.addMarker(position, config.get('currentFile'), 'highlightcall')
```
chriseth: Why highlightcall? Should be something like "highlightCurrentDeclaration" or "...CurrentRefereces"

src/app/editor/contextualListener.js
yann300: in this context, `Expressions` would be a general term to describe use of a declared item.

src/app.js
yann300: yes right

chriseth: Be careful - references across files are easily possible.

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,93 @@
...
+        ReferencedDeclarations: {},
+        Expressions: []
+      }
+      if (success) {
+        this._buildIndex(data, source)
+      }
+    })
+    this.sourceMappingDecoder = new SourceMappingDecoder()
+    this.astWalker = new AstWalker()
+    setInterval(() => {
```
chriseth: Is this cancelled if we recompile in the meantime?

chriseth: (or change the text)?

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,93 @@
...
+      this._warnExpressions(api.getCursorPosition(), api.getCompilationResult())
+    }, 1000)
+  }
+
+  _warnExpressions (cursorPosition, compilationResult) {
+    if (this.currentPosition === cursorPosition) return
```
chriseth: Is this condition enough? We should also require that the source text is the same, think of copy-paste or changing the source file, etc.

src/app.js
yann300: Yes now only one sourceunit is managed. but that stuff need to be fixed when we'll add other feature...

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,93 @@
...
+    this._stopWarning()
```
chriseth: Hm, this should perhaps also be called directly whenever we move the cursor out of the current source location.

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,93 @@
...
+      }
+      this.astWalker.walk(compilationResult.sources[source.target].AST, callback)
+    }
+  }
+
+  _warnExpression (node, compilationResult) {
+    var self = this
+    function highlight (id) {
+      if (self._index['ReferencedDeclarations'] && self._index['ReferencedDeclarations'][id]) {
+        var calls = self._index['ReferencedDeclarations'][id]
```
chriseth: `references` instead of `calls`?

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,93 @@
...
+        for (var call in calls) {
+          self._warn(calls[call].src, compilationResult)
```
chriseth: We should not only highlight the reference but also the declaration.

yann300: the source location of the declaration is a bit tricky to hightlight, cause it's the whole block and not the function name.
though we could highlight the whole block with like 0.2 opacity

src/app.js
chriseth: I think currently, this is really confusing. It sounds like this would be used for compiler warnings.

chriseth: Isn't "highlight" actually more general than "warn"?

src/app/editor/contextualListener.js
chriseth: Why not `declaration` then?

chriseth: The two sides of the "link" should be called `declaration` and `reference`. Neither of them has to be an expression.

src/app.js
chriseth: Sure, that's fine, I just did not see the check makes sure that `position` and `config.get('currentFile')` reference the same file.

src/app/editor/contextualListener.js
chriseth: Ah, I see. Perhaps we could also search for a sub-node called "name" inside the node and highlight that?

src/app/editor/contextualListener.js
yann300: no but this is not really linked to the compilation event, the function in setInterval will be executed every 1s and check: if the position of the cursor changed and if there's a compilation result to work with.

src/app/editor/contextualListener.js
yann300: ok so we could just unwarn when content changed

src/app/editor/contextualListener.js
yann300: this is the case

src/app/editor/contextualListener.js
yann300: yes

src/app/editor/contextualListener.js
chriseth: ok!

src/app.js
```diff
@@ -383,6 +384,43 @@ function run () {
...
+  // ---------------- ContextualListener -----------------------
+  this._components.contextualListener = new ContextualListener({
+    getCursorPosition: () => {
+      return this._components.editor.getCursorPosition()
+    },
+    getCompilationResult: () => {
+      return compiler.lastCompilationResult
+    },
+    highlight: (position, node) => {
+      if (compiler.lastCompilationResult && compiler.lastCompilationResult.source && compiler.lastCompilationResult.source.target === config.get('currentFile')) {
```
chriseth: Please replace `compiler.lastCompilationResult.source.target === config.get('currentFile')` by a check that `position.file` is the file currently shown in the editor.

src/app.js
```diff
@@ -383,6 +384,43 @@ function run () {
...
+        position = offsetToLineColumnConverter.offsetToLineColumn(position, position.file, compiler.lastCompilationResult)
+        var css = 'highlightreference'
+        if (node.children && node.children.length) {
```
chriseth: Please add a comment about what this block does.

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,102 @@
...
+var AstWalker = require('ethereum-remix').util.AstWalker
+
+class ContextualListener {
+  constructor (api, events) {
+    this._api = api
+    this._index = {
+      Declarations: {},
+      FlatReferences: {}
+    }
+    this._events = []
```
chriseth: Please rename to something like `activeHighlights`

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  _highlightItems (cursorPosition, compilationResult) {
+    if (this.currentPosition === cursorPosition) return
+    this._stopHighlighting()
+    this.currentPosition = cursorPosition
+    if (compilationResult && compilationResult.data && compilationResult.source) {
+      var nodes = this.sourceMappingDecoder.nodesAtPosition(null, cursorPosition, compilationResult.data.sources[compilationResult.source.target])
+      if (nodes && nodes.length && nodes[nodes.length - 1]) {
+        this._hightlightExpressions(nodes[nodes.length - 1], compilationResult)
```
chriseth: Typo: high`t`light

src/app/editor/contextualListener.js
```diff
@@ -0,0 +1,102 @@
...
+    if (eventId) {
+      this._activeHighlights.push({ eventId, position, fileTarget: compilationResult.source.target })
+    }
+  }
+
+  _hightlightExpressions (node, compilationResult) {
+    var self = this
+    function highlights (id) {
+      if (self._index['Declarations'] && self._index['Declarations'][id]) {
+        var calls = self._index['Declarations'][id]
```
chriseth: Please rename `calls`

PR #833, opened 2017-09-27
Don't run blockgaslimit update requests on provider set to vm

holgerd77:
This prevents constant request error output on the console when using ``vm`` as a provider.

Tested both the new negative ``vm`` case as well as the positive (still requesting) ``injected`` case with MetaMask.


yann300:
thanks!

PR #837, opened 2017-09-29
Refactor/fix execution context change

holgerd77:
This should avoid a lot of confusion when switching the execution context. PR fixes different switch cases and gives feedback in error cases (no provider connection), the commit messages give a detailed overview on the changes.

Every switch touched has been tested for both the positive and the negative case in a live environment (MetaMask on/off, local Geth).

### DIFF
--- a/src/execution-context.js
+++ b/src/execution-context.js
@@ -128,24 +128,28 @@ function ExecutionContext () {
       })
     }
 
+    if (context === 'vm') {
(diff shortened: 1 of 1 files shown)


holgerd77:
Added an extra note on page reload to the injected-provider-not-found msg to address the latest support requests on the remix gitter channel.

PR #838, opened 2017-09-29
Fixes false file-exists error message on file renaming

holgerd77:
This fixes the file renaming issue mentioned in https://github.com/ethereum/browser-solidity/issues/829 and caused by double-triggering of events through ``blur`` and ``enter`` ``keyEvent``.

I've taken a more conservative fix path by adding a semaphore variable providing access and not removing one of the event types, since I wasn't sure how different browsers would react.

Tested:

- Renaming/Accept and Reload
- Renaming/Cancel and Reload
- Renaming to same file name
- Different browsers

### DIFF
--- a/src/app/files/file-explorer.js
+++ b/src/app/files/file-explorer.js
@@ -128,6 +128,7 @@ function fileExplorer (appAPI, files) {
   var filepath = null
   var focusElement = null
   var textUnderEdit = null
+  var textInRename = false
 
   var events = new EventManager()
   this.events = events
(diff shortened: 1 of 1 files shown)


yann300:
thanks for that!
Are you on safari?

holgerd77:
No, I've tested Chrome and Firefox. Is Safari not working? (change shouldn't really touch browser specific behavior though)

yann300:
no, cause i never had this bug before (in Chrome), but I know that it happens all the time in safari.
@ninabreznik could you check if that fix the issue in safari ?

PR #848, opened 2017-10-04
Remove needless enclosing div tag

kurotaky:
Hi! I deleted the div tag because I found it in the following error.
`multiple root elements must be wrapped in an enclosing tag`
TravisCI also fell with the same error 😢 
https://travis-ci.org/ethereum/browser-solidity/builds/282985894

The following is my environments :)
### Error message
```
Error: /Users/kurotaky/ghq/github.com/ethereum/browser-solidity/src/app/panels/terminal.js: multiple root elements must be wrapped in an enclosing tag
```

### OS
```
ProductName:	Mac OS X
ProductVersion:	10.11.6
BuildVersion:	15G1611
```

### node --version
```
v8.1.3
```

### npm version
```
4.6.1
```


yann300:
thanks! :)

PR #853, opened 2017-10-05
Warn comp fail

ryestew:
redesign and refactor of css for the compiler alert error on the run tab

### DIFF
--- a/src/app/tabs/run-tab.js
+++ b/src/app/tabs/run-tab.js
@@ -250,7 +250,11 @@ function contractDropdown (appAPI, appEvents, instanceContainer) {
   instanceContainer.appendChild(noInstancesText)
 
   appEvents.compiler.register('compilationFinished', function (success, data, source) {
-    getContractNames(success, data)
+    if (success) {
+      getContractNames(success, data)
+    } else {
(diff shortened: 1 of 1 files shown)


src/app/tabs/run-tab.js
```diff
@@ -263,13 +270,17 @@ function updateAccountBalances (container, appAPI) {
 
 function contractDropdown (appAPI, appEvents, instanceContainer) {
   instanceContainer.appendChild(noInstancesText)
-  var compFails = yo`<i title="Contract compilation failed. Please check the compile tab for more information." class="fa fa-thumbs-down ${css.errorIcon}" ></i>`
+  var compFails = yo`<span title='Contract compilation failed. Please check the compile tab for more information' class="${css.errorSymb}"></span>`
```
yann300: we should perhaps display the same icon in the compile-tab ..

yann300:
@ryestew @ninabreznik I pushed one change: added a red border and title in compile tab.
is this compliant with the style guide and ok to be merged?
There are still inline css that should go in a class, but  i m not sure if that should go to style-guide.js . could you take care of that?

PR #854, opened 2017-10-05
Use standard jsonio

yann300:


src/app/compiler/compiler-input.js
```diff
@@ -0,0 +1,25 @@
...
+
+/*
+  opts:
+   - optimize
+   - { file_name: { library_name: address } }
+*/
+module.exports = (sources, opts) => {
+  return JSON.stringify({
+    language: 'Solidity',
+    sources,
```
axic: I think this should be `sources: sources` ?

yann300: `sources, ` like this also work

axic: What does it do? What is `sources`? Is it `{ sources: {} }`?

src/app/compiler/compiler-input.js
```diff
@@ -0,0 +1,25 @@
...
+    settings: {
+      optimizer: {
+        enabled: opts.optimize ? opts.optimize : false,
```
axic: This seems like a not very useful construct, if you don't trust `optimize` to be a boolean you could do `enabled: opts.optimize === true`.

src/app/compiler/compiler-worker.js
```diff
@@ -22,10 +23,11 @@ module.exports = function (self) {
 
         compileJSON = function (input, optimize) {
           try {
-            return JSON.stringify(compiler.compile(JSON.parse(input), optimize, function (path) {
+            var inputStandard = compilerInput(JSON.parse(input), {optimize})
```
axic: Same question with `optimize` here as with the sources.

src/app/compiler/compiler.js
```diff
@@ -261,15 +324,15 @@ function Compiler (handleImportCall) {
...
-          files[m] = content
+          files[m] = { content }
           gatherImports(files, target, importHints, cb)
         }
       })
 
       return
     }
 
-    cb(null, { 'sources': files, 'target': target })
```
axic: I think `target` was needed in the compiler part and I don't see it removed above?

src/app/tabs/compile-tab.js
```diff
@@ -305,17 +305,17 @@ function compileTab (container, appAPI, appEvents, opts) {
       }
       if (!error) {
         if (data.contracts) {
-          for (var contract in data.contracts) {
-            appAPI.compilationMessage(contract, $(errorContainer), {type: 'success'})
-          }
+          appAPI.visitContracts((contrract) => {
```
axic: Typo: `contrract`.

src/app/tabs/compile-tab.js
```diff
@@ -305,17 +305,17 @@ function compileTab (container, appAPI, appEvents, opts) {
...
+            appAPI.compilationMessage({ formattedMessage: contrract.name, severity: 'success' }, $(errorContainer))
```
axic: There is no `severity: success`. What is this code doing?

src/app/tabs/compile-tab.js
```diff
@@ -365,11 +365,9 @@ function compileTab (container, appAPI, appEvents, opts) {
         keys.map(x => {
           var copyDetails = yo`<span class="${css.copyDetails}"><i title="Copy value to clipboard" class="fa fa-clipboard" onclick=${() => { copy(details[x]) }} aria-hidden="true"></i></span>`
           var questionMark = yo`<span class="${css.questionMark}"><i title="${detailsHelpSection()[x]}" class="fa fa-question-circle" aria-hidden="true"></i></span>`
-          var keyDisplayName
-          (x === 'interface') ? keyDisplayName = 'interface - abi' : keyDisplayName = x
           log.appendChild(yo`
             <div class=${css.log}>
-              <div class="${css.key}">${keyDisplayName} ${copyDetails} ${questionMark}</div>
+              <div class="${css.key}">${x} ${copyDetails} ${questionMark}</div>
```
axic: Hm, perhaps worth considering renaming `x` here.

src/app/tabs/compile-tab.js
```diff
@@ -381,11 +379,11 @@ function compileTab (container, appAPI, appEvents, opts) {
     function insertValue (details, x) {
       var value = yo`<pre class="${css.value}"></pre>`
       var node
-      if (x === 'bytecode' || x === 'metadataHash' || x === 'swarmLocation' || x === 'Runtime Bytecode' || x === 'Opcodes') {
+      if (x === 'bytkecode' || x === 'Runtjime Bytecode' || x === 'Opcoldes') {
```
axic: `bytkecode`, `runtjime` `opcoldes` seem to have typos

src/app/compiler/compiler-input.js
yann300: sources is a map of sources.
that's just a shorthand for `sources: sources,`

axic: I think the problem with this construct is that not many are familiar with it (at least I aren't) and that if someone goes to refactor the variable name, it will break the JSON.

src/app/tabs/compile-tab.js
yann300: ah yes I meant `type` not `severity`

src/app/tabs/compile-tab.js
```diff
@@ -381,11 +379,9 @@ function compileTab (container, appAPI, appEvents, opts) {
     function insertValue (details, x) {
```
axic: This variable name `x` is not really telling much about its intention.

axic:
Is it possible to split this up into smaller pull requests? One would be cleaning those variable names like `x` as it is extremely hard to review this pull request in a reliable manner right now.

PR #856, opened 2017-10-05
Check special char

yann300:


src/app.js
```diff
@@ -369,7 +369,12 @@ function run () {
   // Add files received from remote instance (i.e. another browser-solidity)
   function loadFiles (filesSet) {
     for (var f in filesSet) {
-      filesProviders['browser'].set(helper.createNonClashingName(f, filesProviders['browser']), filesSet[f].content)
+      var name = helper.createNonClashingName(f, filesProviders['browser'])
+      if (filesProviders['browser'].checkSpecialChars(name)) {
```
chriseth: While this is a good counter-measure, I don't think this is a proper fix. The code using yoyo has to do proper escaping.

src/app/files/shared-folder.js
```diff
@@ -45,6 +45,10 @@ class SharedFolder {
     cb()
   }
 
+  checkSpecialChars (name) {
+    return name.match(/(\/|:|\*|\?|"|<|>|\\|\||')/) != null
```
chriseth: Can you create this function only once?

chriseth: Ah and something like `/[/:*?"<>\\\|]/` would be better I think.

src/app/tabs/settings-tab.js
```diff
@@ -140,6 +140,10 @@ function loadVersion (version, queryParams, appAPI, el) {
 
     url = location + 'soljson.js'
   } else {
+    if (version.indexOf('soljson') !== 0 || version.match(/(\/|:|\*|\?|"|<|>|\\|\||')/) != null) {
```
chriseth: A third version of the regex :-)

src/app/files/shared-folder.js
yann300: one is for browser-storage and the other for shared-folder (remixd). there might have a different requirement for `checkSpecialChars` ... but I guess that is fine to keep the same behavior for both

src/app/tabs/settings-tab.js
```diff
@@ -140,6 +141,10 @@ function loadVersion (version, queryParams, appAPI, el) {
 
     url = location + 'soljson.js'
   } else {
+    if (version.indexOf('soljson') !== 0 || helper.checkSpecialChars(version) != null) {
```
chriseth: why do you compare to `null`?

yann300: yes I guess `if (version.indexOf('soljson') !== 0 || helper.checkSpecialChars(version))` would have been sufficient...

PR #861, opened 2017-10-11
Use standard jsonio - (2)

yann300:


src/app/compiler/compiler-input.js
```diff
@@ -0,0 +1,25 @@
...
+    settings: {
+      optimizer: {
+        enabled: opts.optimize === true,
+        runs: 500
+      }
+    },
+    libraries: opts.libraries,
+    outputSelection: {
+      '*': {
+        '*': [ 'metadata', 'evm.bytecode', 'abi', 'legacyAST', 'metadata', 'evm.assembly', 'evm.methodIdentifiers', 'evm.gasEstimates' ]
```
axic: Some code below refers to `evm.deployedBytecode`, that must be included here.

src/app/compiler/compiler.js
```diff
@@ -281,12 +348,9 @@ function Compiler (handleImportCall) {
...
 
   function updateInterface (data) {
-    for (var contract in data.contracts) {
-      var abi = JSON.parse(data.contracts[contract].interface)
-      abi = solcABI.update(truncateVersion(currentVersion), abi)
-      data.contracts[contract].interface = JSON.stringify(abi)
-    }
-
+    txHelper.visitContracts(data.contracts, (contract) => {
+      data.contracts[contract.file][contract.name].abi = solcABI.update(truncateVersion(currentVersion), contract.object.abi)
```
axic: What is `contract.object`? If that refers to `evm.bytecode.object` then this is wrong since `.abi` is not part of that,.

yann300: `contract` ìs something like:
```
{
file,
name,
object
}
```
`object` does not refers to `evm.bytecode.object`

yann300: `object` refers to `contracts[fileName][contractName]`

src/app.js
```diff
@@ -315,12 +315,21 @@ function run () {
             }
           }
         }
-        return editor.addMarker(lineColumn, compiler.lastCompilationResult.data.sourceList[position.file], css)
+        var fileName = compiler.getSourceName(position.file)
```
axic: This seems like a good refactoring which could be split off.

src/app.js
```diff
@@ -595,7 +601,7 @@ function run () {
       this.fullLineMarker = null
       this.source = null
       if (lineColumnPos) {
-        this.source = compiler.lastCompilationResult.data.sourceList[location.file] // auto switch to that tab
+        this.source = compiler.getSourceName(location.file)
```
axic: Seems like a refactor too.

src/app.js
```diff
@@ -535,12 +547,6 @@ function run () {
     compilationMessage: (message, container, options) => {
       renderer.error(message, container, options)
     },
-    currentCompiledSourceCode: () => {
```
axic: If this is still needed it should still work.

src/app/compiler/compiler.js
```diff
@@ -219,10 +282,14 @@ function Compiler (handleImportCall) {
     worker.postMessage({cmd: 'loadVersion', data: url})
   }
 
-  function gatherImports (files, target, importHints, cb) {
```
axic: I remember now, `target` was used for this import loop to avoid cycles or something like that.

src/app/compiler/compiler.js
```diff
@@ -219,10 +282,14 @@ function Compiler (handleImportCall) {
...
+  function gatherImports (files, importHints, cb) {
     importHints = importHints || []
+    var fileList = Object.keys(files)
+    if (!fileList.length) {
+      return cb(null, {})
+    }
     if (!compilerAcceptsMultipleFiles) {
-      cb(null, files[target])
+      cb(null, files[fileList[0]])
```
axic: I am not sure `fileList[0]` will work in every situation? Have you tried multiple imports and import chains? With and without failing imports?

src/app.js
yann300: yes sure but that line needs to be changed anyway cause `sourceList` is not here anymore. so better using `getSourceName` already

src/app/contract/contractParser.js
```diff
@@ -1,49 +1,40 @@
...
-    detail.web3Deploy = gethDeploy(contractName.toLowerCase(), contract['interface'], contract.bytecode)
+  if (contract.evm.bytecode.object) {
+    detail.bytecode = contract.evm.bytecode
+    detail.web3Deploy = gethDeploy(contractName.toLowerCase(), contract.abi, contract.evm.bytecode.object)
 
-    detail.metadataHash = retrieveMetadataHash(contract.bytecode)
+    detail.metadataHash = retrieveMetadataHash(contract.evm.bytecode.object)
     if (detail.metadataHash) {
-      detail.swarmLocation = 'bzz://' + detail.metadataHash
+      detail.swarmLocation = 'bzzr://' + detail.metadataHash
```
axic: This seems like a bug fix not related to this PR.

src/app/contract/contractParser.js
```diff
@@ -1,49 +1,40 @@
...
-  if (contract.runtimeBytecode && contract.runtimeBytecode.length > 0) {
-    detail['Runtime Bytecode'] = contract.runtimeBytecode
-  }
-
-  if (contract.opcodes !== undefined && contract.opcodes !== '') {
-    detail['Opcodes'] = contract.opcodes
-  }
-
-  if (contract.assembly !== null) {
-    detail['Assembly'] = formatAssemblyText(contract.assembly, '', source)
```
axic: Actually the old assembly still exists, why not keep it?

src/app/contract/contractParser.js
```diff
@@ -1,49 +1,40 @@
...
 
-  detail.gasEstimates = formatGasEstimates(contract.gasEstimates)
+  detail.gasEstimates = formatGasEstimates(contract.evm.gasEstimates)
 
-  if (contract.runtimeBytecode && contract.runtimeBytecode.length > 0) {
-    detail['Runtime Bytecode'] = contract.runtimeBytecode
-  }
-
-  if (contract.opcodes !== undefined && contract.opcodes !== '') {
-    detail['Opcodes'] = contract.opcodes
```
axic: This also still exists.

src/app/compiler/compiler.js
yann300: oh ok, does that needs to be added to jsonio as well?

src/app/contract/contractParser.js
```diff
@@ -128,7 +88,7 @@ var formatGasEstimates = function (data) {
   var ret = {}
   var fun
   if ('creation' in data) {
-    ret['Creation'] = gasToText(data.creation[0]) + ' + ' + gasToText(data.creation[1]) + '\n'
```
axic: Hm wondering why this even had a new line?

src/app/contract/contractParser.js
```diff
@@ -128,7 +88,7 @@ var formatGasEstimates = function (data) {
...
+    ret['Creation'] = data.creation
```
axic: The new creation object is actually more complex than being a single number:
```
            creation: {
              codeDepositCost: "420000",
              executionCost: "infinite",
              totalCost: "infinite"
            },
```

Do you account for that in code outside of this function?

src/app/contract/contractParser.js
yann300: yes and that will be displayed , `opcode` is included in the `contract.evm.bytecode`

src/app/compiler/compiler.js
axic: It is not related to json io but to the import process browser-solidity uses.

src/app/compiler/compiler.js
yann300: That is the same behavior as before:  we just take one file, and as we don't have `target` anymore we take the first one.
I plan to add more test in the coming days, testing import more carefully is planned

src/app/contract/contractParser.js
yann300: the other place where gas cost is used is for static analysis and yes `infinite` is managed.

src/app.js
yann300: this was used at some point for `target`

src/app/compiler/compiler.js
yann300: hmm ok I don't really see how this is used yet, I will take a look.
Just wondering cause `target` was part of the input of the `compile()`

src/app/contract/contractParser.js
axic: I mean before this `ret['Creation']` was a string and now it is an object.

src/app/compiler/compiler.js
axic: I'm not sure if `target` always equals to the first file.

src/app/compiler/compiler.js
axic: But the actual solc interface is `compileInternal` and it was not passed to that one.

src/app/compiler/compiler.js
yann300: well actually I don't really get where `target` is used...
except here: https://github.com/ethereum/browser-solidity/blob/master/src/app/compiler/compiler.js#L239
but this seems wrong to me...
rather `var path = /(.*\/).*/.exec(fileName)` than `var path = /(.*\/).*/.exec(target)`

src/app/compiler/compiler.js
yann300: `target` is passed to `compileJSONCallback`

src/app/compiler/compiler.js
axic: I am not sure, it is one of those magic codes which worked at some point and if I recall correctly it was to avoid issues with imports (perhaps loops - where A imports B and B imports A). solc handles this properly, but remix does its own import feature before solc.

src/app/contract/contractParser.js
yann300: oh ok i readd it

src/app/contract/contractParser.js
yann300: I think before that block was displayed in a big <pre>

src/app/contract/contractParser.js
yann300: I will readd it

src/app/contract/contractParser.js
yann300: yes that is ok

src/app/compiler/compiler.js
```diff
@@ -295,10 +291,6 @@ function Compiler (handleImportCall) {
 
   function gatherImports (files, target, importHints, cb) {
     importHints = importHints || []
-    if (!compilerAcceptsMultipleFiles) {
```
chriseth: This is handled by the standard wrapper.

chriseth:
Simplified communication with worker a little.

PR #863, opened 2017-10-11
upgrade ethereumjs-vm version to byzantium

jwasinger:
Updates ethereumjs-vm to dependency to 2.3.1 in order to include Byzantium support.


axic:
Was added in e0208f11.

PR #865, opened 2017-10-12
Fix critical bug using Ether as base unit for value input / add some context help

holgerd77:
This fixes a critical bug using ``Ether`` as a base unit for the value input field in the ``Run`` tab, which could in combination without warning / extra hint lead to people loosing thousands of dollars.

Beside from that there is some extra context help texts added to ``Run`` tab input elements.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -215,7 +215,13 @@ function run () {
     getValue: (cb) => {
       try {
         var comp = $('#value').val().split(' ')
-        cb(null, executionContext.web3().toWei(comp[0], comp.slice(1).join(' ')))
+        if (comp.length === 1) {
+          cb(null, comp[0])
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -215,7 +215,13 @@ function run () {
     getValue: (cb) => {
       try {
         var comp = $('#value').val().split(' ')
-        cb(null, executionContext.web3().toWei(comp[0], comp.slice(1).join(' ')))
+        if (comp.length === 1) {
+          cb(null, comp[0])
+        } else if (comp.length === 2) {
+          cb(null, executionContext.web3().toWei(comp[0], comp.slice(1).join(' ')))
```
yann300: that can still set an empty unit to `web3` without throwing an error

PR #868, opened 2017-10-13
Refactor run tab (value, create and atAddress buttons, instance title)

ninabreznik:
screen shot 2017-10-13 at 2 38 49 am

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -217,9 +217,17 @@ function run () {
         var number = document.querySelector('#value').value
         var select = document.getElementById('unit')
         var index = select.selectedIndex
-        var selectedUnit = select.querySelectorAll('option')[index].innerHTML
-        var unit = ''
-        if (selectedUnit === 'wei') { unit = 'wei' }
+        var selectedUnit = select.querySelectorAll('option')[index].dataset.unit
+        var unit = 'ether' // default
+        if (selectedUnit === 'ether') {
+          unit = 'ether'
+        } else if (selectedUnit === 'microether') {
+          unit = 'microether'
+        } else if (selectedUnit === 'mwei') {
+          unit = 'mwei'
+        } else if (selectedUnit === 'wei') {
+          unit = 'wei'
+        }
         cb(null, executionContext.web3().toWei(number, unit))
       } catch (e) {
         cb(e)
--- a/src/app.js
+++ b/src/app.js
@@ -221,10 +221,10 @@ function run () {
         var unit = 'ether' // default
         if (selectedUnit === 'ether') {
           unit = 'ether'
-        } else if (selectedUnit === 'microether') {
-          unit = 'microether'
-        } else if (selectedUnit === 'mwei') {
-          unit = 'mwei'
+        } else if (selectedUnit === 'finney') {
+          unit = 'finney'
+        } else if (selectedUnit === 'gwei') {
+          unit = 'gwei'
         } else if (selectedUnit === 'wei') {
           unit = 'wei'
         }
--- a/src/app.js
+++ b/src/app.js
@@ -214,8 +214,13 @@ function run () {
     },
     getValue: (cb) => {
       try {
-        var comp = $('#value').val().split(' ')
-        cb(null, executionContext.web3().toWei(comp[0], comp.slice(1).join(' ')))
+        var number = document.querySelector('#value').value
+        var select = document.getElementById('unit')
+        var index = select.selectedIndex
+        var selectedUnit = select.querySelectorAll('option')[index].innerHTML
+        var unit = ''
+        if (selectedUnit === 'wei') { unit = 'wei' }
+        cb(null, executionContext.web3().toWei(number, unit))
       } catch (e) {
         cb(e)
       }
--- a/src/app/execution/txLogger.js
+++ b/src/app/execution/txLogger.js
@@ -21,7 +21,7 @@ var css = csjs`
     justify-content: space-between;
   }
   .txLog {
-    width: 75;
+    width: 75%;
   }
   .tx {
     color: ${styles.terminal.text_Title_TransactionLog};
--- a/src/app/execution/txLogger.js
+++ b/src/app/execution/txLogger.js
@@ -20,6 +20,9 @@ var css = csjs`
     align-items: end;
     justify-content: space-between;
   }
+  .txLog {
+    width: 75;
+  }
   .tx {
     color: ${styles.terminal.text_Title_TransactionLog};
     font-weight: bold;
@@ -45,19 +48,16 @@ var css = csjs`
   .buttons {
     display: flex;
   }
-  .debug {
-    ${styles.terminal.button_Log_Debug}
-  }
-  .details {
-    ${styles.terminal.button_Log_Details}
-  }
   .debug, .details {
+    color: ${styles.terminal.link_Debug};
+    text-decoration: underline;
+    font-weight: bold;
     min-height: 18px;
     max-height: 18px;
-    width: 45px;
-    min-width: 45px;
-    font-size: 10px;
+    width: 55px;
+    min-width: 55px;
     margin-left: 5px;
+    cursor: pointer;
   }
   .clipboardCopy {
     margin-right: 0.5em;
@@ -157,8 +157,8 @@ function renderKnownTransaction (self, data) {
       <div class="${css.log}">
         ${context(self, {from, to, data})}
         <div class=${css.buttons}>
-        <button class=${css.details} onclick=${txDetails}>Details</button>
-        <button class=${css.debug} onclick=${debug}>Debug</button>
+        <div class=${css.details} onclick=${txDetails}>Details</div>
+        <div class=${css.debug} onclick=${debug}>Debug</div>
         </div>
       </div>
     </span>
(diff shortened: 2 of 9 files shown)


holgerd77:
Hi @ninabreznik, such great work, thank you, this is so much clearer now and I love, that you guys implementing the extra unit selection. I will then close my PR on this https://github.com/ethereum/browser-solidity/pull/865/files since this is overlapping, I added some context help texts over there as well, feel free to add them to this PR if you find them appropriate.

I have got one comment on the security of the unit selection JS code and some other minor notes, will add this as an extra review.

src/app.js
```diff
@@ -214,8 +214,13 @@ function run () {
...
     getValue: (cb) => {
       try {
-        var comp = $('#value').val().split(' ')
-        cb(null, executionContext.web3().toWei(comp[0], comp.slice(1).join(' ')))
+        var number = document.querySelector('#value').value
+        var select = document.getElementById('unit')
+        var index = select.selectedIndex
+        var selectedUnit = select.querySelectorAll('option')[index].innerHTML
+        var unit = ''
+        if (selectedUnit === 'wei') { unit = 'wei' }
```
holgerd77: Sorry to be so picky here, but since this is so delicate: can you please change the unit selection to explicitly ask for ``ether``, and switch to ``wei`` for all other occasions?

```
if (selectedUnit === 'ether') {
  unit = 'ether'
} else {
  unit = 'wei'
}
```

Otherwise if something goes wrong with the ``#value``  selection (let's say e.g. someone accidentally introduces another ``#value`` field in the future or JS for whatever reasons can't read it from the ``select`` field) this falls again back to ``ether``.

src/app/tabs/run-tab.js
```diff
@@ -413,7 +423,11 @@ function settings (appAPI, appEvents) {
       </div>
       <div class="${css.crow}">
       <div class="${css.col1_1}">Value</div>
-        <input type="text" class="${css.col2}" id="value" value="0" title="(e.g. .7 ether ...)">
+        <input type="text" class="${css.col2_1}" id="value" value="0" title="Default the value is set to be in ether, if you want the value to be in wei, type i.e. 7 wei.">
```
holgerd77: This context help is not really fitting to the current implementation with the ``unit`` selector.

src/app.js
holgerd77: I would also suggest to read the choice from the ``id`` field with ``querySelectorAll('option')[0].id`` instead of using ``innerHTML``.

Then this is more stable regards to renaming, if you e.g. decide to change ``wei`` to ``Wei``.

holgerd77: And one last thing (then I'll stop, I promise :-): just thought about it a bit, it would be probably nice (since this is closer to real world number inputs) to also use one unit from the middle, here is an overview: http://ethdocs.org/en/latest/ether.html (when you see the number of zeros you generally see why this is so dangerous stuff... :-))

I think also taking ``Gwei`` would make sense, this would be ``unit = 'gwei' in addition.

src/app/tabs/run-tab.js
```diff
@@ -264,12 +274,12 @@ function contractDropdown (appAPI, appEvents, instanceContainer) {
       </div>
       <div class="${css.buttons}">
         <div class="${css.button}">
-          <div class="${css.atAddress}" onclick=${function () { loadFromAddress(appAPI) }}>At Address</div>
-          ${atAddressButtonInput}
+        ${createButtonInput}
```
yann300: why having changed the order?

src/app/tabs/run-tab.js
```diff
@@ -264,12 +274,12 @@ function contractDropdown (appAPI, appEvents, instanceContainer) {
...
+        <div class="${css.create}" onclick=${function () { createInstance() }} >Create</div>
         </div>
         <div class="${css.button}">
-          <div class="${css.create}" onclick=${function () { createInstance() }} >Create</div>
-          ${createButtonInput}
+          ${atAddressButtonInput}
+          <div class="${css.atAddress}" onclick=${function () { loadFromAddress(appAPI) }}>Load</div>
```
yann300: I m not sure `Load` is the right word for that... that can be confusing ... perhaps `Watch` would be better

src/app/tabs/run-tab.js
yann300: if there's no special reason, could you keeping the same order as before?

src/app/tabs/run-tab.js
```diff
@@ -413,7 +423,11 @@ function settings (appAPI, appEvents) {
       </div>
       <div class="${css.crow}">
       <div class="${css.col1_1}">Value</div>
-        <input type="text" class="${css.col2}" id="value" value="0" title="(e.g. .7 ether ...)">
+        <input type="text" class="${css.col2_1}" id="value" value="0" title="Default the value is set to be in ether, if you want the value to be in wei, type i.e. 7 wei.">
+        <select name="unit" class="${css.col2_2}" id="unit">
+          <option id="unit-eth">eth</option>
```
yann300: this should be `ether` rather that `eth`

src/app.js
yann300: Well I don't really get why you need an extra variable here... could this be only one `unit` variable?

yann300: +1 for Gwei . Here is a list of units: https://etherconverter.online . I guess ether, wei, gwei is sufficient

holgerd77: The one is the value read from the UI element, the other the input passed to the library, I would recommend to keep this separate.

holgerd77: (then it is more explicit/readable what is actually passed, and you are more flexible on naming/renaming things in the UI)

ninabreznik: Thanks you both for the review. Makes sense, if-else, not using inerHTML and also this with other units. Let me do the fixes.

@yann300 For the new unit, I will make a new PR.

src/app/tabs/run-tab.js
ninabreznik: I was thinking that Create instance is more common action, so I put in on top. But right now I am thinking we need to improve this whole thing again. I think we should also have a dropdown here. So `create` by default, but if you want, you can also `load`. 

Because it's either-or, so dropdown makes more sense. Otherwise it's a confusion like @holgerd77 already stated yesterday that users don't know if they need to press both buttons or what needs to be done here...

I will make necessary changes and then you can review and see if it works. I think the UI logic is way better this way.

src/app.js
holgerd77: bildschirmfoto 2017-10-13 um 15 46 14

When I'm minimizing the created contract boxes, I am not able to expand them again (see screenshot, MacOS, Chrome).

holgerd77: Generally I think it is really important that these boxes are at least discardable/removable. This is otherwise really annoying, after 4-5 creations one has to reload the page to get rid of them.

src/app.js
```diff
@@ -217,9 +217,17 @@ function run () {
...
         var index = select.selectedIndex
-        var selectedUnit = select.querySelectorAll('option')[index].innerHTML
-        var unit = ''
-        if (selectedUnit === 'wei') { unit = 'wei' }
+        var selectedUnit = select.querySelectorAll('option')[index].dataset.unit
+        var unit = 'ether' // default
+        if (selectedUnit === 'ether') {
+          unit = 'ether'
+        } else if (selectedUnit === 'microether') {
+          unit = 'microether'
```
axic: Is anyone using this? I'd argue not the only ones used are `ether`, `gwei`, `wei` and at some point in the past `finney` was also used.

ninabreznik:
screen shot 2017-10-13 at 3 26 02 pm
screen shot 2017-10-13 at 3 25 51 pm

src/app.js
yann300: +1 finney

ninabreznik: So this is what users are really operating with? I was coming from the side of a new user, something that I can wrap my head around.

But tell me what are units you guys rotate in your heads?

ninabreznik: I will go with wei, gwei, finney and ethere for now then!

axic: Actually, should we add `szabo` (nobody uses it) only because Solidity supports it? Downside: if it is added people might feel encourage.

src/app/execution/txLogger.js
```diff
@@ -20,6 +20,9 @@ var css = csjs`
     align-items: end;
     justify-content: space-between;
   }
+  .txLog {
+    width: 75;
```
yann300: just `75` ?

yann300:
It is not possible anymore to collapse an instance cause the entire block will hide...

src/app/tabs/run-tab.js
ninabreznik: I will switch it back to atAddress for now and since there's a plan to create this dropdown where user can choose whether they want to create or 'load/watch/see existing...' we can then rethink this again.

src/app/execution/txLogger.js
ninabreznik: Oops, % got lost.

yann300:
When an instance is collapsed, the close button hides ... I would see perhaps less height for the contract title...

src/app/tabs/run-tab.js
```diff
@@ -280,7 +280,7 @@ function contractDropdown (appAPI, appEvents, instanceContainer) {
         </div>
         <div class="${css.button}">
           ${atAddressButtonInput}
-          <div class="${css.atAddress}" onclick=${function () { loadFromAddress(appAPI) }}>Load</div>
+          <div class="${css.atAddress}" onclick=${function () { loadFromAddress(appAPI) }}>atAddress</div>
```
yann300: that should be `At Address` no?

ninabreznik:
@yann300 There is a fix of instance UI in the next PR (I moved CLOSE outside). Maybe then this is already fixed and we don't need to fix is here.

PR #876, opened 2017-10-19
package.json: jQuery version updated

stellargo:
jQuery version is updated to 3.2.1

Fixes https://github.com/ethereum/browser-solidity/issues/286


yann300:
that is finally merged , thanks for the PR

PR #881, opened 2017-10-23
Fix spe char

yann300:
- rework on error/warning renderer
 - switch staticanalysis warning to be a dom node

### DIFF
--- a/src/app/files/fileManager.js
+++ b/src/app/files/fileManager.js
@@ -2,6 +2,7 @@
 
 var $ = require('jquery')
 var remix = require('ethereum-remix')
+var yo = require('yo-yo')
 var EventManager = remix.lib.EventManager
 
 /*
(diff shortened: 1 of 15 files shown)


src/app/files/fileManager.js
```diff
@@ -98,7 +99,7 @@ class FileManager {
     $filesEl.find('.file').remove()
 
     for (var file in this.tabbedFiles) {
-      $filesEl.append($('<li class="file"><span class="name">' + file + '</span><span class="remove"><i class="fa fa-close"></i></span></li>'))
```
chriseth: Oh wow, so I guess this is place of the bug, isn't it?

src/app/ui/renderer.js
```diff
@@ -76,48 +76,72 @@ function Renderer (appAPI) {
...
+  if (!opt.noAnnotations && errLocation) {
+    this.appAPI.error(errLocation.errFile, {
+      row: errLocation.errLine,
+      column: errLocation.errCol,
+      text: text,
+      type: opt.type
     })
   }
 
+  var $pre = $(opt.useSpan ? '<span />' : '<pre />').html(message)
```
chriseth: Is there no way to do that with yoyo?

src/app/files/fileManager.js
yann300: yes it is

src/app/ui/renderer.js
yann300: yes but didn't want to change everything, and this part is not going to be harmful
I can use yoyo here though

PR #890, opened 2017-10-31
Update README.md

riyagarg47:


axic:
I think the README headings are correct currently. Please reopen if you reason otherwise.

PR #899, opened 2017-11-18
Default to JavaScript VM even with injected Web3

federicobond:
I have MetaMask installed in Chrome so I always have to first change the Environment option to use the features in the Run tab. Since the biggest use case for Remix right now is to quickly try out some code, especially for beginners, I find the current behavior annoying.

I discussed this at DEVCON with fellow developers and most of them agreed with it, though I am open to hearing alternative points of view.

### DIFF
--- a/src/execution-context.js
+++ b/src/execution-context.js
@@ -80,7 +80,7 @@ web3VM.setVM(vm)
 function ExecutionContext () {
   var self = this
   this.event = new EventManager()
-  var executionContext = injectedProvider ? 'injected' : 'vm'
+  var executionContext = 'vm'
 
   this.getProvider = function () {
     return executionContext


axic:
👍  on this. Even though Metamask here is set to Ropsten it is still very confusing and **dangerous**.

axic:
Alternatively there could be a configuration setting, `always-use-vm`, influencing the default which is set to `true` by default. Knowledgable users could change it after the first use.

federicobond:
Or just make the choice of environment persist to localStorage like some other settings.

axic:
> Or just make the choice of environment persist to localStorage like some other settings.

That's what I've meant with "configuration setting".

yann300:
that would also be great to display the current network name (if Web3 provider or injected provider is selected).
I could put that in the coming PRs.

PR #919, opened 2017-11-30
Request proper legacyAST and legacyAssembly

axic:


axic:
@yann300 this is needed for the next release

PR #926, opened 2017-12-05
ADD network indicator

serapath:


src/execution-context.js
```diff
@@ -91,7 +91,21 @@ function ExecutionContext () {
   }
 
   this.web3 = function () {
-    return this.isVM() ? web3VM : web3
+    return this.isVM() ? web3VM.web3 : web3
+  }
+
+  this.detectNetwork = function (callback) {
```
axic: Please add a comment this follows EIP155: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

Perhaps worth supporting all those from the EIP.

serapath: done

PR #936, opened 2017-12-08
showGasEstimation

ninabreznik:
@yann300 I made a first commit to the issue - show gas estimations when function is selected.

Please check it out and give some feedbacks. I am still very shaky with this compilationResults and stuff so I don't know if I am pulling the right data...

screen shot 2017-12-08 at 10 54 47 pm


src/app/editor/contextView.js
```diff
@@ -159,7 +168,14 @@ class ContextView {
       <span class=${css.referencesnb}>${references}</span>
       <i data-action='previous' class="fa fa-chevron-up ${css.jump}" aria-hidden="true" onclick=${jump}></i>
       <i data-action='next' class="fa fa-chevron-down ${css.jump}" aria-hidden="true" onclick=${jump}></i>
+        ${showGasEstimation()}
     </div>`
+
+    function showGasEstimation () {
+      var estimatedGas = self._api.contextualListener.gasEstimation(node) + ' gas'
+      if (node.name === 'FunctionDefinition'  && !node.attributes.isConstructor) var el = yo`<div class=${css.gasEstimation}> <img class=${css.gasStationIcon} title='Estimated gas price' src='https://png.icons8.com/gas-station/win8/50/000000'> ${estimatedGas}</div>`
```
axic: Why ignore it for the constructor?

src/app/editor/contextualListener.js
```diff
@@ -94,6 +91,27 @@ class ContextualListener {
...
+    if (node.name === 'FunctionDefinition') {
+      var functionName = node.attributes.name
+      var fn
+      for (var x in this.functions) {
+        if (!!~this.functions[x].indexOf(functionName)) {
+          fn = this.functions[x]
+          break
+        }
+      }
+      var estimation = this.target[this.contractName].evm.gasEstimates.external[fn]
```
ninabreznik: @axic, I got estimation like this, not sure if this is correct, because for some functions it says infinite and for some there is a number, but for a constructor there was nothing. So I thought I have to retreive that one from somewhere else... Any thoughts?

src/app/editor/contextView.js
ninabreznik: @axic good point, forgot to mention this. Please check the other comment I left for you.

src/app/editor/contextualListener.js
axic: It can be a number or the `infinite`, both are correct. 

This is the structure returned by the compiler (not sure about remix internals now):
```
          // Function gas estimates
          gasEstimates: {
            creation: {
              codeDepositCost: "420000",
              executionCost: "infinite",
              totalCost: "infinite"
            },
            external: {
              "delegate(address)": "25000"
            },
            internal: {
              "heavyLifting()": "infinite"
            }
          }
```

For the constructor the `creation.totalCost` should be used.

Also make sure that external functions are listed in `external`, while internal ones in `internal`.

ninabreznik: Nice, will add it tomorrow. Thanks!

src/app/editor/contextView.js
ninabreznik: @axic I have new version of PR that includes your suggestions. Please take a look if you have time. I would like to test it with a contract that has internal functions, but don't have one. Hope it works, but would need to double check.

src/app/editor/contextView.js
```diff
@@ -159,7 +168,14 @@ class ContextView {
       <span class=${css.referencesnb}>${references}</span>
       <i data-action='previous' class="fa fa-chevron-up ${css.jump}" aria-hidden="true" onclick=${jump}></i>
       <i data-action='next' class="fa fa-chevron-down ${css.jump}" aria-hidden="true" onclick=${jump}></i>
+        ${showGasEstimation()}
     </div>`
+
+    function showGasEstimation () {
+      var estimatedGas = self._api.contextualListener.gasEstimation(node) + ' gas'
+      if (node.name === 'FunctionDefinition') var el = yo`<div class=${css.gasEstimation}> <img class=${css.gasStationIcon} title='Estimated gas price' src='https://png.icons8.com/gas-station/win8/50/000000'> ${estimatedGas}</div>`
```
yann300: is that possible to you use an icon from font-awesome ?

ninabreznik: @yann300 Unfortunately not :( Was searching there first, but it doesn't exist. I really wanted to use a gas station icon. [This one](https://ethgasstation.info/images/ETHGasStation.png) would be my favorite, but it's a logo of some project... Maybe @ryestew could help us create one? Or we go with this one for now and if/when fontawesome adds gas station icon, we switch. I saw people already opening issues and asking for this icon at fontawesome, but so far nothing yet...

yann300: ok then we need to import it in the project (we can't use an external link). you could convert it to base64

ninabreznik: @yann300 Should I maybe rather download it?

yann300: Yes it's fine

ninabreznik: @yann300 Is everything else ok otherwise?

src/app/editor/contextualListener.js
yann300: @ninabreznik is this possible to add
`codeDepositCost: <value> executionCost: <value>` in the title (in case of constructor)? or even on the main view (as you feel like). As we have the information it would be nice to show it.

src/app/editor/contextView.js
yann300: just one little add that could be nice to add (please check above)

src/app/editor/contextualListener.js
```diff
@@ -94,6 +91,43 @@ class ContextualListener {
...
+    this.creationCost = this.estimationObj.creation.totalCost
+    this.codeDepositCost = this.estimationObj.creation.codeDepositCost
+  }
+
+  gasEstimation (node) {
+    if (node.name === 'FunctionDefinition') {
+      if (!node.attributes.isConstructor) {
+        var functionName = node.attributes.name
+        if (this.externalFunctions) {
+          return {executionCost: this.estimationObj.external[this._getFn(this.externalFunctions, functionName)]}
```
axic: Instead of looking up the function in both external and internal, it should make use of the node attributes to decide which one it is.

Also the external lookup will not work because that has function signature, function name + arguments.

ninabreznik: Commited some changes, pls check if that's what you had in mind @axic

yann300: we still need to better search for the function signature, cause `functionName` does not contain it.

ninabreznik: @yann300 What do you mean by this exactly?

yann300: that `functionName` would be `get` for instance but in `this.externalFunctions` you will have `get(uint _v)`

ninabreznik: I see. It works now, right? It's just not very clean and elegant...

ninabreznik: Any suggestions?

ninabreznik: We could maybe rather use split? Let me try

src/app/editor/contextualListener.js
```diff
@@ -180,6 +129,59 @@ class ContextualListener {
...
+      if (this.nodes[i].id === node.attributes.scope) {
+        var contract = this._getGasEstimation(this.nodes[i])
+        break
+      }
+    }
+    return contract
+  }
+
+  _getInputParams (fnName) {
+    var abi = this.contract.abi
```
yann300: that 'll not help to use the abi here, you have to loop over InputParameter of the FunctionDefinition AST node.

src/app/editor/contextualListener.js
```diff
@@ -130,6 +129,66 @@ class ContextualListener {
...
+    this._getContract(node)
+    var executionCost
+    var codeDepositCost
+    if (node.name === 'FunctionDefinition') {
+      if (!node.attributes.isConstructor) {
+        var fnName = node.attributes.name
+        var fn = fnName + this._getInputParams(node)
+        if (node.attributes.visibility === 'public') {
+          executionCost = this.estimationObj.external[fn]
+        } else if (node.attributes.visibility === 'internal') {
```
axic: `private` or `internal` will make it to show up in the internal list

src/app/editor/contextualListener.js
```diff
@@ -130,6 +129,66 @@ class ContextualListener {
...
+
+  gasEstimation (node) {
+    this._getContract(node)
+    var executionCost
+    var codeDepositCost
+    if (node.name === 'FunctionDefinition') {
+      if (!node.attributes.isConstructor) {
+        var fnName = node.attributes.name
+        var fn = fnName + this._getInputParams(node)
+        if (node.attributes.visibility === 'public') {
```
axic: `public` or `external` will make it to show up in the external list

src/app/editor/contextualListener.js
```diff
@@ -130,6 +129,66 @@ class ContextualListener {
...
+      executionCost = '-'
+    }
+    return {executionCost, codeDepositCost}
+  }
+
+  _getContract (node) {
+    for (var i in this.nodes) {
+      if (this.nodes[i].id === node.attributes.scope) {
+        var contract = this._getGasEstimation(this.nodes[i])
+        break
```
axic: This seems to be bad practice. Even though horrible javascript scoping rules allow it, I'd would advise against it.

The simplest code is `return this._getGasEstimation...`

ninabreznik: Makes total sense. Thanks for noticing!

PR #944, opened 2017-12-13
add support for large directories to file-explorer

serapath:


serapath:
@yann300 
* added `.resolveDirectory(...)` to files object
* removed `.list(...)` and `.listAsTree(...)` from files object
* refactored project to use `.resolveDirectory` instead of `.list` or `.listAsTree`
* fixed bugs related to switching from sync to async api when always delegating to **remixd**

For the next step I would need the TreeView events for expand/collapse.

Then I could change the FileExplorer to use the new `.resolveDirectory(...)` to only load what the user requests.

Once I have that, I can remove more code from the files object, because `.resolveDirectory` is currently recursively fetching everything like `.listAsTree` did in the past, but on every user click only one level of depths is required. This would get rid of quite some code that I added now to make `.resolveDirectory` full replacement for `listAsTree`

src/app/compiler/compiler.js
```diff
@@ -49,7 +49,9 @@ function Compiler (handleImportCall) {
         self.lastCompilationResult = null
         self.event.trigger('compilationFinished', [false, {'error': { formattedMessage: error, severity: 'error' }}, files])
       } else {
-        compileJSON(input, optimize ? 1 : 0)
+        setTimeout(function () {
```
yann300: why?

src/app/files/browser-files.js
```diff
@@ -11,15 +11,13 @@ function Files (storage) {
...
-      return false
-    }
+    if (path === '.remix.config') return false
 
     return this.isReadOnly(unprefixedpath) || storage.exists(unprefixedpath)
   }
 
   this.init = function (cb) {
-    cb()
+    this.resolveDirectory('', (error, filesTree) => cb && cb(error))
```
yann300: not sure we need to `resolveDirectory` in the init function

src/app/files/browser-files.js
```diff
@@ -49,6 +47,7 @@ function Files (storage) {
         return false
       }
       if (!exists) {
+        this.init()
```
yann300: I am pretty sure `this.init()` should not be called here (we are in `set()`)

src/app/files/browser-files.js
```diff
@@ -63,6 +62,7 @@ function Files (storage) {
     var unprefixedpath = this.removePrefix(path)
     if (!storage.exists(unprefixedpath)) {
       readonly[unprefixedpath] = content
+      this.init()
```
yann300: same as above

src/app/files/browser-files.js
```diff
@@ -140,40 +119,57 @@ function Files (storage) {
...
-      var i = 0
-
-      for (; i < nodes.length - 1; i++) {
-        var node = nodes[i]
-        if (obj[node] === undefined) {
-          obj[node] = {}
+  this.resolveDirectory = function (path, callback) {
+    var self = this
+    // path = '' + (path || '')
+    setTimeout(function () {
```
yann300: why `setTimeout` here?

yann300:
the TreeView events are already here: https://github.com/ethereum/remix/pull/645/files

src/app/files/fileManager.js
```diff
@@ -166,3 +172,23 @@ class FileManager {
 }
 
 module.exports = FileManager
+
+function flatten (tree) {
```
yann300: the param is a `tree` but above a list is used : `var fileList = Object.keys(flatten(filesList))` as input

src/app/compiler/compiler.js
serapath: because it doesnt work without :/

src/app/files/browser-files.js
serapath: yes, probably it's not necessary. I will try without - i don't know why i thought it was necessary - i have to get back into it :-)

src/app/files/browser-files.js
serapath: yeah, i'll try

src/app/files/browser-files.js
serapath: to make `browser-files.js` and `shared-folder.js` work the same.
The latter returns in the next tick, because it is really asynchronous.
The former returns syncronous in the same tick without `setTimeout`

src/app/files/fileManager.js
serapath: oh good catch. yes, i fixed it. it's now `flatten(filesTree)`

PR #946, opened 2017-12-13
ADD npm run remixd + mochfolder

serapath:


serapath:
to make it more convenient to develop https://github.com/ethereum/browser-solidity/issues/945 and other file-explorer features

yann300:
Instead of adding these contracts in `test-browser`, could it be directly in a `contracts` folder.

serapath:
ok

PR #968, opened 2017-12-29
Fix TypeError when using older solc

### LINKED ISSUE #967: old solc versions don't work
The old solc versions don't work when selected in the Settings tab. Only working versions should appear in the drop-down.

ewiner:
Fixes #967.

On solc 0.4.10 or earlier, `linkReferences` is undefined.  Without this change, you can compile contracts against old solidity compilers, but you can't create them in the IDE.


yann300:
thanks for this push!

PR #971, opened 2018-01-02
Ask to confirm tx

ninabreznik:
Applied basic logic and basic styling

TODOS:
- [x] seems like i get pending transaction when on injected web3, even if I click cancel on modal dialog

### DIFF
--- a/src/app/execution/txRunner.js
+++ b/src/app/execution/txRunner.js
@@ -245,6 +245,8 @@ function confirmDialog (tx, gasEstimation, self) {
     </div>
   </div>
   `
+
+  var warnMessage = ' Please fix this issue before sending any transaction. '
   function gasPriceChanged () {
     try {
       var gasPrice = el.querySelector('#gasprice').value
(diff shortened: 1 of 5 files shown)


axic:
Is it possible to have a setting for this? Metamask will pop up its own confirmation dialog, having two might be annoying, while when using other endpoints which do not have a confirmation window this change is very beneficial.

src/app/execution/txRunner.js
```diff
@@ -44,43 +57,51 @@ TxRunner.prototype.execute = function (args, callback) {
...
 
-        tx.gas = gasEstimation
-        var sendTransaction = self.personalMode ? executionContext.web3().personal.sendTransaction : executionContext.web3().eth.sendTransaction
-        try {
-          sendTransaction(tx, function (err, resp) {
-            if (err) {
-              return callback(err, resp)
-            }
+              tx.gas = gasEstimation
+              var sendTransaction = self.personalMode ? executionContext.web3().personal.sendTransaction : executionContext.web3().eth.sendTransaction
```
yann300: the `modalDialog` should be just before this line and after the gasEstimation has been calculated

src/universal-dapp.js
```diff
@@ -156,6 +156,7 @@ function UniversalDApp (opts = {}) {
   var self = this
 
   self._api = opts.api
+  console.log(self._api.config)
```
yann300: please, could you remove that log?

src/app/tabs/settings-tab.js
```diff
@@ -79,46 +94,65 @@ function SettingsTab (container, appAPI, appEvents, opts) {
...
+          <div>
+            <textarea rows="4" cols="70" id="plugininput" type="text" class="${css.pluginTextArea}" ></textarea>
+            <input onclick=${loadPlugin} type="button" value="Load" class="${css.pluginLoad}">
+            </div>
         </div>
       </div>
     </div>
+
   `
+  function askToConfirmTx () {
```
yann300: the doNotShowAgain should not be persisted

src/app/execution/txRunner.js
```diff
@@ -68,17 +122,25 @@ TxRunner.prototype.execute = function (args, callback) {
         }
 
         tx.gas = gasEstimation
-        var sendTransaction = self.personalMode ? executionContext.web3().personal.sendTransaction : executionContext.web3().eth.sendTransaction
-        try {
-          sendTransaction(tx, function (err, resp) {
+
+        if (!self.config.get('doNotShowAgain')) {
```
yann300: same as below: `doNotShowAgain` should not be persisted

src/app/execution/txRunner.js
```diff
@@ -42,15 +60,51 @@ TxRunner.prototype.execute = function (args, callback) {
...
+            transactionHash: result.transactionHash
+          })
+        })
+      }
       tx.gas = gasLimit
-      executionContext.web3().eth.call(tx, function (error, result) {
-        callback(error, {
-          result: result,
-          transactionHash: result.transactionHash
+      if (!self.config.get('doNotShowAgain')) {
```
yann300: in case of a `call` ( `useCall` is `true` ), the modal should not be displayed: we are not going to create a tx in that case

src/app/tabs/settings-tab.js
```diff
@@ -79,46 +94,65 @@ function SettingsTab (container, appAPI, appEvents, opts) {
...
-        <input class="${css.col1}" name="theme" id="themeDark" type="radio">
-        <label for="themeDark">Dark Theme</label>
+      <div class="${css.info}">
+      <div class=${css.title}>General settings</div>
+        <div class="${css.crow}">
+          <div>${optionVM}</div>
+          <span class="${css.checkboxText}">Always use Ethereum VM at Load</span>
+        </div>
+        <div class="${css.crow}">
+          <div><input id="askToConfirm" onchange=${askToConfirmTx} type="checkbox"></div>
```
yann300: as `askToConfirmTx` is not persisted, no need to as a config for that

yann300:
a few comment about the modal dialog content:
 - when the tx is canceled, we there's still `1 pending transaction`
 - `Amount` instead of `value`
 
for the next PR:
 - hex decimal value should be converted to decimal number (e.g 5 ether instead of 0x05)
 - `data` should display the whole content in a `pre` tag
 - gas estimate should be displayed
 - display gas Price
 - display max transaction fee (using the gas limit)

PR #989, opened 2018-01-15
add support for large file hierarchies to file explorer

serapath:


src/app/files/basicReadOnlyExplorer.js
```diff
@@ -81,7 +81,18 @@ class BasicReadOnlyExplorer {
...
+    var tree = {}
+    // This does not include '.remix.config', because it is filtered
+    // inside list().
+    Object.keys(this.list()).forEach(function (path) {
+      hashmapize(tree, path, {
+        '/readonly': self.isReadOnly(path),
+        '/content': self.get(path)
+      })
+    })
+    return callback(null, tree[path] || {})
```
yann300: return `tree[path]` always undefined

src/app/files/file-explorer.js
```diff
@@ -99,8 +100,9 @@ function fileExplorer (appAPI, files) {
         })) : undefined
       }
     },
-    formatSelf: function (key, data, li) {
-      var isRoot = data.path.indexOf('/') === -1
+    formatSelf: function formatSelf (key, data, li) {
+      var isRoot = !data.path.indexOf('browser') && data.path.length === 'browser'.length
```
yann300: why `var isRoot = data.path.indexOf('/') === -1` is not more applicable? I think it still is

src/app/files/file-explorer.js
```diff
@@ -138,10 +162,9 @@ function fileExplorer (appAPI, files) {
   var textUnderEdit = null
   var textInRename = false
 
-  var events = new EventManager()
-  this.events = events
-  var api = {}
-  api.addFile = function addFile (file) {
+  self.events = new EventManager()
+  self.api = {}
+  self.api.addFile = function addFile (file) {
```
yann300: the `addFile`  button seems to be broken

src/app/files/basicReadOnlyExplorer.js
serapath: it's not always undefined.
sometimes it return `{ localhost: { foo: {}, bar: {}, baz: {} }`

...or ...hmm, maybe not? I think without it, the names like `gist`, `github`, etc... showed up in the file explorer

src/app/files/file-explorer.js
serapath: hmm... maybe.
I think i had  scenario where this wasnt working, but don't remember what it was.

:/

src/app/files/file-explorer.js
serapath: hm - i didn't touch the `addFile` button, but it's probably related to emitting an event - and i think i added `@TODO` tags to those - so yes that needs fixing

src/app/files/basicReadOnlyExplorer.js
yann300: well when I tested it with remixd it would always return `undefined`.
use publish to swarm to reproduce

yann300: ```I think without it, the names like gist, github, etc... showed up in the file explorer```
that's ok

src/app/files/file-explorer.js
yann300: Let's fix in the next PR

src/app/files/file-explorer.js
yann300: could you rollback for now?

yann300:
the Ballot.sol default file is not loaded by default anymore...

serapath:
oh really? 
hmm, it loaded for me

wowa1983:
Kann mich nicht einloggen zu github habe neue Teil.nummer wie kann ich
Tellefon zurück setzen athetivikator

Am 15.01.2018 18:38 schrieb "Alexander Praetorius" <notifications@github.com

src/app/files/file-explorer.js
serapath: ok

PR #991, opened 2018-01-16
Recorder: resolve input parameters that might contains "created{address}" 

yann300:


daoudamr:
my computer is broke lost all my developer apps still have acces to etherum explorer i need to finish my conract tokens the whole ethereum system is lagging because of this

daoudamr:
my email is daoudamr52@gmail.com my eyes hurt from working with this different computer and trying to find all my stuff i need assistance i have the block 40029999 its been stuck because the tokens i created i did the coding wrong without any value so dao and omg tokens are all stuck i dont know what to do i would give up but its lagging down the whole system

PR #992, opened 2018-01-16
Refactor jquery => JS for static analysis messages

ninabreznik:


yann300:
I think we don't need anymore this PR

PR #473, opened 2017-03-16
Support personal mode of RPC [WIP]

axic:
Fixes #198.

I've did this a while back, but I won't have time to finish with a nice UI.

PR #993, opened 2018-01-17
Integrate personal mode

yann300:


axic:
Can you please pull in the all the features from #473 (and close it after)?

e.g. after account creation the account list must be refreshed.

PR #1000, opened 2018-01-21
Initial commit to allow hex encoded transaction param

eswarasai:
https://github.com/ethereum/remix/issues/85

### DIFF
--- a/src/app/execution/txFormat.js
+++ b/src/app/execution/txFormat.js
@@ -47,28 +47,34 @@ module.exports = {
     */
   buildData: function (contractName, contract, contracts, isConstructor, funAbi, params, udapp, callback, callbackStep) {
     var funArgs = ''
-    try {
-      funArgs = $.parseJSON('[' + params + ']')
-    } catch (e) {
-      callback('Error encoding arguments: ' + e)
-      return
-    }
     var data = ''
     var dataHex = ''
-    if (!isConstructor || funArgs.length > 0) {
+
+    if (params.startsWith('"0x')) {
+      dataHex = params.match(/"(.*?)"/)[1].replace('0x', '')
+      data = Buffer.from(dataHex, 'hex')
+    } else {
       try {
-        data = helper.encodeParams(funAbi, funArgs)
-        dataHex = data.toString('hex')
+        funArgs = $.parseJSON('[' + params + ']')
       } catch (e) {
         callback('Error encoding arguments: ' + e)
         return
       }
-    }
-    if (data.slice(0, 9) === 'undefined') {
-      dataHex = data.slice(9)
-    }
-    if (data.slice(0, 2) === '0x') {
-      dataHex = data.slice(2)
+      if (!isConstructor || funArgs.length > 0) {
+        try {
+          data = helper.encodeParams(funAbi, funArgs)
+          dataHex = data.toString('hex')
+        } catch (e) {
+          callback('Error encoding arguments: ' + e)
+          return
+        }
+      }
+      if (data.slice(0, 9) === 'undefined') {
+        dataHex = data.slice(9)
+      }
+      if (data.slice(0, 2) === '0x') {
+        dataHex = data.slice(2)
+      }
     }
     var contractBytecode
     if (isConstructor) {
--- a/src/app/execution/txFormat.js
+++ b/src/app/execution/txFormat.js
@@ -50,7 +50,7 @@ module.exports = {
     var data = ''
     var dataHex = ''
 
-    if (params.startsWith('"0x')) {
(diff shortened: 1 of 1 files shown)


eswarasai:
@yann300 -- So I've done the following change :
1) Added a check to skip parse and encode the param if it startsWith `0x`

I wasn't able to test this though as I couldn't figure out with the Contract setup having transaction input as an array. Let me know if you can help me with that and also any other changes that are needed to be taken care of

src/app/execution/txFormat.js
```diff
@@ -88,7 +92,11 @@ module.exports = {
         dataHex = bytecodeToDeploy + dataHex
       }
     } else {
-      dataHex = Buffer.concat([helper.encodeFunctionId(funAbi), data]).toString('hex')
+      if (params.startsWith("0x")) {
+        dataHex = params;
+      } else {
+        dataHex = Buffer.concat([helper.encodeFunctionId(funAbi), data]).toString('hex')
```
eswarasai: @yann300 -- I'm not sure of this check here though. I've added it just to avoid the access of `data` variable because it'd be empty string if param is a hex encoded value as per the logic above

yann300: Well then in your case the input given by the user has to contain the function id.
I think for a first commit, we should stick to the fact that `params` will only contain the encoded input params and not the function id . We could be more fine-grain later on if necessary.
So from `dataHex` , I am sure we can generate `data`.

eswarasai: Sure. Let me try that and push the change. Thanks.

yann300: @eswarasai  is everything ok with that PR, do yo need something?

eswarasai: @yann300 -- Sorry for the delay in getting back to you. Yeah, so about getting the `data` from the `dataHex` is essentially not required as we won't have the `functionId`. I guess this PR should handle the requirement. Let me know a way to test it out and confirm that everything works as expected. Thanks.

yann300: you do have the `functionId` at this point (it is part of `funabi`)
The actual state of the PR assumes that the hex value you input contains the function id.
you should assumes that the hex value you input only contain the encoded data (there's room for improvement but that's another PR)
to test it, you can have:

```
contract test {
    function t (uint _p) constant returns (uint) {
        return _p;
    }
}
```
call `t ()` with the following input
`"0x000000000000000000000000000000000000000000000000000000000000000a"`
the return value should be `10`

eswarasai: @yann300 -- I've updated the code now. Below are the modifications :
1) Added specific check to see if the `params` value specifically `startsWith` `"0x`. Let me know is this needs to be changed to `contains` - `0x`
2) Converting the input param to `dataHex`
3) Getting `data` from the `dataHex`

I've used the above test scenario and output always seemed to be `uint256: 10` with and without `0x` input. Let me know if there's any other modification required. Thanks.

eswarasai:
@yann300 -- I've updated the code now. Below are the modifications :
1) Added specific check to see if the `params` value specifically `startsWith` `"0x`. Let me know is this needs to be changed to `contains` - `0x`
2) Converting the input param to `dataHex`
3) Getting `data` from the `dataHex`

I've used the above test scenario and output always seemed to be `uint256: 10` with and without `0x` input. Let me know if there's any other modification required. Thanks.

eswarasai:
@yann300 -- I've seen that you've pushed a change to modify the check whether `params` starts with `0x` or not. The reason I've added `"` at the start of `0x` is that, I'm getting the `params` value as `"0x13sda12......"`. Keeping this in mind, I'm not sure if the check `indexOf('0x') === 0` would evaluate to `true`. Please do correct me if I'm wrong. Thanks.

yann300:
I changed `startsWith` to `indexOf` for browser support.

Also if user input `"0xca35b7d915458ef540ade6068dfe2f44e8fa733c", 3`  this start with `"0x` and is not an encoded input (an address as first arg and uint as second) - and that should be encoded.
whereas if user just input `0xabcd1234` that should be interpreted by us as an encoded input.
There are possibly improvement to do here but not in this PR.

so the regex check is not necessary too. could you remove it? then I think it can be merged. I will push test for that later today.

eswarasai:
@yann300 -- Got it. I'll quickly do that change and push it here.

Thanks for pointing out the browser support issue. I'll keep out an eye going forward 👍

wowa1983:
Yann300 wie mache ich das runter mit welches schprache und Befehl

wowa1983:
Gas*runter

PR #1003, opened 2018-01-22
Update README.md

yann300:
Renaming


yann300:
that can be closed. renaming was done in another PR.
just going create a small one

PR #1013, opened 2018-01-24
Fix flexbox for Environment field in Run tab + info icon

ninabreznik:
This PR doesn't work if PR in Remix repo is not merged too, since i added some styling to Style-guide.

PS: We have to think about how to improve this flow, it's really hard to have to work on 2 repos + you can't see the result, because you can also get the variable from style-guide when it's merged in the Remix repo.

### DIFF
--- a/src/app/tabs/run-tab.js
+++ b/src/app/tabs/run-tab.js
@@ -44,6 +44,11 @@ var css = csjs`
     float: left;
     align-self: center;
   }
+  .environment {
+    display: flex;
+    align-items: center;
+    position: relative;
+  }
   .col2 {
     ${styles.rightPanel.runTab.input_RunTab}
   }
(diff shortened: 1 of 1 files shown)


src/app/tabs/run-tab.js
```diff
@@ -546,7 +555,7 @@ function settings (container, appAPI, appEvents) {
               Web3 Provider
             </option>
           </select>
-          <a style="margin-left: 8px;" href="https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md" target="_blank"><i class="fa fa-info"></i></a>
+          <a style="margin-left: 8px;" href="https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md" target="_blank"><i class="${css.info} fa fa-info"></i></a>
```
yann300: there is no `info` property define

PR #1014, opened 2018-01-24
automate remix-* installation

serapath:


yann300:
closing, we should find another way. 
The current is not optimal but functional for now

Issue #1012, opened 2018-01-24
Show contract outputs even if contract does not compile


fulldecent:
MOVED FROM UPSTREAM: https://github.com/ethereum/solidity/issues/3428

Currently, Remix will not compile if contract/interface methods are left unimplemented.

Instead Remix should run if there are no syntactical/grammar errors.

Why? Because the compiler parses NatSpec and NatSpec is useful even for incomplete contracts and for interfaces.

![screen shot 2018-01-23 at 7 52 08 pm](https://user-images.githubusercontent.com/382183/35308745-294cc518-0077-11e8-922f-bbe940ea9d9e.png)
![screen shot 2018-01-23 at 7 52 55 pm](https://user-images.githubusercontent.com/382183/35308751-315ae7e4-0077-11e8-8107-ce6fc8b56931.png)
![screen shot 2018-01-23 at 7 53 11 pm](https://user-images.githubusercontent.com/382183/35308758-356a258e-0077-11e8-88de-6de9d06ed768.png)
![screen shot 2018-01-23 at 7 53 14 pm](https://user-images.githubusercontent.com/382183/35308761-37c99ad0-0077-11e8-87dc-aa98f86e0b46.png)

yann300:
The user doc and dev doc is not shown yet (even if this is not a virtual contract). 
That's still to be added, hopefully soon

yann300:
we should add the userdoc and devdoc tag to the compiler standard input

axic:
Fixed by #1018.

fulldecent:
Very nice, thank you!

PR #1018, opened 2018-01-25
add devdoc and userdoc to compiler input and contract details #1012

iurimatias:
#1012


iurimatias:
note: solc-js itself still outputs an empty metadata ""

axic:
Metadata generation happens as part of bytecode outputs at the moment, IIRC. Please open an issue on the Solidity repo with some examples (running it on `solc`), we'll address ti.

PR #1022, opened 2018-01-29
Accordian meth props

ryestew:


yann300:
replace by https://github.com/ethereum/remix-ide/pull/1212

PR #1024, opened 2018-01-31
Do not skip balance and nonce

yann300:


src/app/execution/txRunner.js
```diff
@@ -175,7 +175,7 @@ TxRunner.prototype.execute = function (args, callback) {
         executionContext.vm().stateManager.checkpoint()
       }
 
-      executionContext.vm().runTx({block: block, tx: tx, skipBalance: true, skipNonce: true}, function (err, result) {
+      executionContext.vm().runTx({block: block, tx: tx, skipBalance: false, skipNonce: false}, function (err, result) {
```
axic: The only reason this was present was inadequate support for tracking both in browser-solidity and the VM had bugs as well. I think it should be safe to remove it - it definitely will be better in the long term but some issues may creep up which need fixing.

Issue #210, opened 2016-09-01
'At address' function should verify that the input is an address


axic:
It should:
- disallow any whitespace
- validate that it is a hex number (starts with `0x`)
- it is 20 bytes (40 digits) long (excluding the `0x`)
- optionally: verify checksum if it is a checksummed address

There are two useful methods in ethereumjs-util which can be used for the above: `isValidAddress()` (does whitespace, hex and length validation) and `isValidChecksumAddress()`.

chriseth:
@axic please provide more information about how to solve this issue (especially as it is an "up-for-grabs" issue). What checks are you thinking of? Do you want to verify the checksum or just check that it is hex with 20 bytes?

axic:
It should:
- disallow any whitespace
- validate that it is a hex number (starts with `0x`)
- it is 20 bytes (40 digits) long
- optionally: verify checksum if it is a checksummed address

There are two useful methods in ethereumjs-util which can be used for the above: `isValidAddress()` (does whitespace, hex and length validation) and `isValidChecksumAddress()`.

chriseth:
Anyone: You can start working on this if you want. Please ask questions in https://gitter.im/ethereum/solidity or https://gitter.im/ethereum/cpp-ethereum-development

axic:
It is still active.

gitcoinbot:
__This issue now has a funding of 0.03 ETH (31.99 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/browser-solidity/issues/210).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/browser-solidity/issues/210)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $11920.2 more Funded OSS Work Available at: https://gitcoin.co/explorer

gitcoinbot:
__Work has been started on 0.03 ETH (31.99 USD) by: [@eswarasai](https://gitcoin.co/profile/eswarasai).__ 

 If you've started work, please leave a comment to let the funder (@owocki) and the other parties involved what you're working, with respect to this issue and your plans to resolve it.  If you don't leave a comment, the funder may expire your submission at their discretion. 

[...]

PR #1025, opened 2018-01-31
Initial commit to check for valid address

eswarasai:
#210

PR #1026, opened 2018-01-31
#210 Checking checksum and valid address

Alexintosh:
Simple implementation, looking into testing strategy.

I think e2e would the right approach here, I couldn't run Selenium test so far, will look into it.


yann300:
duplicate (https://github.com/ethereum/remix-ide/pull/1025#pullrequestreview-103088461) has been merged

PR #1029, opened 2018-02-02
Terminal ui 

ninabreznik:


test-browser/tests/ballot.js
```diff
@@ -79,8 +79,8 @@ function runTests (browser, testData) {
       })
       .perform((client, done) => {
         console.log('delegate - transact (not payable)')
-        browser.testFunction('delegate - transact (not payable)', '0xd3cd54e2f76f3993078ecf9e1b54a148def4520afc141a182293b3610bddf10f',
-              '[vm] from:0xca3...a733c, to:Ballot.delegate(address) 0x692...77b3a, value:0 wei, data:0x5c1...4d2db, 0 logs, hash:0xd3c...df10f',
+        browser.testFunction('delegate - transact (not payable)', '0x057d54e2f76f3993078ecf9e1b54a148def4520afc141a182293b3610bd3ce2e',
```
yann300: display has changed but the executed tx remain the same. So changing hash here will break test

test-browser/tests/compiling.js
```diff
@@ -43,8 +43,8 @@ function testSimpleContract (browser, callback) {
       .click('#runTabView .instance div[class^="title"]')
       .click('#runTabView .instance div[class^="title"]')
       .testFunction('f - transact (not payable)',
-        '0xa178c603400a184ce5fedbcfab392d9b77822f6ffa7facdec693aded214523bc',
-        '[vm] from:0xca3...a733c, to:TestContract.f() 0x692...77b3a, value:0 wei, data:0x261...21ff0, 0 logs, hash:0xa17...523bc', null,
+        '0x057d54e2f76f3993078ecf9e1b54a148def4520afc141a182293b3610bd3ce2e',
```
yann300: same as above

test-browser/tests/ballot.js
ninabreznik: How should it be now? I live everything the same here, just add new lines to the log?

yann300: the secon argument should remain the same

yann300: hash is the 2nd argument

ninabreznik: So, if I understand correctly I should switch hash back to `hash:0xd3c...df10f`?

yann300: yes the hash is the same as before everywhere , you just changed the UI.

ninabreznik:
Hash in compiling.js and ballot.js is changed back to `hash:0xd3c...df10f`. Changes pushed!

ninabreznik:
@yann300 Changed a hash as suggested but still getting same error. Can you pls check it out!

ninabreznik:
@yann300 Unfortunately your fixes didn't work. Table with details didn't open. 

I tried to fix the PR without switching onclick txDetails to only caret, because I think it's going to be very hard for users to find the details if the whole log is not clickable.

I also fixed some additional little visual things, but as always can't make the tests pass.

I would suggest to rethink the tests, because if every time we want to do any visual changes, we need to go through that much hassle, then we will be stuck. Let's see how tests can work for us not against us.

How to proceed here?

yann300:
let's discuss about that.

yann300:
I will take a look on it

Hasbullah110782:
Aplock kontrak

PR #1037, opened 2018-02-06
separate universal-dapp logic from its UI code

iurimatias:


src/universal-dapp.js
```diff
@@ -159,17 +23,14 @@ function UniversalDApp (opts = {}) {
   self._api = opts.api
   self.removable = opts.opt.removable
   self.removable_instances = opts.opt.removable_instances
-  self.el = yo`<div class=${css.udapp}></div>`
-  self.contracts
-  self.transactionContextAPI
+  self.personalMode = opts.opt.personalMode || false
```
yann300: we don't need this property anymore , `personalMode()` should be a function of api

PR #1041, opened 2018-02-08
refactor udapp runTx and related functions

iurimatias:


src/universal-dapp.js
```diff
@@ -201,99 +202,52 @@ UniversalDApp.prototype.getInputs = function (funABI) {
...
+        return self.transactionContextAPI.getGasLimit(next)
       }
-      tx.from = ret[0]
-      next(null, env)
-    })
-  }
-}
+      next(null, 3000000)
+    },
+    function getAccount (gasLimit, next) {
```
yann300: seems that this will always be called. `getAccount` and `getValue` should be called only if the value is not set (see line 219).

PR #1042, opened 2018-02-08
Display a different warning for the alpha website

axic:


axic:
@yann300 since you have merged this, a change should be made which pushes `master` to `remix-ide-alpha` instead of `yann300`.

axic:
Well I guess should wait first until the build is pushed to `yann300` to start displaying the popup :)

PR #1045, opened 2018-02-08
move most tx code to remix-lib

iurimatias:
sister PR on [remix](https://github.com/ethereum/remix/pull/684)


iurimatias:
note: decodeResponseToTreeView was moved to udapp from txFormat as it was causing issues due to the remix-debugger dependency on remix-lib. It also makes sense to have it outside remix-lib anyway since it's UI specific

iurimatias:
note: the tests here will fail until the equivalent remix-lib PR is merged

PR #1051, opened 2018-02-10
UPDATE best practice with project structure

serapath:


yann300:
let's reopen this one when we actually support `require('ui/dropdown')`

PR #1053, opened 2018-02-10
Restyle transactionActions and unify icons with other icons

ninabreznik:
First and second task from https://github.com/ethereum/browser-solidity/issues/824
screen shot 2018-02-11 at 12 34 30 am

### DIFF
--- a/src/app/tabs/styles/run-tab-styles.js
+++ b/src/app/tabs/styles/run-tab-styles.js
@@ -112,6 +112,7 @@ var css = csjs`
     display: flex;
     justify-content: space-evenly;
     align-items: center;
+    flex-wrap: wrap;
   }
   .item {
     margin-right: 1em;
--- a/src/app/tabs/styles/run-tab-styles.js
+++ b/src/app/tabs/styles/run-tab-styles.js
@@ -177,8 +177,11 @@ var css = csjs`
   }
   .transactionActions {
     display: flex;
-    width: 50px;
+    width: 70px;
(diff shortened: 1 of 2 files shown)


ninabreznik:
@yann300 This one is also ready to merge...

yann300:
the `0 pending transaction` label is fixed?

ninabreznik:
@yann300 Fun fact. I was now checking this pull request for like an hour, re-adding but by bit to see where the pending transactions text disappeared and couldn't find it. And then I checked the live version on remix.ethereum.org and saw that it also doesn't appear there. So it looks like the problem is not in this PR, but we lost the label before somewhere...

yann300:
oh ok right :) thanks for checking that

src/app/tabs/styles/run-tab-styles.js
```diff
@@ -170,13 +175,21 @@ var css = csjs`
...
   .transactionActions {
-    float: right;
+    display: flex;
+    width: 70px;
+    justify-content: space-between;
+    border: 1px solid ${styles.rightPanel.runTab.additionalText_Color};
+    padding: 5px;
+    border-radius: 3px;
+}
+
```
yann300: ... standard checking should not pass here....

yann300: ah it is because this is inside the string

ninabreznik: @yann300 All good?

PR #1070, opened 2018-02-13
Fix file exist (from fileProvider)

### LINKED ISSUE #1065: Fix createNonClashingName
it should be `async`
it also means that we should make `fileprovider.exists` async for every provider.

yann300:
fixes #1065

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -3,6 +3,7 @@
 var $ = require('jquery')
 var csjs = require('csjs-inject')
 var yo = require('yo-yo')
+var async = require('async')
 var remixLib = require('remix-lib')
 var EventManager = remixLib.EventManager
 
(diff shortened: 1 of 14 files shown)


src/app.js
```diff
@@ -3,6 +3,7 @@
 var $ = require('jquery')
 var csjs = require('csjs-inject')
 var yo = require('yo-yo')
+var async = require('async')
```
axic: Probably not a good idea `async` is a keyword in ES6?

iurimatias: That's a very interesting point. I found this discussion from a year ago in their github repo [caolan/async#1179](https://github.com/caolan/async/issues/1179)

It's not necessarily a big issue right unless we start using ES7 and the functionality itself. We were already using this lib in other sections of the code and this is not specific to this PR. So I suggest that if we do want to address did then we can do a different PR that changes all references to ```async``` to ```asyncJS```

PR #1084, opened 2018-02-14
refactor righthandpanel

serapath:
I changed for the module RightHandPanel and all it's Tabs modules exactly one thing:
```js
var namedTab = new NamedTab({ api: {}, events: {} })
var element = namedTab.render()
```
RighthandPanel and all Tabs now follow the above convention for the interface.
* take an arguments `var opts = { api: {}, events: {} }`
* return an object with a `.render()` method

This will allow me to make a series of additional pull requests for each of these modules without the need to touch more than one file


src/app.js
```diff
@@ -660,7 +660,10 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
     editor: editor.event,
     staticAnalysis: staticanalysis.event
   }
-  self._components.righthandpanel = new RighthandPanel(rhpAPI, rhpEvents, {})
+  var rhpOpts = { api: {}, events: {} }
```
yann300: we just want to remove the container argument in this PR. 
could we keep having `(api, event, opt)` ?

src/app.js
```diff
@@ -840,7 +843,7 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
   })
 
   function startdebugging (txHash) {
-    self.event.trigger('debuggingRequested', [])
+    self._components.righthandpanel.showDebugger()
```
yann300: not sure that is needed to replace that now

src/app/panels/righthand-panel.js
```diff
@@ -2,29 +2,31 @@ var yo = require('yo-yo')
...
+var AnalysisTab = require('../tabs/analysis-tab')
+var DebuggerTab = require('../tabs/debugger-tab')
+var SupportTab = require('../tabs/support-tab')
+var PluginTab = require('../tabs/plugin-tab')
 var PluginManager = require('../../pluginManager')
 
 var css = require('./styles/righthand-panel-styles')
 
-function RighthandPanel (appAPI, events, opts) {
+function RighthandPanel ({ api: {}, events: {} } = {}) {
```
yann300: see comment above

src/app/panels/righthand-panel.js
```diff
@@ -2,29 +2,31 @@ var yo = require('yo-yo')
...
   var optionViews = yo`<div id="optionViews"></div>`
-  var options = yo`
-    <ul class=${css.opts}>
-    </ul>
-  `
   self._view.dragbar = yo`<div id="dragbar" class=${css.dragbar}></div>`
+  // load tabbed menu component
+  var tabEvents = {compiler: events.compiler, app: events.app, rhp: self.event}
+  self._view.tabbedMenu = new TabbedMenu({ api: self._api, events: tabEvents })
+  var options = self._view.tabbedMenu.render()
```
yann300: I think you forgot the `ul` here

src/app/tabs/analysis-tab.js
```diff
@@ -1,13 +1,12 @@
...
 
-function analysisTab (container) {
+function analysisTab ({ api: {}, events: {} } = {}) {
   var el = yo`
     <div class="${css.analysisTabView} "id="staticanalysisView">
     </div>
   `
-  container.appendChild(el)
-  return el
+  return { render () { return el } }
```
yann300: can we add `render` as a proper function of `analysisTab` here?

src/app/tabs/compile-tab.js
```diff
@@ -303,7 +300,7 @@ function compileTab (container, appAPI, appEvents, opts) {
     }
     return el
   }
-  return el
+  return { render () { return el } }
```
yann300: same here could we put `render` as a function of compile tab?

src/app/tabs/analysis-tab.js
serapath: yes, i will make a few more PRs to bring tab modules in shape

src/app/panels/righthand-panel.js
```diff
@@ -2,29 +2,31 @@ var yo = require('yo-yo')
...
+var AnalysisTab = require('../tabs/analysis-tab')
+var DebuggerTab = require('../tabs/debugger-tab')
+var SupportTab = require('../tabs/support-tab')
+var PluginTab = require('../tabs/plugin-tab')
 var PluginManager = require('../../pluginManager')
 
 var css = require('./styles/righthand-panel-styles')
 
-function RighthandPanel (appAPI, events, opts) {
+function RighthandPanel (api = {}, events = {}, opts = {}) {
```
yann300: please keep using `appAPI` here

src/app.js
yann300: is that possible to roll back that?

src/app/tabs/compile-tab.js
```diff
@@ -14,9 +14,9 @@ var remixLib = require('remix-lib')
 var styleGuide = remixLib.ui.themeChooser
 var styles = styleGuide.chooser()
 
-function compileTab (container, appAPI, appEvents, opts) {
-  if (typeof container === 'string') container = document.querySelector(container)
-  if (!container) throw new Error('no container given')
+function compileTab (api = {}, events = {}, opts = {}) {
+  var appAPI = api
```
yann300: by changing the argument from `appAPI` to `api`, you have to do this ^
please keep appAPI here every occurence (if not to many) of `appAPI` to `api`

src/app/panels/righthand-panel.js
```diff
@@ -40,27 +42,37 @@ function RighthandPanel (appAPI, events, opts) {
...
 
   self.pluginManager = new PluginManager(appAPI, events)
   events.rhp.register('plugin-loadRequest', (json) => {
-    var content = pluginTab(optionViews, json.url)
+    var content = new PluginTab(appAPI, events, json)
     this._view.tabbedMenu.addTab(json.title, 'plugin', content)
     self.pluginManager.register(json, content)
   })
 
+  self.showDebugger = function () { appAPI.switchTab('debugView') }
```
yann300: if we rollback and keep using ``debuggingRequested`, I think this ^ is not needed anymore

src/app/tabs/analysis-tab.js
yann300: ok

src/app/tabs/compile-tab.js
```diff
@@ -14,9 +14,9 @@ var remixLib = require('remix-lib')
...
+  var appEvents = events
```
yann300: same here, well that's fine to keep `appEvents`, or change every occurence of `appEvents` to `events`

src/app/tabs/run-tab.js
```diff
@@ -20,14 +20,17 @@ var noInstancesText = yo`<div class="${css.noInstancesText}">0 contract Instance
 
 var pendingTxsText = yo`<span></span>`
 
-function runTab (container, appAPI, appEvents) {
-  var events = new EventManager()
+function runTab (api = {}, events = {}, opts = {}) {
```
yann300: here too, don't update the name or change every occurence

src/app/tabs/compile-tab.js
serapath: of course - i will make pull requests to refactor all the tabs to have them as proper prototype functions of the class that the `compile-tab` will then be.

This is just a quick hack to have minimum change to standardize the API - it's not meant to stay like that.

I did it in order to make small pull requests.

---

If I change it here, then maybe i should change it for all the tabs, which will probably add 3-5 or more lines of code to each of them ... :/

**so do you think it would be ok to leave it like that - because it's working - and then fix that with the next upcoming PRs?**

yann300: ah I said above that this is fine, so we can merge this as long as you'll make another PR to make proper `render` function

src/app/tabs/run-tab.js
serapath: the idea is to change every module signature to `(api = {}, events = {}, opts = {})` according to your PR to update best practices.
Personally I think it matters that the arguments are named the same in every module, so that people get used to rely on `api`, `events` and `opts` and not have different name in each module.

I see this PR as an opportunity to come closer to that goal

src/app/tabs/compile-tab.js
serapath: the idea is to change every module signature to `(api = {}, events = {}, opts = {})` according to your PR to update best practices.
Personally I think it matters that the arguments are named the same in every module, so that people get used to rely on `api`, `events` and `opts` and not have different name in each module.

I see this PR as an opportunity to come closer to that goal

src/app/panels/righthand-panel.js
serapath: the idea is to change every module signature to `(api = {}, events = {}, opts = {})` according to your PR to update best practices.
Personally I think it matters that the arguments are named the same in every module, so that people get used to rely on `api`, `events` and `opts` and not have different name in each module.

I see this PR as an opportunity to come closer to that goal

src/app.js
serapath: i would prefer not to.
The previous line of code triggered inside of `app.js` an event that bubbles down to children components which i perceive as going against our best practices.

To me: `data is passed down` and `events bubble up`.
The only time an "event bubbles down" is when an event is triggered by a child or leaf component and the event listener is/was passed down by app.js to other child or leaf components to listen to it.

The idea behind that:
* A leaf/child component can listen to an event caused by another leaf/child component somwhere else

...but this is not the case here.

**So considering that, would you agree to leave this change as is maybe?**

src/app/tabs/compile-tab.js
serapath: ok

src/app/panels/righthand-panel.js
serapath: ok

yann300:
pushed some fixes

serapath:
checked the pushed fixes.
+1 :-) LGTM

PR #1090, opened 2018-02-15
move most tx code to remix-lib

iurimatias:


src/execution-context.js
```diff
@@ -1,227 +1,4 @@
...
-    } else {
-      web3.setProvider(oldProvider)
-      var alertMsg = 'Not possible to connect to the Web3 provider. '
-      alertMsg += 'Make sure the provider is running and a connection is open (via IPC or RPC).'
-      cb(alertMsg)
-    }
-  }
-  this.setProviderFromEndpoint = setProviderFromEndpoint
-}
+var executionContext = remixLib.execution.executionContext
```
yann300: we need to fix the code in `execution-context` before merging that.
that's a change I made last week to fix tape tests execution...

src/execution-context.js
```diff
@@ -1,227 +1,4 @@
...
 
-module.exports = new ExecutionContext()
+module.exports = executionContext
```
yann300: I kind of think it would be better to let the execution context in `remix-lib` to return a class, and we could instanciate it here.

iurimatias: I agree, however given the issues experienced with the PR, and the pending PRs, I rather do it in a different PR afterwards to avoid more issues

PR #1093, opened 2018-02-18
Add tx logs

ninabreznik:
Logs are added, they appear when Main/Ropsten/Rinkeby or Kovan is chosen, but links seem to not work, so @yann300 pls check what should be changed. I tried many verisons from data.tx but nothing seem to work for etherscan/tx part.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -734,6 +734,9 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
       },
       compiledContracts: function () {
         return compiledContracts()
+      },
+      detectNetwork: (cb) => {
+        executionContext.detectNetwork(cb)
       }
     },
     events: {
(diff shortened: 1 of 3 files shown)


yann300:
hmm well the thing is that we need to log the these links just after having clicked on the action, not when the transaction is already mined.

ninabreznik:
@yann300 Hm... that will need some restructuring now. 
Can you tell me what exactly you had in mind. Or actually, what I would prefer is to tell me what problem we are solving :)

yann300:
right
When a transaction is sent, we display "something is pending" and the log come like 15-30 s after when the transaction is mined, would be preferable to:
 - display the link to etherscan just after the "something is pending", so users can click and see the tx pending in etherscan can too.
 - if it is a contract creation, display the link to the contract in etherscan too (well in that case it might be preferable to display the link together with the log in the terminal)

yann300:
I think we need to address this issue in a different way... ether scan link has to be display *before* the transaction remix log and not at the same time

PR #1094, opened 2018-02-19
Fix import external (github, swarm, ipfs)

yann300:


ninabreznik:
@yann300 I am trying to check the PR, but not sure what should I check. Can't see any changes. Where is this used?

ninabreznik:
Works fine when importing from Github.

But for Swarm I get this
screen shot 2018-02-20 at 2 44 15 pm

src/app.js
```diff
@@ -217,6 +217,21 @@ This instance of Remix you are visiting WILL NOT BE UPDATED.\n
 Please make a backup of your contracts and start using http://remix.ethereum.org`)
   }
 
+  function importExternal (url, cb) {
+    handleImports.import(url,
+      (loadingMsg) => {
+        $('#output').append($('<div/>').append($('<pre/>').text(loadingMsg)))
```
serapath: can't we do this without jQuery?
1. it would be nicer if we do not grab randomly a node from the dom (`#output`) but have some kind of event or api
2. then, the place that controls the `#output` node, should use regular dom methods `.appendChild(...)` rather than jQuery

ninabreznik: +1 for getting rid of jQuery :)

yann300: that lead to an interesting improvement.
replaced jquery by `tooltip` but then the tooltip would pop up at each compilation (basically for every key stroke), so added a cache so we don't display the popup if already imported.

yann300: and also we don't anymore make the request to retrieve the content at each compilation (that's the interesting improvement)

PR #1100, opened 2018-02-19
refactor txRunner

iurimatias:


src/app/execution/txRunner.js
```diff
@@ -43,150 +27,179 @@ TxRunner.prototype.rawRun = function (args, cb) {
...
+          // removed, but for now keeping the original logic
+          try {
+            var fee = executionContext.web3().toBigNumber(tx.gas).mul(executionContext.web3().toBigNumber(executionContext.web3().toWei(gasPrice.toString(10), 'gwei')))
+            txFeeText = ' ' + executionContext.web3().fromWei(fee.toString(10), 'ether') + ' Ether'
+            priceStatus = true
+          } catch (e) {
+            txFeeText = ' Please fix this issue before sending any transaction. ' + e.message
+            priceStatus = false
+          }
+          cb(txFeeText, priceStatus)
```
yann300: isn't `(priceStatus, txFeeText)` ?

src/app/execution/txRunner.js
```diff
@@ -43,150 +27,179 @@ TxRunner.prototype.rawRun = function (args, cb) {
...
+        },
+        (cb) => {
+          executionContext.web3().eth.getGasPrice((error, gasPrice) => {
+            var warnMessage = ' Please fix this issue before sending any transaction. '
+            if (error) {
+              return cb('Unable to retrieve the current network gas price.' + warnMessage + error)
+            }
+            try {
+              var gasPriceValue = executionContext.web3().fromWei(gasPrice.toString(10), 'gwei')
+              cb(null, gasPriceValue)
```
yann300: seems the awaited callback is more something like `cb(txFeeText, gasPriceValue, gasPriceStatus)`

src/app/execution/txRunner.js
iurimatias: nice catch!

src/app/execution/txRunner.js
iurimatias: correct, so this call is valid, I prefer to avoid passing another null as a 3rd parameter if it's not necessary (much like for a typicall callback cb(err, result) when there is an error only cb(err) is necessary instead of cb(err, null) )

PR #1101, opened 2018-02-19
validate address

### LINKED ISSUE #210: 'At address' function should verify that the input is an address

szdavid92:
Resolves #210.

Show alert on invalid address.


axic:
Seems to be a duplicate of #1025 ?

yann300:
duplicate has been merged

PR #1105, opened 2018-02-22
extract UI code from txRunner

iurimatias:


src/app/execution/txRunner.js
```diff
@@ -142,62 +138,16 @@ TxRunner.prototype.runInNode = function (from, to, data, value, gasLimit, useCal
...
-            }
-            var gasPrice = executionContext.web3().toWei(content.querySelector('#gasprice').value, 'gwei')
-            executeTx(tx, gasPrice, self._api, callback)
-          }}, {
-            label: 'Cancel',
-            fn: () => {
-              return callback('Transaction canceled by user.')
-            }
-          })
+      confirmCb(network, tx, gasEstimation, callback, (gasPrice) => {
```
yann300: not really sure about that, but seems `confirmCb` is declared as `(network, tx, continueTxExecution, gasEstimation, cancelCb)` in `universal dapp`

yann300: unfortunately test pass :( that still because we test only against the javascript VM)

iurimatias: fixed params in the universal-dapp

PR #1123, opened 2018-02-28
Update README.md

morozovsk:
Changed an order of an installation.


yann300:
you had this issue cause we merged new code in remix repo that was'nt not still handled in remix-ide. that should be fixed now

PR #1126, opened 2018-03-01
Fix alert msg in remix-alpha

yann300:


src/app.js
```diff
@@ -210,8 +210,8 @@ function run () {
 
   if (window.location.hostname === 'yann300.github.io') {
     modalDialogCustom.alert('This UNSTABLE ALPHA branch of Remix has been moved to http://ethereum.github.io/remix-live-alpha.')
-  } else if (window.location.hostname === 'ethereum.github.io' &&
-  window.location.pathname.indexOf('/remix-live-alpha') === 0) {
+  } else if (window.location.hostname === 'remix-alpha.ethereum.org' ||
```
axic: I think you should keep both URLs.

yann300: I do remix-alpha is tested and ethereum.github.io too.
not sure I get the comment

axic: It may have been too early when I've checked, yes it is below :)

PR #1145, opened 2018-03-13
Warn if using https

yann300:


chriseth:
We could also include instructions about how to bypass the restriction on firefox while still using https (click on the lock icon on the left of the url bar, etc).

PR #1161, opened 2018-03-19
add Dockerfile

jsoizo:
This PR is aim to running IDE on docker container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ADD ./ /work/
 
 WORKDIR /work
 
-RUN npm install && npm run downloadsolc && npm run make-mock-compiler && npm run build
+RUN npm install && npm run downloadsolc_root && npm run build
 
 RUN tar -cvf dist_files.tar assets background.js build icon.png index.html manifest.json README.md soljson.js
 


Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM node as build-env
+
+RUN mkdir /work
+
+ADD ./ /work/
+
+WORKDIR /work
+
+RUN npm install && npm run downloadsolc && npm run make-mock-compiler && npm run build
```
jsoizo: Is `npm run downloadsolc && npm run make-mock-compiler` necessary? If not, I will delete it.

yann300: `npm run make-mock-compiler` is not necessary

yann300: and I think that `npm run downloadsolc_root` should be used instead of `npm run downloadsolc` . https://github.com/ethereum/remix-ide/blob/master/package.json#L148

yann300:
yep https://hub.docker.com/r/4c0n/remix-ide/

jsoizo:
@yann300 Sorry for missing your reply, If I fix it as you wrote, would you re-open this PR?

PR #1162, opened 2018-03-19
Show pending modal for long-lasting transactions

### LINKED ISSUE #1103: Popup an alert after 10 unsucessful attempt to retrieve the tx receipt
here : https://github.com/ethereum/browser-solidity/blob/master/src/app/execution/txRunner.js#L194

bitpshr:
This PR adds logic to show a modal that informs the user that a transaction is still pending if it takes longer than 10 seconds.

modal

Resolves #1103

### DIFF
--- a/src/app/execution/txRunner.js
+++ b/src/app/execution/txRunner.js
@@ -1,9 +1,11 @@
 'use strict'
+var yo = require('yo-yo')
 var EthJSTX = require('ethereumjs-tx')
 var EthJSBlock = require('ethereumjs-block')
 var ethJSUtil = require('ethereumjs-util')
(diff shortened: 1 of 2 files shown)


yann300:
thanks :)  could you rebase please?
currently that popup a modal, perhaps would be better to just add a notification (less intrusive)  ( https://github.com/ethereum/remix-ide/blob/938d03b66f04ab9777fac3265c3b988672563458/src/app/files/NotPersistedExplorer.js#L3 )
@ninabreznik any thought about that?

bitpshr:
I noticed the `TxRunner` moved into the `remix` package and out of `remix-ide`. I'll work up a new PR.

Issue #1156, opened 2018-03-15
[Plugins] Defining messages structure


yann300:
# brainstorming iframe api used by plugins

```js
var tmpID, msg, sessions = {}


// RECEIVE
window.addEventListener('message', args => [].concat(args).forEach(msg => {
  var { id, status, type, key, value } = msg
  if (status === 'error') return console.error(value)
  /* "canonicalise" */ value = [].concat(value)
  if (IDS[id]) console.log('response', value)
  else if (id) console.log('request', type, key, value)
  else         console.log('notification', type, key, value)
}))


// SEND
function example (line1, line2, line3) {
  return sessions[tmpID = Math.random()] = msg = {
    key: 'editor', type: 'mark', id: tmpID, value: [line1, line2, line3]
  }
}
window.postMessage(example(...)) // sending message to highlight/mark 3 source code lines
var examples = [example(...), example(...), example(...)]
window.postMessage(examples) 
```

## `id`
identifies a session which usually consists only of a "request" ans "response" message.
* it allows to identify, that a "response" message answer to a certain "request" message that was sent.

## `status`
was roughly inspired by http status codes and is most interesting in "response" messages, but might also be interesting in other messages.
* The default normal successful message does not need a status or has a status of `undefined`
* if `status === 'error'` then something went wrong
* maybe, if a "response" or "notification" consists of a series of messages, the `status` could indicate,
  whether the last message has been received or if more are expected
* also the `status` could indicate if the permission for a "request" was denied, etc...

## `type`
is meant as `action/method/opcode/eventName/command/....`

## `key`
is meant as `address/origin/target/namespace/...`

## `value`
can be one or many (=array) and is meant as `arguments/data/payload/...`

yann300:
just need to add a property that defines which type of message this is

yann300:
- [ ] Update the current PluginManager.js to match the current spec.
 - [ ] Inject an object in plugin iframe that will abstract the underlying calls.
```
API consumers will then register to notifications:
windows.remix.notification.compiler.compilationFinished.register(function () {})
windows.remix.notification.tab.focus.register(function () {})
and make request
windows.remix.request.editor.setConfig('path', 'value', function () {})
windows.remix.request.editor.getConfig('path', function () {})
```

- [ ] Write tests in browser-test for the injected module

serapath:
[...]

PR #1196, opened 2018-03-26
Update plugin API

yann300:
following https://github.com/ethereum/remix-ide/issues/1156


src/app/plugin/pluginManager.js
```diff
@@ -61,37 +101,47 @@ class PluginManager {
...
         var data = JSON.parse(event.data)
-        data.arguments.unshift(this.inFocus)
+        data.value.unshift(this.inFocus)
         if (allowedapi[data.type]) {
-          data.arguments.push((error, result) => {
-            response(data.type, data.id, error, result)
+          data.value.push((error, result) => {
+            response(data.key, data.type, data.id, error, result)
           })
           api[data.key][data.type].apply({}, data.arguments)
```
Fi3: `data.arguments` should be `data.value`

test-browser/plugin/remix.js
```diff
@@ -8,25 +8,31 @@ window.addEventListener('message', receiveMessage, false)
 window.onload = function () {
   document.querySelector('input#testmessageadd').addEventListener('click', function () {
     window.parent.postMessage(JSON.stringify({
+      action: 'request',
+      key: 'editor',
```
Fi3: `key: 'editor'` should be `key: 'config'`

test-browser/plugin/remix.js
```diff
@@ -8,25 +8,31 @@ window.addEventListener('message', receiveMessage, false)
...
+      value: [document.getElementById('filename').value, document.getElementById('valuetosend').value],
       id: 34
-    }), 'http://127.0.0.1:8080')
+    }), '*')
   })
 
   document.querySelector('input#testmessageremove').addEventListener('click', function () {
     window.parent.postMessage(JSON.stringify({
+      action: 'request',
+      key: 'editor',
```
Fi3: `key: 'editor'` should be `key: 'config'`

test-browser/plugin/remix.js
```diff
@@ -8,25 +8,31 @@ window.addEventListener('message', receiveMessage, false)
...
+      value: [document.getElementById('filename').value],
       id: 35
-    }), 'http://127.0.0.1:8080')
+    }), '*')
   })
 
   document.querySelector('input#testmessagerget').addEventListener('click', function () {
     window.parent.postMessage(JSON.stringify({
+      action: 'request',
+      key: 'editor',
```
Fi3: `key: 'editor'` should be `key: 'config'`

src/app/plugin/pluginManager.js
```diff
@@ -1,13 +1,47 @@
...
+ *
+ * - Response sent by Remix and receive by the plugin:
+ *{
+ *  id: <number>,
+ *  action: 'response',
+ *  key: <string>,
+ *  type: <string>,
+ *  value: <array>,
+ *  error: (see below)
+ *}
```
serapath: I just checked again: https://github.com/ethereum/remix-ide/issues/1156

* The original proposal is `var { id, status, type, key, value } = msg`
* You said you would like to add the `action` field to indicate "response/notification/request".
* That would make it `var { action, id, status, type, key, value } = msg`
* But now the `status` is/was renamed to `error` i guess.

Just to make sure how i have to think about it :-)
So i guess we now have: `var { action, id, error: status, type, key, value } = msg`

I would say that is fine.

serapath: The above comment is just for me to clarify.
If this is how it's meant, i would say the PR looks good to me :-)

yann300: yes except that error is undefined if there's no error

PR #1202, opened 2018-03-27
Show one context menu at a time

### LINKED ISSUE #1092: Right clicking in the file manager causes multiple menu containers to open
Issue:
When a user right clicks on a file in the file browser a menu container appears. When the user right clicks again another container appears on top of the first. This is shown in the screen shot below:
image

Enhancement: 
When a user right clicks on a file name menu  containers in the background should be closed. This would clean up the UI.

bitpshr:
This PR fixes a UX issue where multiple right-click context menus were showing at a given time. `contextMenu` was updated to return a handle that exposes its `hide` function, and `file-explorer` was updated to cache a singleton reference to the last-opened context menu. This handle is used to hide any existing context menu before opening a new one.

Resolves #1092

### DIFF
--- a/src/app/files/file-explorer.js
+++ b/src/app/files/file-explorer.js
@@ -10,6 +10,8 @@ var helper = require('../../lib/helper')
 
 var css = require('./styles/file-explorer-styles')
 
+let MENU_HANDLE
+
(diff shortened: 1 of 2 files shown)


yann300:
thanks for this!

PR #1212, opened 2018-03-29
Multi params component

yann300:

### DIFF
--- a/src/universal-dapp-styles.js
+++ b/src/universal-dapp-styles.js
@@ -19,6 +19,7 @@ var css = csjs`
     word-break: break-word;
     line-height: initial;
     overflow: visible;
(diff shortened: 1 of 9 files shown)


ryestew:
multi params are working in instances as long as strings are quoted

src/universal-dapp-styles.js
```diff
@@ -55,13 +58,25 @@ var css = csjs`
...
+    clear: both;
+  }
   .buttonsContainer {
     margin-top: 2%;
     display: flex;
     overflow: hidden;
   }
   .contractActions {
-    display: flex;
+    // display: flex;
```
yann300: should we remove this ?

ryestew: yes we should

PR #1219, opened 2018-04-09
Run tab

ninabreznik:
Current view (all features working)
screen shot 2018-04-09 at 1 54 00 am


ninabreznik:
screen shot 2018-04-16 at 2 43 48 am
screen shot 2018-04-16 at 2 43 38 am
screen shot 2018-04-16 at 2 43 26 am
screen shot 2018-04-16 at 2 43 22 am
screen shot 2018-04-16 at 2 43 16 am

src/universal-dapp-ui.js
```diff
@@ -31,31 +34,34 @@ UniversalDAppUI.prototype.renderInstance = function (contract, address, contract
...
+    <div class="${css.title}" onclick=${toggleClass}>
+      <div class="${css.titleText}"> ${contractName} at ${shortAddress} (${context}) </div>
+      ${copyToClipboard(() => address)}
+    </div>`
 
+  instance.appendChild(title)
   if (self.udapp.removable_instances) {
     var close = yo`<div class="${css.udappClose}" onclick=${remove}><i class="${css.closeIcon} fa fa-close" aria-hidden="true"></i></div>`
-    instance.append(close)
+    title.append(close)
```
yann300: A fix has been pushed on `master` , that should be `title.appendChild(close)`

src/recorder.js
```diff
@@ -131,6 +130,7 @@ class Recorder {
   append (timestamp, record) {
     var self = this
     self.data.journal.push({ timestamp, record })
+    self.event.trigger('showtxRecordedCount', [self.data.journal.length])
```
yann300: perhaps the name of the event should more general, like `newTxRecorded`, and parameters would be `[timestamp, record, self.data.journal.length]` ?

src/universal-dapp-ui.js
ninabreznik: Thanks, fixed now!

src/recorder.js
ninabreznik: Ok

src/app/panels/terminal.js
```diff
@@ -105,23 +105,40 @@ class Terminal {
...
           ${self._view.icon}
           <div class=${css.clear} onclick=${clear}>
-          <i class="fa fa-ban" aria-hidden="true" onmouseenter=${hover} onmouseleave=${hover}></i>
+            <i class="fa fa-ban" aria-hidden="true" title="Clear console"
+            onmouseenter=${hover} onmouseleave=${hover}></i>
+          </div>
+          ${self._view.pendingTxCount}
+          <div class=${css.verticalLine}></div>
+          <div class=${css.listen}>
+            <input onchange=${listenOnNetwork} type="checkbox"
```
yann300: removing `listen to network` was intentionally?

src/app/panels/terminal.js
```diff
@@ -105,23 +105,40 @@ class Terminal {
         ${self._view.input}
       </div>
     `
-    self._view.icon = yo`<i onmouseenter=${hover} onmouseleave=${hover} onmousedown=${minimize} class="${css.toggleTerminal} fa fa-angle-double-down"></i>`
-    self._view.dragbar = yo`<div onmousedown=${mousedown} class=${css.dragbarHorizontal}></div>`
+    self._view.icon = yo`
+      <i onmouseenter=${hover} onmouseleave=${hover} onmousedown=${minimize}
```
yann300: we might miss a title for the new `pending transaction` div, or we could display the entire `x pending transactions` ....

yann300:
seems the recorder does not work (while replaying transaction). could you give it a try?

src/app/panels/terminal.js
ninabreznik: Yes, I removed the text, but left the checkbox for now. I think in the next round, when I do the Terminal, I will then see how to improve this, but I think maybe for now this can stay only as a checkbox. Later I would maybe rather include it in the dropdown/filter.

src/app/panels/terminal.js
yann300: shall we add this label here to clarify what the counter means?

ninabreznik: Absolutely. It must have disappeared during rebasing.

src/universal-dapp-ui.js
```diff
@@ -54,14 +54,7 @@ UniversalDAppUI.prototype.renderInstanceFromABI = function (contractABI, address
...
-    var instanceContainer = document.querySelector('[class^="instanceContainer"]')
-    if (instanceContainer.children.length === 1) {
-      var noInstancesText = yo`
-        <div class="${css.noInstancesText}">
-          Currently you have no contract instances to interact with.
-        </div>`
-      instanceContainer.appendChild(noInstancesText)
-    }
+    // @TODO perhaps add a callack here to warn the caller that the instance has been removed
   }
```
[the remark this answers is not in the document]
yann300: to reproduce try loading a contract using the `At address` action. the label "You don't have ..." is not hiddenn

src/app/tabs/run-tab.js
```diff
@@ -225,13 +225,19 @@ function makeRecorder (appAPI, appEvents, opts, self) {
             modalDialogCustom.alert('Failed to create file ' + newFile)
           } else {
             appAPI.switchFile(newFile)
+            self._view.recorderCount.innerText = 0
```
yann300: you are putting the record count to 0 but are you sure that the actually records is reset?

yann300: commit https://github.com/ethereum/remix-ide/pull/1219/commits/41790513f833ebbe7e90f77fc127f2ae9ac77abf

PR #1223, opened 2018-04-09
Create release-process.md

yann300:


serapath:
LGTM

ninabreznik:
@yann300 LGTM too, but of course I have no experience with it so far, so let's try and then we can see if we should change/improve anything

PR #1228, opened 2018-04-10
refactor tabbed menu

serapath:


src/app/tabs/tabbed-menu.js
```diff
@@ -1,52 +1,97 @@
...
-      self.selectTab(tabView.querySelector(`li.${tabName}`))
-    })
+module.exports = class TabbedMenu {
+  constructor (api = {}, events = {}, opts = {}) {
+    const self = this
+    self.event = new EventManager()
+    self._api = api
+    self._events = events
+    self._view = { el: null, viewport: null, tabs: {}, contents: {} }
+    self.data = {}
```
yann300: seems you don't use `data` neither `_components` . perhaps not necessary to declare them

src/app/tabs/tabbed-menu.js
```diff
@@ -1,52 +1,97 @@
...
-    this.events = events
-    this.tabs = {}
-    this.contents = {}
-
-    events.app.register('debuggingRequested', () => {
-      self.selectTab(tabView.querySelector('li.debugView'))
-    })
+var EventManager = remixLib.EventManager
 
-    events.rhp.register('switchTab', tabName => {
```
yann300: events registration has to be kept. that might be the reason why tests are failing

src/app/tabs/tabbed-menu.js
```diff
@@ -1,52 +1,97 @@
...
+  renderViewport () {
+    const self = this
+    if (self._view.viewport) return self._view.viewport
+    self._view.viewport = yo`
+      <div id="optionViews" class=${css.optionViews}>
+        ${Object.values(self._view.contents)}
+      </div>`
+    return self._view.viewport
+  }
+  addTab (title, cssClass, content) {
```
yann300: `helper.checkSpecialChars(title)` should be kept here too ?

src/app/tabs/tabbed-menu.js
```diff
@@ -1,52 +1,97 @@
...
+    if (self._view.el) self._view.el.appendChild(self._view.tabs[title])
+    if (self._view.viewport) self._view.viewport.appendChild(self._view.contents[title])
   }
-  render () { return this.tabView }
   selectTabByTitle (title) {
-    this.selectTab(this.tabs[title])
+    const self = this
+    self.selectTab(self._view.tabs[title])
+  }
+  switchTab (tabClass) {
```
yann300: should we use the same name pattern as others ? that would be `selectTabByClassName`

src/app/tabs/tabbed-menu.js
```diff
@@ -1,52 +1,97 @@
 var yo = require('yo-yo')
-var helper = require('../../lib/helper')
+var csjs = require('csjs-inject')
+var remixLib = require('remix-lib')
 
-var css = require('./styles/tabbed-menu-styles')
```
yann300: yeah putting the css in the same file make sense that's a tiny one, then `tabbed-menu-styles` should be removed

src/app/tabs/tabbed-menu.js
serapath: oh i see.

I think it's not completely clear to me when we use events and when we use methdod calls.

My understanding is, that events are fired, when we don't know who is interested in them.
Then components who want to know, will register for a certain event.

When a "switchTab" is triggered, i don't see it as a notification of RHP to the rest of the world,
but I see it as a request - not to the world, but actually only to the tabbed-menu specifically.
The RHP even created and has access to the tabbed-menu directly, so I feel a method should be called instead of doing it through events.

I think I would prefer the RHP to just call `tabbedMenu.switchTab(tabName)`

yann300: ok I see . yes it make sense
so in that case you should update this part https://github.com/ethereum/remix-ide/blob/1c0472f0e806b39fdebbcf7aa619b670ec159ff8/src/app/panels/righthand-panel.js#L41

src/app/tabs/tabbed-menu.js
serapath: added it back in

src/app/tabs/tabbed-menu.js
```diff
@@ -1,52 +1,98 @@
...
+module.exports = class TabbedMenu {
   constructor (api = {}, events = {}, opts = {}) {
-    var self = this
-    var tabView = document.createElement('ul')
-    this.tabView = tabView
-    this.events = events
-    this.tabs = {}
-    this.contents = {}
-
-    events.app.register('debuggingRequested', () => {
```
yann300: I think tests fails because of that being removed now

serapath: oh, i will check - not sure what happened here

PR #1229, opened 2018-04-10
add test-tab (wip)

serapath:


src/app.js
```diff
@@ -710,6 +710,10 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
     getSource: (fileName) => {
       return compiler.getSource(fileName)
     },
+    getAllSources: () => {
+      window.compiler = compiler
```
yann300: is that still needed?

PR #1234, opened 2018-04-10
refactor plugin tab

serapath:


src/app/plugin/pluginManager.js
```diff
@@ -37,42 +37,39 @@
...
+        self.post(tabName, JSON.stringify({
           type: 'focus'
         }))
-        this.inFocus = tabName
-        this.post(tabName, JSON.stringify({
+        self.inFocus = tabName
+        self.post(tabName, JSON.stringify({
           type: 'compilationData',
-          value: api.compiler.getCompilationResult()
+          value: self._api.getCompilationResult()
```
yann300: `self._api.compiler.getCompilationResult()` ?

src/app/tabs/plugin-tab.js
```diff
@@ -1,12 +1,37 @@
...
 
-module.exports = plugintab
+module.exports = class plugintab {
+  constructor (api = {}, events = {}, opts = {}) {
+    const self = this
+    self.event = new EventManager()
+    self._api = api
+    self._events = events
+    self._view = { el: null }
+    self.data = { url: opts.url }
```
yann300: why not directly `self._opts = opts`, like for the other param or maybe `self._url = opts.url` ? 
The code is totally fine, i would not object it ...  but i am not sure I would add this `.data` rule as a best practices if you ever want it.

src/app/plugin/pluginManager.js
serapath: oh sry - maybe a mistake. I'll check. It should probably be `self._api.compiler.getCompilationResult()` - but beside that, isn't our policy to more or less try to only make available what is needed - which would mean we should not maybe make the whole `compiler` available.

src/app/tabs/plugin-tab.js
serapath: yes i agree - the `.data` was basically some kind of idea for a convention before we started to do `._opts` ...so i agree and will change that :-)

src/app/plugin/pluginManager.js
yann300: we 'll need to discuss about that all together. but sometimes these `api` object got really huge and messy (like for `rhpAPI`), in that case, I think it is better pass the object directly

serapath: ok

PR #1235, opened 2018-04-10
refactor run tab (wip)

serapath:


src/app/tabs/run-tab.js
```diff
@@ -442,13 +447,196 @@ function settings (container, appAPI, appEvents, opts) {
...
     if (!lookupOnly) el.querySelector('#value').value = '0'
-    updateAccountBalances(container, appAPI)
+    updateAccountBalances(container, _api)
   })
 
   return el
 }
 
-module.exports = runTab
+const css = csjs`
```
yann300: I think it is better to keep this style out of  the *.js, it is kinda huge

serapath: ### me
1. when i personally work on UI it saves me time to:
  * either jump "back and forth" in the same file to quickly change styling by "finding" the selected class name definition (in atom with CTRL+F)
  * or if i need, by opening the same file twice in different tabs scrolled to different positions
2. somebody who doesnt care about styling is not bothered, because it's at the end of the file
3. if i work on 2 or 3 or more files in parallel, i need to open 4 or 6 or 8 and more file tabs and get confused which file is which - so usually just having everything in one place makes me faster.

### rob
rob mentioned once that he likes having stuff in one place after he tried compared to before

### nina
nina actually recommended and lobbied for it, because she finds it easier too

---
I personally think, the people doing most UI work should be comfortable with the way and that seems to be rob and nina. I also do UI work - but the person who actually refactored the styling out of the files (which is iuri) doesn't even work on the repo usually but has his own repo and otherwise is doing UI-unrelated stuff.

**isn't a good solution to put the `css` at the end of the file to be out of the way for those who need to look at the file and don't want to be bothered?**

ninabreznik: 👍 I definitely need css back in the files. I think we already agreed on this one. @iurimatias said he thought this was a good idea, but we never discussed this before css was moved away.

yann300: ok if you all prefer it, that's fine :)

yann300:
Closing this one for now, a lot of changes has been been already pushed to the `run tab` since then.

PR #1238, opened 2018-04-10
refactor analysis tab

serapath:


src/app/tabs/analysis-tab.js
```diff
@@ -1,12 +1,32 @@
...
+  }
+  render () {
+    const self = this
+    if (self._view.el) return self._view.el
+    self._view.el = yo`
+      <div class="${css.analysisTabView} "id="staticanalysisView"></div>`
+    return self._view.el
+  }
+}
+const css = csjs`
```
yann300: seems some CSS has been lost `staticanalysisView`, `infoBox`, `textBox`

serapath: Yes, but i think i removed it because i checked that it wasn't in use or didn't make a difference, because another class was overwriting it or it simply put stuff on that was already inherited or a default. ...so basically instead of just copy/pasting, i was trying to refactor the styling too to get rid of redundant styling attributes

PR #1239, opened 2018-04-10
refactor compile tab

serapath:


src/app/tabs/compile-tab.js
```diff
@@ -317,3 +322,141 @@ function detailsHelpSection () {
 }
 
 module.exports = compileTab
+
+const css = csjs`
```
yann300: well I think it would really make sense to keep them separated

src/app/tabs/compile-tab.js
```diff
@@ -1,6 +1,6 @@
-/* global */
-var $ = require('jquery')
 var yo = require('yo-yo')
+var remixLib = require('remix-lib')
```
yann300: why do you had these dependencies ?

src/app/tabs/compile-tab.js
```diff
@@ -146,26 +148,30 @@ function compileTab (appAPI = {}, appEvents = {}, opts = {}) {
       var error = false
       if (data['error']) {
         error = true
-        opts.renderer.error(data['error'].formattedMessage, $(errorContainer), {type: data['error'].severity})
+        // opts.renderer.error(data['error'].formattedMessage, $(errorContainer), {type: data['error'].severity})
```
yann300: could you keep `opts.renderer.error`  for now? that is going to be part of the coding best practices we are trying to setup.

src/app/tabs/compile-tab.js
serapath: i will check - maybe refactoring made it unnecessary. i'll remove it.

src/app/tabs/compile-tab.js
serapath: really?
my understanding is
* `opts` carries settings/config/data
* `events` carries event emitters so a module can listen
* `api` carries methods so a module can call the "SDK" the app offers

Why would we put a `render` object that contains a method directly on `opts`?
Shouldn't it be at least `appAPI.renderer.error(...)` at least?

But what is this `renderer` object about? ...i think it's the first time i see it.

src/app/tabs/compile-tab.js
serapath: see https://github.com/ethereum/remix-ide/pull/1235#discussion_r180827235

yann300: ok

src/app/tabs/compile-tab.js
yann300: I meant `for now`. This way of having an API was a good idea first, but now these API contains (rhp API for instance) a messy mix of function and objects. So I think we should address that too.
So the current refactoring you are doing is good but I would update these api call later.

serapath: i see - so the `Recorder` needs access to the `compiler` and that's why it has been decided in `app.js` to put it on the `opts` object when passed to the RightHandPanel and from there it goes basically everywhere where `opts` is passed ...
...i think in the long run we might wanna re-think our general strategy.

Maybe this is how it needs to be, but i feel it's not very convenient all this passing around of stuff and wonder if there is a better approach.

----
so, YES,
i will put the `opts.renderer.error` back into place because i don't feel like refactoring even more.

yann300: ok, I think there were an (very old) issue that is meant to remove the `renderer` ... but never had a chance to actually do that

PR #1243, opened 2018-04-11
Remove uneeded dep && remove uneeded script

yann300:


yann300:
merging as it is only removing dep in package.json

PR #1251, opened 2018-04-12
Include git clone of ethereum/remix for dev instructions

axic:


ninabreznik:
Hey @axic, thanks for the PR. 

I think that this command already runs the clone, but maybe it would be nice to add in the parenthesis what exactly the command does. 

@yann300 ?

axic:
Oh sorry, I just realised I didn't do `pullremix`, only `linkremixlib` because that was the one I needed to change.

It could be more clear tbh.

ninabreznik:
I agree. If you wasn't sure, then it's obviously the problem we didn't communicate it clear enough. So thanks for the feedback. Will improve this one!

axic:
Created an issue and closing this then.

PR #1266, opened 2018-04-19
Fix debugging test (comply with 0.4.23 solidity version)

yann300:


yann300:
merging as this is just fix browser test to comply with new solidity version (0.4.23)

PR #1272, opened 2018-04-21
Add linkremixdebugger script to package.json

kcole16:
This script is run as part of the `setupremix` script, and was not included in the latest build.


yann300:
Thanks for the PR. I removed the call to `linkremixdebugger`
https://github.com/ethereum/remix-ide/pull/1273

PR #1278, opened 2018-04-25
Multi param angles

ryestew:


src/multiParamManager.js
```diff
@@ -77,13 +76,13 @@ class MultiParamManager {
       this.clickCallBack(this.funABI.inputs, ret)
     }
 
-    var button = yo`<button onclick=${() => { multiOnClick() }} class="${css.instanceButton}">Submit</button>`
+    var button = yo`<button onclick=${() => { multiOnClick() }} class="${css.instanceButton}">Call</button>`
```
yann300: Not sure it is a good idea to rename it `Call`, that could add confusion cause we also use the word `Call` when executing a function without changing the state (in remix blue action versus red action). maybe `Transact`?

ryestew: True!  Probably Transact or Snd Txn - I'll make it Transact for now.

PR #1307, opened 2018-05-14
Multi param angles content

ryestew:
Automatically wrap strings and numbers and 0x in double quotes from the multi inputs ( the expanded view) & copying over the content between the expanded and contracted views.

### DIFF
--- a/src/multiParamManager.js
+++ b/src/multiParamManager.js
@@ -59,13 +59,10 @@ class MultiParamManager {
   makeMultiVal () {
     var inputString = this.basicInputField.value
     var inputJSON = JSON.parse('[' + inputString + ']')
-    console.log('inputJSON is ' + inputJSON)
     var multiInputs = this.multiFields.querySelectorAll('input')
-    console.log('ml length ' + multiInputs.length)
     for (var k = 0; k < multiInputs.length; k++) {
       if (inputJSON[k]) {
-        multiInputs[k].value = inputJSON[k]
-        console.log('inputJSON number ' + k + ' is ' + inputJSON[k])
+        multiInputs[k].value = JSON.stringify(inputJSON[k])
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/multiParamManager.js
```diff
@@ -2,6 +2,7 @@
 
 var yo = require('yo-yo')
 var css = require('./universal-dapp-styles')
+// var helper = require('../remix/remix-lib/src/execution/txHelper')
```
yann300: we should perhaps remove the comment if this is not going to be readded in the future

src/multiParamManager.js
```diff
@@ -20,16 +21,50 @@ class MultiParamManager {
...
+  emptyInputs () {
+    var valArray = this.multiFields.querySelectorAll('input')
+    for (var k = 0; k < valArray.length; k++) {
+      valArray[k].value = ''
+    }
+    this.basicInputField.value = ''
+  }
+
+  makeMultiVal () {
+    var inputString = this.basicInputField.value
```
yann300: I think we might need to wrap with double quote here
```
 params = params.replace(/(^|,\s+|,)(\d+)(\s+,|,|$)/g, '$1"$2"$3') // replace non quoted number by quoted number
        params = params.replace(/(^|,\s+|,)(0[xX][0-9a-fA-F]+)(\s+,|,|$)/g, '$1"$2"$3') // replace non quoted hex string by quoted hex string
```
cause the JSON.parse will break if some values are not wrapped ...

src/multiParamManager.js
```diff
@@ -52,28 +87,31 @@ class MultiParamManager {
...
     this.contractActionsContainerSingle = yo`<div class="${css.contractActionsContainerSingle}" >
-      <button onclick=${() => { onClick() }} class="${css.instanceButton}">${title}</button>${basicInputField}<i class="fa fa-angle-down ${css.methCaret}" onclick=${() => { this.switchMethodViewOn() }} title=${title} ></i>
+      <button onclick=${() => { onClick() }} class="${css.instanceButton}">${title}</button>${this.basicInputField}<i class="fa fa-angle-down ${css.methCaret}" onclick=${() => { this.switchMethodViewOn() }} title=${title} ></i>
       </div>`
 
-    var multiFields = this.createMultiFields()
+    this.multiFields = this.createMultiFields()
     var multiOnClick = () => {
-      var valArray = multiFields.querySelectorAll('input')
+      var valArray = this.multiFields.querySelectorAll('input')
```
yann300: perhaps we can use `getMultiValsString` here

PR #1318, opened 2018-05-18
refactor compile tab

serapath:


yann300:
@serapath  the tab content is only displayed when we : click on the support tab, click another tab and reclick back to the support tab

PR #1325, opened 2018-05-25
Fixed delegation loop condition

TamaMcGlinn:
When the while clause "voters[to].delegate != msg.sender" breaks the loop, the instruction assigning that to "to" is not executed.

### DIFF
--- a/contracts/ballot.sol
+++ b/contracts/ballot.sol
@@ -84,7 +84,7 @@ contract Ballot {
         }
 
         // We found a loop in the delegation, not allowed.
-        if (to == msg.sender) {
+        if (voters[to].delegate == msg.sender) {
             throw;
         }
 


yann300:
Thanks for the PR, well I think the code was fine.
Have you got a bad behavior, could you describe step to reproduce the issue?

contracts/ballot.sol
```diff
@@ -84,7 +84,7 @@ contract Ballot {
         }
 
         // We found a loop in the delegation, not allowed.
-        if (to == msg.sender) {
+        if (voters[to].delegate == msg.sender) {
```
chriseth: This could perhaps be synchronized with https://solidity.readthedocs.io/en/latest/solidity-by-example.html#voting

TamaHobbit:
The problem is that the code does not throw when encountering a delegation loop. The 'solidity by example' linked by chriseth has indeed fixed this issue, in a slightly different way.

You will need to open the code as you read along this example:
voters['A'] = tom and voters['B'] = harry. harry.delegate = 0, tom.delegate = 0

If tom calls delegate('B'), i.e. delegate my vote to harry, the while loop is never entered, and the if(to == msg.sender) does not enter; so far so good. As a result, tom.delegate = 'B'.

Now harry calls delegate('A'). sender = harry - this should throw as it will be circular. But the forwarding while loop only sets to = the penultimate delegatee, and then break before actually assigning to = msg.sender.

voters[to] is initially tom. tom.delegate == 'B', which is msg.sender, so the while loop skips entirely, never reassigning to = tom.delegate. So afterwards, to see if we need to throw, we have to check in the if with the same condition as broke the loop; voters[to].delegate == msg.sender. If we check "to == msg.sender", that is false since to is 'A' but msg.sender is 'B'.

But even after my fix, it's really weird to check something in the while loop, then check it again before throwing. The linked code is just better; assign to as you go along, and throw as soon as the value becomes equal to the sender; there is no need to exit the while loop only to throw again immediately after.

TamaMcGlinn:
Closed because the maintainers don't case.

PR #1326, opened 2018-05-25
Adding copy to clipboard in the expanded view

ryestew:
This version has the copy icon - without much instruction about what its copying or that it is hex encoding the inputs.


ryestew:
This is what the icon will look like for encoding the inputs:
image

PR #1339, opened 2018-06-04
Override control-s and command-s to auto-recompile

greg-shyft:
### The problem this solves?

Remix is treated by many as their main text editor for solidity smart contracts. Therefore it should act as a conventional text editor. This PR acts as a hot-reloading tool by overriding command+save and control+save to auto-recompile the editor. Up until now this would cause the browser to attempt to download the remix ide website.

### What changed?
Only 8 lines in `./src/app.js` (line 238 - 245). It uses the window object to capture keypresses for those specific keys to trigger the recompile function found in `app.js`.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -235,6 +235,14 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
     return 'Are you sure you want to leave?'
   }
 
+  // Run the compiler instead of trying to save the website
+  $(window).keydown(function (e) {
+    if ((e.metaKey || e.ctrlKey) && e.keyCode == 83) { /*ctrl+s or command+s*/
+      e.preventDefault();
+      runCompiler();
+    }
+  });
+
   function importExternal (url, cb) {
     self._components.compilerImport.import(url,
       (loadingMsg) => {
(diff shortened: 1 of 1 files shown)


ninabreznik:
Great idea, @greg-shyft. We will check it out as soon as possible. Did you maybe check why the tests are failing?

ninabreznik:
Cheers to Stu and Elizabeth at Chainsafe. Was visiting you 2 weeks ago :) @greg-shyft

greg-shyft:
@ninabreznik Thanks!  Elizabeth is beside me says hello! It's just a linter issue I used `==` instead of `===` fixing it now!

ninabreznik:
Yay. Awesome.

Btw,  I also always do control + save everywhere so this is really cool that it can actually trigger a compile. 

Have to check if @yann300 has any objections, but otherwise I think it can be merged very soon. Thanks a lot for this.

Have fun!

GregTheGreek:
@ninabreznik All tests passed!

yann300:
thanks for that PR @GregTheGreek ;)

PR #1349, opened 2018-06-07
Add ability to filter out compiler warnings

shyft-greg:
Hey guys!

I'm back with a UX feature...

### The problem this solves?
On compile remix renders a lot of errors that are typically not important, and from what I've gathered from around the community, commonly ignored. What happens is these yellow warnings hide the critical (red warnings). In order to combat this, I added a checkbox, identical to the `auto-compile` checkbox, that will remove the yellow warnings exposing only the red.

### What changed?
Only `compile-tab.js` changed, to be honest I'm not sure if this is the best way to go about the state within the module, but I store to `self._opts.config.set('hideWarnings', false)`. When the errors are rendered I added a bit of logic to filter yellow when checked and added a style.

~I'm wondering if perhaps there is a better place to set the inital state of the checkbox? I tried digging through `app.js` and `righthand-pannel.js` which use `compile-tab.js` but couldn't identify the right spot.~

### NOTE
There are a few tests ~5 that are failing due to them expecting to see a warning but none was generated. If this looks like something that you want to merge I'll be more than happy to fix those tests up!

EDIT 1:
Got the checkbox state to persist on page refresh.

Cheers!

### DIFF
--- a/src/app/tabs/compile-tab.js
+++ b/src/app/tabs/compile-tab.js
@@ -35,6 +35,7 @@ module.exports = class CompileTab {
       maxTime: 1000,
       timeout: 300
     }
+    self._opts.config.set('hideWarnings', false)
     self._events.editor.register('contentChanged', scheduleCompilation)
     self._events.editor.register('sessionSwitched', scheduleCompilation)
     function scheduleCompilation () {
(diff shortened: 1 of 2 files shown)


src/app/tabs/compile-tab.js
```diff
@@ -277,6 +294,11 @@ const css = csjs`
     display: flex;
     align-items: center;
   }
+  .hideWarningsContainer {
```
ryestew: This css should get moved to /universal-dapp-styles.js

noot:
any updates? :)

GregTheGreek:
@ryestew I make the necessary changes to the styling, is it good to go?

Issue #1292, opened 2018-04-30
terminal scroll bar can't be scrolled


serapath:
The element containing the ethereum background in the terminal is transparent but overlapping the scrollbar so it can't be scrolled on my computer.

A quick fix would be to add the css `pointer-events: none;` to that element, so users can click through it and use the scrollbar again.

dwardu:
I would like to work on this issue. Thanks.

PR #1354, opened 2018-06-10
Fixed issue #1292; terminal scrollbar may now be scrolled

dwardu:
* Removed div.terminal_bg whose purpose was to hold the background SVG, as this div was occluding the scrollbar, and instead set SVG as the background-image of div.terminal_container using CSS.
* As CSS does not have a background-opacity directive, manually decoded, edited and re-encoded the SVGs (light, dark) so that all \<g\>s are wrapped in a top-level \<g opacity="0.1"\>
* Improved the rules of how the background image resizes


serapath:
@dwardu awesome :-) thank you.

In theory, you can have a container be `position: relative;` and then a `div` inside to be `position: absolute; width:100%; height:100%; background-image: ...; opacity: 0.5; z-index:999; pointer-events: none;` ...which would also do the trick, but i totally am fine with your solution.

thank you :-)

dwardu:
Yes it’s true, but I thought I’d clear it up to have less things to go wrong :)

You’re very welcome! This is my first contribution to this project, and I enjoyed it :) Thank _you_.

Another nice solution would be to do it as you are suggesting, and then 1. place the `<svg>` inline, instead of base64-encoded data-URI, which would be nice for diffs, and then 2. simply style light/dark themes using CSS.

But maybe now since it’s working, and it’s not a big thing, we could leave it as is… what do you think?

—
P.S. I have some small ideas, e.g. I would like to add a simple rudimentary web-app manifest so that it would be possible to “Add to Home Screen” and open full-screen; it can be very useful to maximize the working space if you’re working on a tablet, for example. For something small like that, should I just make the change and create a pull request, or should I first open it as a “suggested improvement” somewhere? Also, is there a better place to ask a question like this, than in this thread? Thanks!

serapath:
to me the PR looks fine :-)

i like the "Add to Home Screen" feature.

PR #1363, opened 2018-06-13
Added copy buttons for ABI and Bytecode

### LINKED ISSUE #1249: Copy contract ABI not in an obvious place
Now users have to go to compile tab, click on details, and scroll to the ABI.
Perhaps we should put it in a more obvious place.

tschoffelen:
Fixes #1249.

image


tschoffelen:
Thanks for approving @ryestew! Any idea when this will be merged?

PR #1369, opened 2018-06-18
Tests logic

ninabreznik:
screenshot_2018-06-18_16-36-50
screenshot_2018-06-18_16-36-32

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -474,6 +474,7 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
   self._components.filePanel = new FilePanel()
   self._view.leftpanel.appendChild(self._components.filePanel.render())
   self._components.filePanel.event.register('resize', delta => self._adjustLayout('left', delta))
+  registry.put({api: self._components.filePanel, name: 'filepanel'})
 
   // ----------------- Renderer -----------------
   var renderer = new Renderer()
--- a/src/app/files/fileManager.js
+++ b/src/app/files/fileManager.js
@@ -154,6 +154,7 @@ class FileManager {
   }
 
   fileProviderOf (file) {
+    if (!file) return null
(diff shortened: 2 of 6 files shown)


yann300:
Insight for the first implementation of the test tab:

The panel contains:

- List of checkboxes + labels:
    This list all the files ending with `*_test.sol` from the current path
- Action to start test:
    Selected `*_test.sol` files have their own tests executed.
- Panel showing test results:
    
    
    sample (that already how the test tab currently show results) :
    ```
    test1
        ✓ Check winning proposal
        ✓ Check winnin proposal with return value
        
    2 passing (1s)
    
    
    test2
        ✘ Check winning proposal
        ✓ Check winnin proposal with return value
    test3
        ✓ Check winning proposal
        ✓ Check winnin proposal with return value
    
    3 passing (2s)
    
    1 failing
    1) test2 Check winning proposal
    error: 1 should be the winning proposal
    ```


test1, test2 and test3 are solidity contracts implementing tests, they can be located in any `*_test.sol`.
There can be multiple contracts per file.


_______________________________________________________
1) should it run all tests it can find in any folder?
No, it run test from the current folder

3) should it run all tests it can find in the same path as the currently displayed file in the editor (browser, gist, github, ...)?
Yes

5) should it do (2.) but only for browser and remixd?
it does 2) for all the files / explorer

7) something else?
--


And how should that change when we make things selectable (when we add checkboxes to select which tests we want to run)?

 1) should it only list the options available in either above (1., 2. 3. or 4.?)
It should list all `*_test.sol` files from the current path.
by default all the files are checked.

 3) should it be able to display all options and you could run a selection of test files from different root folders (e.g. browser, gist and github)?
It should just display the `*_test.sol` from the current folder.

 5) something else?
 --
 
 _______________________________________________________

PR #1370, opened 2018-06-19
add warning if hash collision in function signature

### LINKED ISSUE #1360: Warn if hash collision in function signature
Upon compilation complete, 
use https://www.4byte.directory/docs/ to resolve function signature, display a warning if there is collision (more than 1 result for a single signature hash)

jvmaia:
closes #1360

### DIFF
--- a/src/app/contract/contractParser.js
+++ b/src/app/contract/contractParser.js
@@ -46,10 +46,14 @@ async function getDetails (contractName, contract, source) {
     detail['Assembly'] = formatAssemblyText(contract.evm.legacyAssembly, '', source.content)
   }
 
-  return verifySignatureCollisions(contract.evm.methodIdentifiers).then((collisions) => {
-    detail.collisions = collisions
-    return detail
-  })
+  return verifySignatureCollisions(contract.evm.methodIdentifiers)
+    .then((collisions) => {
+      detail.collisions = collisions
+      return detail
+    })
+    .catch((err) => {
+      console.log(err)
+    })
 }
 
 var retrieveMetadataHash = function (bytecode) {
--- a/src/app/contract/contractParser.js
+++ b/src/app/contract/contractParser.js
@@ -2,6 +2,7 @@
 
 var remixLib = require('remix-lib')
 var txHelper = remixLib.execution.txHelper
+var request = require('xhr-request')
 
 module.exports = (contractName, contract, compiledSource) => {
   return getDetails(contractName, contract, compiledSource)
--- a/src/app/contract/contractParser.js
+++ b/src/app/contract/contractParser.js
@@ -45,6 +45,29 @@ var getDetails = function (contractName, contract, source) {
     detail['Assembly'] = formatAssemblyText(contract.evm.legacyAssembly, '', source.content)
   }
 
+  detail.collisions = []
+  for (let func in contract.evm.methodIdentifiers) {
+    var signature = contract.evm.methodIdentifiers[func]
+    var url = 'https://raw.githubusercontent.com/ethereum-lists/4bytes/master/signatures/' + signature
+    request(url, function (err, data, response) {
+      if (err) {
+        console.log(err)
+      } else {
+        var displayWarn = false
+        if (response.statusCode === 200) {
+          if (data.split(';').length > 1) {
+            displayWarn = true
+          } else if (data !== func) {
+            displayWarn = true
+          }
+        }
+        if (displayWarn) {
+          detail.collisions.push(func)
+        }
+      }
+    })
+  }
+
   return detail
 }
 
--- a/src/app/contract/contractParser.js
+++ b/src/app/contract/contractParser.js
@@ -2,13 +2,13 @@
 
 var remixLib = require('remix-lib')
 var txHelper = remixLib.execution.txHelper
-var request = require('xhr-request')
+var verifySignatureCollisions = require('./verifySignatureCollisions')
 
 module.exports = (contractName, contract, compiledSource) => {
   return getDetails(contractName, contract, compiledSource)
 }
 
-var getDetails = function (contractName, contract, source) {
+async function getDetails (contractName, contract, source) {
   var detail = {}
   detail.name = contractName
   detail.metadata = contract.metadata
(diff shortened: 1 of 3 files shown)


ligi:
I see 2 things missing: 
 * there should be a warning message describing the user what the problem is
 * there must be a check if the local function text signature is the same as the remote one - sorry forgot to mention this when commenting in the original issue. Otherwise you would e.g. get warnings when you implement ERC-20 contracts as the functions are actually in the database. The warning should only come when you have a different text signature - but the same 4byte

src/app/tabs/compile-tab.js
```diff
@@ -95,12 +96,36 @@ module.exports = class CompileTab {
           self.data.contractsDetails[contract.name] = parseContracts(contract.name, contract.object, self._opts.compiler.getSource(contract.file))
           var contractName = yo`<option>${contract.name}</option>`
           self._view.contractNames.appendChild(contractName)
+          for (let func in contract.object.evm.methodIdentifiers) {
+            var signature = contract.object.evm.methodIdentifiers[func]
+            var url = 'https://raw.githubusercontent.com/ethereum-lists/4bytes/master/signatures/' + signature
```
yann300: @ligi  correct me if im wrong, but i think the endpoint should be 
https://www.4byte.directory/api/v1/signatures/?hex_signature=94f6e8ed

the result is something like
```
{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 80538,
            "created_at": "2018-06-28T01:05:14.381003Z",
            "text_signature": "initBallotProxy(uint8,bytes32,bytes32,bytes32[4])",
            "hex_signature": "0x94f6e8ed",
            "bytes_signature": "öèí"
        }
    ]
}
```
the warning should be displayed
if `count > 1`
if `count === 1 && the local text signature is different from the remote one`

jvmaia: Ook, I got it, the actual code does not check if there is more than one function, updating it now!

jvmaia: it's a pleasure =)
I don't know which is the correct file to put this call, so I put the verification inside contractParser.js and storing the collisions inside detail.collisions, the compile-tab.js will only display the warnings.

yann300:
tests fails with standard issues:
>standard: Use JavaScript Standard Style (http://standardjs.com)
>  /home/circleci/repo/src/app/contract/contractParser.js:52:5: 'request' is not defined.
> /home/circleci/repo/src/app/tabs/compile-tab.js:11:5: 'request' is assigned a value but never used.

vs77bb:
@yann300 is this one almost good for payout? @jvmaia great work with multiple updates here!

yann300:
@vs77bb yes, this gonna be merged very soon

src/app/contract/contractParser.js
```diff
@@ -45,6 +46,29 @@ var getDetails = function (contractName, contract, source) {
...
+          }
+        }
+        if (displayWarn) {
+          detail.collisions.push(func)
+        }
+      }
+    })
+  }
+
   return detail
```
ryestew: I believe-  return detail - is returning synchronously - so the asyn function has not finished running yet.  This asynch code should probably be in its own file.

ryestew:
Hi jvmaia - I am sorry we have taken so long to give you another review of the code.  Because of the time that has passed it was easier to add your changes to a branch from the current master branch ( I did this locally).  I couldn't get the collision warnings showing up.  I believe it is because of this line:
https://github.com/ethereum/remix-ide/pull/1370/files#diff-60ed26d4b562f79a77d00a8cd16c0958R72
I realized quite a bit of time has passed, please let me know if you would like to finish this PR.   I believe you are returning the variable 'details' before the asynch has finished.  And also it would be better if the asynch function is in its own file.

jvmaia:
Hi @ryestew, I will be glad to finish this PR. Thanks for the feedback.

ryestew:
HI @jvmaia  - thanks so much - if you have any questions - I'll try to answer them as quickly as possible.

jvmaia:
what is this error in remix-ide tests? how can I fix it?

ryestew:
The first thing to try is to rebase the code. That's probably the issue.

ryestew:
I'm not sure which would be faster - but you could make a new fork from the current master and then make a new branch and then copy your changes over and then push to your PR.  I'll check your code as well.

jvmaia:
I did the rebase, let's see what happens

ryestew:
A test didn't pass - I'll have more info about what is going on tomorrow.

src/app/contract/verifySignatureCollisions.js
```diff
@@ -0,0 +1,23 @@
...
+    var url = 'https://raw.githubusercontent.com/ethereum-lists/4bytes/master/signatures/' + signature
+    request(url, (err, data, response) => {
+      if (err) {
+        console.log(err)
+      } else if (response.statusCode === 200 && (data.split(';').length > 1 || data !== func)) {
+        collisions.push(func)
+      }
+    })
+  }
+  return collisions
```
yann300: That would not solve the original issue here, cause you are returning `collisions` before it is filled

jvmaia: I'm calling this module as async in the others files, it fixed that problem.

jvmaia: Is there any other way to solve that? Which do you prefer?

DanielRX: The solutions I would suggest are to either return either a promise (and adjust code elsewhere to reflect that?) or to add a callback function as a param and call that function upon a collision (after `collisions.push`), however I don't know which @yann300 would prefer to fit into the style of the codebase as-is

yann300: >  to either return either a promise

that would be fine to do like that

jvmaia: Hi, sorry for the delay, my day was to busy. Tomorrow I will try to implement the your suggestion (@DanielRX ), may I come here with some questions, I'm not too good with Promises.

jvmaia:
Sorry for the delay.
It returns a promise now =) if there is something missing here please tell me.

ryestew:
@jvmaia - I'll try to have an answer about this later today.

yann300:
closing as it's very old. @ryestew anything we could do for reopening this in remix-project

Issue #1321, opened 2018-05-22
Highlight input field in UDapp when missing param (encoding error)


ninabreznik:

yann300:
in the multiparam component (https://github.com/ethereum/remix-ide/blob/master/src/multiParamManager.js) 
When the control is expanded:
 - should listen on focus out
 - if the encoding failed, input border should become red

@ryestew please could you double check that this correspond to what we planned

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.4 ETH (217.38 USD @ $543.46/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/remix-ide/1321/671).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $27,621.94 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


1. [iamonuwa](https://gitcoin.co/profile/iamonuwa) has started work. 	
 * Q: No questions for now.

iamonuwa:
@yann300 can you add images where I can reproduce this error?

yann300:
@iamonuwa that's not a bug. The idea here is to properly warn the user when data can't be encoded:

![screenshot from 2018-06-24 10 47 52](https://user-images.githubusercontent.com/6940742/41817475-169cda74-779c-11e8-859f-cff2e74c6e3c.png)

![screenshot from 2018-06-24 10 47 23](https://user-images.githubusercontent.com/6940742/41817476-1cd2e168-779c-11e8-85a9-c9eb2597c9bc.png)

in this example, the function take one parameter (address) - in both case, when the focus is out and if encoding failed , the input border should become red.

gitcoinbot:
@iamonuwa Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
[...]

Issue #1380, opened 2018-06-30
Max Call Stack Exceeded


willdeuschle:
I'm getting the same error as #1373 when I try to execute `npm run build`, and it was not resolved by switching to node 9.11.2.

`SyntaxError: Maximum call stack size exceeded while parsing /Users/willdeuschle/Dev/gitcoin/remix-ide/node_modules/solc/soljson.js while parsing file: /Users/willdeuschle/Dev/gitcoin/remix-ide/node_modules/solc/soljson.js`

MacOS 10.13.4, node v9.11.2, npm 5.6.0.

sekmet:
Same issue here.

SyntaxError: Maximum call stack size exceeded while parsing .../remix-ide/node_modules/remix-solidity/node_modules/solc/soljson.js

Debian 9, node 10.5.0, npm 6.1.0

yann300:
That is fixed by now (added `--exclude solc` in the browserify script.

prashantprabhakar:
@yann300 Facing the issue with `Current version:0.4.24+commit.e67f0147.Emscripten.clang`. Seems the issue is not resolved yet.

sachushaji:
@yann300 I am also facing a similar issue with the current remix ide. I will get the error while copy pasting the code.
![screenshot from 2018-12-21 11-43-44](https://user-images.githubusercontent.com/15767732/50327722-3ea60d80-0516-11e9-97a2-4f5f5e96bbb5.png)
But after I make a small change like adding space or anything else, the code recompiles and everything works perfectly.

PR #1381, opened 2018-07-01
Fixes Issue #1321 (Highlight input field in UDapp when missing param or encoding error)

willdeuschle:
A fix for #1321, found via bounty on Gitcoin.

We check the validity of the address, and render it red or grey accordingly:

screen shot 2018-06-30 at 8 40 23 pm

### DIFF
--- a/src/multiParamManager.js
+++ b/src/multiParamManager.js
@@ -7,7 +7,7 @@ var remixLib = require('remix-lib')
 var txFormat = remixLib.execution.txFormat
 var $ = require('jquery')
 
-function doValidation(that) {
+function doValidation (that, id) {
   var multiString = that.getMultiValsString()
   var multiJSON = JSON.parse('[' + multiString + ']')
   var encodeObj
@@ -17,18 +17,10 @@ function doValidation(that) {
     encodeObj = txFormat.encodeData(that.funABI, multiJSON)
   }
   if (encodeObj.error) {
-    console.error("MAKE IT RED")
     $(`#${id}`).addClass(css.oops.className)
   } else {
-    console.error("MAKE IT GRAY")
     $(`#${id}`).removeClass(css.oops.className)
   }
-}
-
-function checkValid(comp, id, that) {
-  console.log("are you valid", id, that)
-  doValidation(that)
-  //$(`#${id}`).addClass(css.oops.className)
 }
 
 class MultiParamManager {
@@ -43,7 +35,8 @@ class MultiParamManager {
     * @param {string} evmBC
     *
     */
-  constructor (lookupOnly, funABI, clickCallBack, inputs, title, evmBC) {
+  constructor (uniqueId, lookupOnly, funABI, clickCallBack, inputs, title, evmBC) {
+    this.uniqueId = uniqueId
     this.lookupOnly = lookupOnly
     this.funABI = funABI
     this.clickCallBack = clickCallBack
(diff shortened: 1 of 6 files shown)


src/multiParamManager.js
```diff
@@ -5,6 +5,23 @@ var css = require('./universal-dapp-styles')
 var copyToClipboard = require('./app/ui/copy-to-clipboard')
 var remixLib = require('remix-lib')
 var txFormat = remixLib.execution.txFormat
+var $ = require('jquery')
+
+function doValidation (that, id) {
```
yann300: we also need to apply this function just before sending the transaction

src/multiParamManager.js
```diff
@@ -96,11 +114,17 @@ class MultiParamManager {
...
       return yo`<div>
-        ${this.funABI.inputs.map(function (inp) {
-          return yo`<div class="${css.multiArg}"><label for="${inp.name}"> ${inp.name}: </label><input placeholder="${inp.type}" title="${inp.name}"></div>`
+        ${this.funABI.inputs.map(function (inp, idx) {
+          function setUpValidator (id) {
+            $(`#${id}`).off('blur')
+            $(`#${id}`).blur(() => doValidation(that, id))
+          }
+          var id = `${uniqueId}${inp.type}${inp.name}${idx}`
+          return yo`<div class="${css.multiArg}"><label for="${inp.name}"> ${inp.name}: </label><input id="${id}" onclick=${() => { setUpValidator(id) }} placeholder="${inp.type}" title="${inp.name}"></div>`
```
yann300: Perhaps using `onchange` event instead of onclick would be better for this? 
And to avoid spamming perhaps having a mechanism like https://github.com/ethereum/remix-ide/blob/master/src/app.js#L941
which start validating after like 700ms of inactivity

src/multiParamManager.js
```diff
@@ -96,11 +114,17 @@ class MultiParamManager {
...
+      var that = this
       return yo`<div>
-        ${this.funABI.inputs.map(function (inp) {
-          return yo`<div class="${css.multiArg}"><label for="${inp.name}"> ${inp.name}: </label><input placeholder="${inp.type}" title="${inp.name}"></div>`
+        ${this.funABI.inputs.map(function (inp, idx) {
+          function setUpValidator (id) {
+            $(`#${id}`).off('blur')
+            $(`#${id}`).blur(() => doValidation(that, id))
+          }
+          var id = `${uniqueId}${inp.type}${inp.name}${idx}`
```
yann300: Are you using `uniqueId` because of the `$(`#${id}`).addClass(css.oops.className)` in `doValidation`
If so I think doing: 
`onclick=${function() { setUpValidator(this) }}`
would be better as you don't need to rely on the `id`

src/multiParamManager.js
willdeuschle: That depends - is the desired behavior that we are validating as someone is typing or after they have left the text box? If it's while they are typing, then I can alter this to work according to `oninput`; however, the `onchange` event isn't actually fired until after you have left the textbox and so will offer the same functionality as what is already here (where the event listener is set up on the initial `onclick`).

src/multiParamManager.js
willdeuschle: do you know what event triggers the sending of the transaction so I can add the validation there?

src/multiParamManager.js
yann300: ah ok I see, the thing with "onclick" is that the validation start before the user start typing.
so `oninput` is definitely better than `onchange`

src/multiParamManager.js
yann300: the sending happens here https://github.com/ethereum/remix-ide/blob/master/src/multiParamManager.js#L134

src/multiParamManager.js
yann300: oh ok sry i didn't see that.
so yes let's keep the `onclick`.
could the red border be in a different color / thickness then when the input has the focus. In chrome that's almost the same

src/multiParamManager.js
yann300: in `doValidation` , should the other input be reset as well?
Note that there could be more than 1 param, for instances:
```
 function tes (uint _p, string _oo, address _o) {}
```

src/multiParamManager.js
willdeuschle: yes for sure I can make it thicker, right now it gets covered by the highlight

src/multiParamManager.js
willdeuschle: the other input be reset? I'm not sure what you mean, and why would there be more than one param?

yann300: `function sampleFn (uint _p, string _oo, address _o) {}` has 3 inputs
so the panel looks like
screenshot from 2018-07-05 13 09 13
if the validation success, the red border should be removed on all the inputs

vs77bb:
Hi @yann300 are you reviewing this one currently?

vs77bb:
Hi @willdeuschle are you able to make the changes requested? Would love to pay this one out soon assuming we can get this one merged!

willdeuschle:
Hey @vs77bb I would like to make the final changes to this, but I'm still blocked on a painful devx bug here: https://github.com/ethereum/remix-ide/issues/1380

vs77bb:
Thanks for the update @willdeuschle - I see the #1380 is still an issue. @yann300 - do you think that one is good for a bounty? Any ideas on what the best path forward here is?

yann300:
@willdeuschle sorry for the delay. Please could you retry now? the bug you've mentioned should be fixed by now.

Issue #1359, opened 2018-06-13
Add new (reserved) Solidity 0.5.0 keywords to syntax highlighting


yann300:
PR https://github.com/ethereum/remix-ide/pull/1333

@raphaelhuefner

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.22 ETH (103.07 USD @ $468.49/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/remix-ide/1359/613).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $24,293.42 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

vs77bb:
@raphaelhuefner Would you like to claim this one on Gitcoin?

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


1. [Legitimate](https://gitcoin.co/profile/Legitimate) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ethereum/remix-ide/1359/613?mutate_worker_action=approve&worker=Legitimate) | [reject worker](https://gitcoin.co/issue/ethereum/remix-ide/1359/613?mutate_worker_action=reject&worker=Legitimate))_. 	
 * Q: I plan on installing thebrequired software to begin developing for this project. Is there a slack channel I may join?

raphaelhuefner:
Sorry, forgot the most important step for the

## Action Plan:

- Re-integrate separate NPM package into remix-ide.

yann300:
@raphaelhuefner what do you mean exactly by ` remix-ide's testing.` and `testing setup` ?

yann300:
oh by `remix-ide's testing ` you mean using the compile, run and debugger tab from the right panel right?

yann300:
or do you men end-to-end browser testing?

raphaelhuefner:
@yann300 No, not the Solidity testing in the UI, I meant everything in these directories:

[...]

PR #1415, opened 2018-07-17
WIP Add NPM package for Solidity syntax highlighting.

raphaelhuefner:
This PR is for https://github.com/ethereum/remix-ide/issues/1359

Start using https://www.npmjs.com/package/ace-mode-solidity , which is an improved version of `src/app/editor/mode-solidity.js`.


raphaelhuefner:
This is a list of the Solidity keywords which are now recognized:

`
.balance
.callcode
.delegatecall
.transfer
@author
@notice
_
abi.encode
abi.encodePacked
abi.encodeWithSelector
abi.encodeWithSignature
abstract
addmod
after
alias
anonymous
apply
as
assert
auto
block.blockhash
block.coinbase
block.difficulty
block.gaslimit
block.number
block.timestamp
blockhash
calldata
catch
constructor
copyof
days
default
define
ecrecover
emit
ether
final
finney
fixed
fixedMxN
from
gasleft
hours
immutable
implements
import
indexed
inline
interface
is
keccak256
let
macro
match
minutes
msg.data
msg.gas
msg.sender
msg.value
mulmod
mutable
now
null
of
override
partial
pragma solidity ^x.y.z
pragma experimental "x.y.z"
promise
reference
relocatable
require
revert
ripemd160
sealed
seconds
selfdestruct
sha256
sha3
sizeof
static
suicide
super
supports
switch
szabo
tx.gasprice
tx.origin
type
typedef
ufixed
ufixedMxN
unchecked
weeks
wei
years
|
|=
`

axic:
@raphaelhuefner why is this WIP? I think this may be working properly?

@yann300 can this be merged / looked at? This is the last step for finishing off a bounty.

raphaelhuefner:
@axic Please let me get back to you on this one on Monday.

ryestew:
@raphaelhuefner  - in this version you are not making any additions to the style-guide.js.  Is there no need to now?

raphaelhuefner:
@ryestew Thank you for looking into this!

The syntax highlighter itself works as-is, without any changes to either `style-guide.js` or any ACE theme. It still tags all Solidity keywords with `<span>`s & classes, even if some of those classes go un-styled meanwhile. (I think so far I have failed to state this last bit clearly enough to anyone.)

We **did indeed** talk on gitter about pulling color values from `style-guide.js` and injecting them into an ACE theme. I have done some work towards that already, just got delayed a bit by ... life. ;-)

I would like to split that color work out into it's own separate PR for these reasons:

- It's much more work than I originally thought.
- It will end up not touching the syntax highlighter at all.

I am fine with the new "color injection" PR having no bounty:

- I'll do it anyways because I have invested quite a bit of time already.
- I agree with you that the syntax highlighter **and** the color injection **together** will be a more well-rounded functionality.

raphaelhuefner:
@ryestew Even if some auxiliary highlight classes go un-styled for now, this PR is still an improvement over the current situation where a whole bunch of keywords are not even tagged at all.

Are you OK with splitting this work into 2 self-contained PRs?

Please let me know if you have any other questions about this PR.

If you feel that this is easier to talk about on a call, I'm up for it. Please let me know when would be a good time for you. I assume you would use https://meet.jit.si/ , right?

raphaelhuefner:
@ryestew I rebased this PR onto the latest master, and now circle CI fails with something I don't understand:

> Running:  SharedFolderExplorer
> [...]
> ✖ Timed out while waiting for element <[data-path="localhost/folder1/contract_firefox.sol"]> to be removed for 10000 milliseconds.  - expected "not found" but got: "found"

yann300:
@raphaelhuefner 
just restarted circle ci

raphaelhuefner:
Thank you, it works again!

PR #1452, opened 2018-08-08
Use Web3 1.0.0 in remix-ide

yann300:


serapath:
ok, with this version it works :-)

Also the script that i'm trying is this one: `https://pastebin.com/raw/SXWcXn2t`
It actually just by accident has a `console.warn` in which breaks it, but adding it would fix it.

Otherwise the terminal seems to work fine

PR #1453, opened 2018-08-09
Add swarmgw to the remix terminal

yann300:


src/app/panels/terminal.js
```diff
@@ -8,6 +8,7 @@ var vm = require('vm')
 var remixLib = require('remix-lib')
 var EventManager = remixLib.EventManager
 var Web3 = require('web3')
+var swarmgw = require('swarmgw')
```
axic: Please use version 0.3.1 which also has a new API: `require('swarmgw')()` (you could specify opts, but it defaults to `https://swarm-gateways.net` so that should be ok for remix)

Issue #952, opened 2017-12-14
Signing capabilities with keys in JS VM environment


VoR0220:
Currently it is quite a difficult process to sign some message and perform operations like ECRecover with the given available keys in the javascript VM environment. A proposed solution would be a separate tab for signing some message of some kind and returning the msg hash, v, r, and s as outputs to perform the ecrecover.

axic:
Also #1064 would make this easier.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 240.0 DAI (240.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/remix-ide/952/860).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $30,155.12 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


1. [oogetyboogety](https://gitcoin.co/profile/oogetyboogety) has started work. 	
 * Q: Using these docs, I will seek out best practicies to create a new template that signs a message using the Javascript VM keys provided 

View/CSS: new template inside the directory https://github.com/ethereum/remix-ide/tree/master/src/app/tabs, following the convention listed there.

Model/Controller updaes:
Based on https://github.com/chriseth/browser-solidity/pull/154/files
Update and use the rest of the accounts functions in https://github.com/axic/browser-solidity/blob/83b2bf3ba5924b5f6833c3861db8fec493a617ed/assets/js/universal-dapp.js to be called in the new tab to sign a message using the docs here: https://github.com/ethereumjs/ethereumjs-util/blob/master/docs/index.md

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/remix-ide/952/860).

oogetyboogety:
I will also use the docs here: 
HD Wallet API
To use BIP32 HD wallets, first include the hdkey submodule:
[...]

PR #1454, opened 2018-08-10
Fix 952 (sign messages)

oogetyboogety:
https://github.com/ethereum/remix-ide/issues/952
I think the last thing to do would be to create the test code


src/app/tabs/run-tab.js
```diff
@@ -581,6 +583,29 @@ function settings (container, self) {
       }
     })
   }
+  function signMessage (event) {
```
yann300: we might need to modify this function so it it also working with Web3:
if the current env is `JavaScript VM` use the current behavior.
if the current env is not JavaScript VM and if personal mode is enabled `self._deps.config.get('settings/personal-mode')`, 
use the following function to sign: web3.personal.sign(message, account, password) 
https://github.com/ethereum/go-ethereum/wiki/Management-APIs#personal_sign
(the ui then need to ask for the password - using https://github.com/ethereum/remix-ide/blob/master/src/app/ui/modal-dialog-custom.js#L12)

oogetyboogety: @yann300 I tried to use the personal_sign RPC call from web3 but now Metamask is not responding to any calls to get the account list or anything, similar to this issue here.
https://ethereum.stackexchange.com/questions/52197/metamask-remix-ide-cannot-create-an-account-error-internal-json-rpc-error
Any ideas as to what could be the root cause of the problem?
FYI, I noticed the personal module in exectionContext.web3().personal doesnt seem to contain a sign function. Then something I did caused this issue to occur.

src/app/tabs/run-tab.js
```diff
@@ -581,6 +584,53 @@ function settings (container, self) {
...
+    self._deps.udapp.getAccounts((err, accounts) => {
+      if (err) { addTooltip(`Cannot get account list: ${err}`) }
+      var stringifyRSV = (rsv) => {
+        var rsvJson = () => { return JSON.stringify(rsv, null, '\t') } 
+        copyToClipboard(rsvJson).onclick(event)
```
yann300: The copy to clipboard seems not working when using `web3` ...

yann300: I was connected to a local geth using web3 provider

src/app/tabs/run-tab.js
```diff
@@ -581,6 +584,53 @@ function settings (container, self) {
...
+          }
+        }, false)
+      } else {
+        modalDialogCustom.promptPassphrase('Passphrase to sign a mesasge', 'Enter your passphrase for this account to sign the message', 'Passphrase', (passphrase) => {
+          modalDialogCustom.promptMulti(signMessageDialog, (message) => {
+            console.log(executionContext.web3())
+            const hashedMsg = executionContext.web3().sha3(message)
+            try {
+              var personal = new Personal(executionContext.web3().currentProvider)
+              personal.sign('0x' + Buffer.from(message).toString('hex'), account, passphrase, (error, signedData) => {
```
yann300: for metamask, doing web3.eth.sign(address, message)

src/app/tabs/run-tab.js
yann300: perhaps would be more convenient copy to clipboard the raw hexadecimal string of the encoded value.

oogetyboogety: @yann300 Hmm I was able to get it working using local geth with web3 provider...I will provide a VM in a few minutes...
did you run ./geth -syncmode "light" --rpc --rpcapi eth,web3,personal --rpccorsdomain "*"?

yann300: it was working but the content was not copied to the clipboard...

oogetyboogety: @yann300 what browser/version are you on

yann300: chromium 62.0.3202.62 
copying a "vm" signed message worked, but not a msg signed using web3...
let me retry though

yann300: got the same, that is very weird... are you able to reproduce?

src/app/tabs/run-tab.js
```diff
@@ -581,6 +584,53 @@ function settings (container, self) {
...
+                console.log(error)
```
yann300: we could display the error message in a tooltip here (mean the passphrase is incorrect)

src/app/tabs/run-tab.js
oogetyboogety: @yann300 hmmm no error with chrome and web3 and geth on my end. Can you forward anything red you see in the Developer Tools console in Chrome?

yann300: i don't have any error with web3 but still the coded msg is not copied to the clipboard..

oogetyboogety: @yann300 could you try now

yann300: same .... the copy function works when the javascript VM is used, but not with web3...
I also tried to directly use `const copy = require('clipboard-copy')` ...
that's really weird

yann300: Could you replicate ? (that 's also not working on firefox)

Why not removing the auto copy to clipboard and to display r,s,v (as hex string) and the raw signed message in a pre HTML tag. then user would need to manually copy it.
I fear there might be issue with this copytoclipboard lib in the context we are using it...

oogetyboogety: @yann300 you are correct! I was able to replicate but I fixed it by avoiding the clipboard libraries with these lines. Could you pull the latest commit and try testing that one?
```document.execCommand('copy')
            if (navigator.clipboard) {
              navigator.clipboard.writeText(rsv)
            }```

Comment (oogetyboogety) [reply src/app/tabs/run-tab.js]: @yann300 Pre HTML tag will fix the issue, and I think the context would actually fit the copyToClipboard library (in Firefox), and allow it to work. I do see the message : 
```
```
    document.execCommand(‘cut’/‘copy’) was denied because it was not called from inside a short running user-generated event handler.
```
So this would circumvent that issue

src/app/tabs/run-tab.js
```diff
@@ -13,11 +13,13 @@ var helper = require('../../lib/helper.js')
 var executionContext = require('../../execution-context')
 var modalDialogCustom = require('../ui/modal-dialog-custom')
 var copyToClipboard = require('../ui/copy-to-clipboard')
+const Buffer = require('safe-buffer').Buffer
 var Card = require('../ui/card')
 var Recorder = require('../../recorder')
 var addTooltip = require('../ui/tooltip')
 var css = require('./styles/run-tab-styles')
 var MultiParamManager = require('../../multiParamManager')
+var Personal = require('web3-eth-personal')
```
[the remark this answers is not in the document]
oogetyboogety: @yann300 I think we might be good, I found it in my node_modules directory already.

src/app/tabs/run-tab.js
oogetyboogety: @yann300 Could you test this change in the newest commit

oogetyboogety: @yann300 done

src/app/tabs/run-tab.js
oogetyboogety: @yann300 I think I got the latest version of metamask to work with personal.sign.

src/app/tabs/run-tab.js
oogetyboogety: @yann300 Done!

src/app/tabs/run-tab.js
```diff
@@ -96,39 +98,66 @@ function runTab (opts, localRegistry) {
...
-      <div class=${css.recorderDescription}>
-        ${self.data.text}
-      </div>
-      <div class="${css.transactionActions}">
-        ${recorderInterface.recordButton}
-        ${recorderInterface.runButton}
-        </div>
-      </div>
-    </div>`
+  self.rpcResponseOpts = {
```
yann300: I think that the rpc response should not be in the main panel, that take too much space. "signing message" is still a side feature of this tab. But thanks for giving a try that way.
would it be possible to display the result 1) in an another modal 2) I think only the hex string value need to be displayed.

src/app/tabs/run-tab.js
oogetyboogety: @yann300 This was working before I upgraded Metamask, now it is no longer getting the account list due to some JSON RPC error?

src/app/tabs/run-tab.js
```diff
@@ -581,6 +584,52 @@ function settings (container, self) {
...
+      var isVM = executionContext.isVM()
+      if (isVM) {
+        modalDialogCustom.promptMulti(signMessageDialog, (message) => {
+          const personalMsg = ethJSUtil.hashPersonalMessage(Buffer.from(message))
+          var privKey = self._deps.udapp.accounts[account].privateKey
+          try {
+            var rsv = ethJSUtil.ecsign(personalMsg, privKey)
+            var rsvJson = JSON.stringify(rsv, null, '\t')
+            var signedData = ethJSUtil.toRpcSig(rsv.v, rsv.r, rsv.s)
+            modalDialogCustom.alert(rsvJson + '\n' + signedData)
```
oogetyboogety: @yann300 Could you review here

src/app/tabs/run-tab.js
```diff
@@ -581,6 +584,52 @@ function settings (container, self) {
...
+            console.log(executionContext.web3())
+            const hashedMsg = executionContext.web3().sha3(message)
+            try {
+              var personal = new Personal(executionContext.web3().currentProvider)
+              personal.sign('0x' + Buffer.from(message).toString('hex'), account, passphrase, (error, signedData) => {
+                if(error && erorr.message !== '') {
+                  console.log(error)    
+                  addTooltip(error.message)
+                }
+                modalDialogCustom.alert(signedData)
```
oogetyboogety: @yann300 Could you review here

src/app/tabs/run-tab.js
yann300: it is working with metamask for me. although I think the password should not be asked if it is `injected provider` like metamask.

src/app/tabs/run-tab.js
```diff
@@ -581,6 +584,52 @@ function settings (container, self) {
...
+        }, false)
+      } else {
+        modalDialogCustom.promptPassphrase('Passphrase to sign a mesasge', 'Enter your passphrase for this account to sign the message', 'Passphrase', (passphrase) => {
+          modalDialogCustom.promptMulti(signMessageDialog, (message) => {
+            console.log(executionContext.web3())
+            const hashedMsg = executionContext.web3().sha3(message)
+            try {
+              var personal = new Personal(executionContext.web3().currentProvider)
+              personal.sign('0x' + Buffer.from(message).toString('hex'), account, passphrase, (error, signedData) => {
+                if(error && erorr.message !== '') {
```
yann300: there 's a typo here `erorr`

src/app/tabs/run-tab.js
yann300: no need to alert here if there's an error

oogetyboogety: Fixed!

src/app/tabs/run-tab.js
oogetyboogety: @yann300 Fixed!

src/app/tabs/run-tab.js
oogetyboogety: Ah I  see. How may I check for an injected provider? Is it ```executionContext.getProvider()```?

yann300: yes , it returns `injected` in that case

yann300:
would you be able to give a rebase on this PR please?

oogetyboogety:
@yann300 I tried forking form the latest commit, but I am getting these errors:
app.js:246653 Uncaught TypeError: Cannot read property 'getListAccounts' of undefined
    at UniversalDApp.getAccounts (app.js:246653)
    at fillAccountsList (app.js:241471)
    at new runTab (app.js:241449)
    at new RighthandPanel (app.js:239480)
    at App.run (app.js:233359)
    at App.init (app.js:233054)
    at Object.1482../app.js (app.js:245222)
    at o (app.js:1)
    at r (app.js:1)
    at app.js:1
2app.js:157221 Uncaught TypeError: this.web3(...).version.getNetwork is not a function
    at ExecutionContext.detectNetwork (app.js:157221)
    at updateNetwork (app.js:241782)

not sure what's going on...

PR #1459, opened 2018-08-14
add console.warn to terminal

serapath:


src/app/panels/terminal.js
```diff
@@ -481,7 +483,7 @@ class Terminal {
         if (args.length) append(args[0])
       }
     }
-    mode = { log: styles.terminal.text_RegularLog, info: styles.terminal.text_InfoLog, error: styles.terminal.text_ErrorLog }[mode] // defaults
+    mode = { log: styles.terminal.text_RegularLog, info: styles.terminal.text_InfoLog, warn: 'yellow', error: styles.terminal.text_ErrorLog }[mode] // defaults
```
yann300: would it be possible to add this in the style ? (there are now directly in remix-ide repo)

serapath: yes

Issue #1457, opened 2018-08-13
Warn user at deploy time if contract size above https://github.com/ethereum/EIPs/issues/170


yann300:

yann300:
by displaying an alert saying the size is above https://github.com/ethereum/EIPs/issues/170 and that the deployment will likely fails.
this alert should give the option to cancel or to force sending by using this component https://github.com/ethereum/remix-ide/blob/master/src/universal-dapp.js#L412

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


1. [bakaoh](https://gitcoin.co/profile/bakaoh) has started work. 
    
    As i understand, i only need to check if `to == null` and `args.data.contractBytecode.length >= 24577` to show dialog?

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/remix-ide/1457/953).

yann300:
@bakaoh I think the change should be done there https://github.com/ethereum/remix-ide/blob/master/src/app/tabs/run-tab.js#L424 

Actually I think we should check against the `runtimeByteCode` (should be available in `selectedContract.contract`) cause this is the one which is then stored.
FYI here how it is implemented in the ethereumjs-vm https://github.com/ethereumjs/ethereumjs-vm/blob/f389fe7cdd199a8339b4f82a05464b34e4712975/lib/runCall.js#L189

bakaoh:
@yann300 -- Thanks for pointing that out to me. ​I didn't read carefully.

PR #1461, opened 2018-08-15
Warn user at deploy time if contract size above ethereum/EIPs#170

bakaoh:
Fix: #1457


src/app/tabs/run-tab.js
```diff
@@ -421,24 +422,42 @@ function contractDropdown (events, self) {
...
+        } else {
+          txFormat.encodeConstructorCallAndLinkLibraries(selectedContract.contract.object, args, constructor, contractMetadata.linkReferences, selectedContract.contract.object.evm.bytecode.linkReferences, (error, data) => {
+            createInstanceCallback(error, selectedContract, data)
+          })
+        }
+      })
+    }
+
+    if (selectedContract.contract.object.evm.deployedBytecode.object.length > 24576) {
+      modalDialog('Contract code size over limit', yo`<div>Contract creation initialization returns data with length of more than 24576 bytes. The deployment will likely fails.</div>`,
```
yann300: could we add a link to the iep too?

bakaoh: @yann300 -- i've added it ^^

yann300: thanks but i meant as a link in the content of the modal too

bakaoh:
@yann300 --- anything else i need to do?

src/app/tabs/run-tab.js
```diff
@@ -421,24 +422,44 @@ function contractDropdown (events, self) {
...
+          })
+        } else {
+          txFormat.encodeConstructorCallAndLinkLibraries(selectedContract.contract.object, args, constructor, contractMetadata.linkReferences, selectedContract.contract.object.evm.bytecode.linkReferences, (error, data) => {
+            createInstanceCallback(error, selectedContract, data)
+          })
+        }
+      })
+    }
+
+    if (selectedContract.contract.object.evm.deployedBytecode.object.length > 24576) {
```
yann300: `object` here is hex string.
1 hex string is 1/2 bytes. 
so you need to divide by 2 here to get the length in bytes

bakaoh: @yann300 - thanks, i fixed it

Issue #1227, opened 2018-04-09
UI/UX IMPROVEMENTS - COMPILE TAB


ninabreznik:
## For the beginning just a list of visual bugs:
- rotating icon on compile tab showing that compilation is in process is not working
![screen shot 2018-04-09 at 10 39 29 pm](https://user-images.githubusercontent.com/1076427/38524619-29eff344-3c47-11e8-9089-b0d926269f19.png)

- auto compile tag should be wider when panel is open and Attention triangle and the tag should be closer together
![screen shot 2018-04-09 at 10 42 19 pm](https://user-images.githubusercontent.com/1076427/38524662-558cab00-3c47-11e8-806a-bc4d9feba94f.png)

- attention triangle (where suggest we turn off auto compile) should appear only when autocompile is on
- is this notification still needed?
![screen shot 2018-04-09 at 10 46 04 pm](https://user-images.githubusercontent.com/1076427/38524819-d8569a6e-3c47-11e8-95e4-30a3717cf9ae.png)

- buttons
- all buttons look the same, but one opens a modal dialog, one publishes to Swarm and two copy to clipboard. Try to find a way to visually communicate the function of each or at least to create different look for each.
![screenshot_2018-08-14_01-46-27](https://user-images.githubusercontent.com/1076427/44065669-ef049bec-9f63-11e8-9ba2-c0d9476a378e.png)

## Some inspiration
- Errors should be logged in the terminal
- on dropdown you should be able to select any .sol file with a contract and compile it (this would also auto open the file in the editor)

yann300:
for the `is this notification still needed?` .
The reason of this is to clearly warn the user that there's no error, no warning and everything is ok.
but there are certainly better wasy to show that

yann300:
`on dropdown you should be able to select any .sol ...` I don't completely get, cause the dropdown does not   show sol file but compiled contracts. let's check that when you have time

yann300:
comment from @ninabreznik from another thread:

> i think compile button is not so nicely visible now, so I would maybe move it inside of the box above so compiler selection and compile button are there togethe

r.

PR #1466, opened 2018-08-17
Clean support tab content

yann300:
- move solidity version selector to compile tab
 - move remix, remixd info to support tab
 - refactor plugin loading/removing

### DIFF
--- a/src/app/panels/righthand-panel.js
+++ b/src/app/panels/righthand-panel.js
@@ -15,10 +15,8 @@ const SupportTab = require('../tabs/support-tab')
 const PluginTab = require('../tabs/plugin-tab')
 const TestTab = require('../tabs/test-tab')
 const RunTab = require('../tabs/run-tab')
-const plugins = require('../plugin/plugins')
 const DraggableContent = require('../ui/draggableContent')
 
-var toolTip = require('../ui/tooltip')
 const EventManager = remixLib.EventManager
 const styles = styleguide.chooser()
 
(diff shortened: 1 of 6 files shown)


ninabreznik:
Feedback

**Support tab**
- run remix locally (does this belong to Remix or would we rather have this kind of info in the documentation?) Maybe we should make a text above the chat, where we link to documentation, more visible and a bit bigger. I think it would be better to have all the info that is not part of the app in one place and this is documentation.
- accessing local files (same as above and/OR where it belongs - in the module that appears when you click on RemixD icon.)

**Settings tab**
- styling in plugin box is broken. Buttons behave strange and text is not align in the center (is it using style guide styling we use for buttons?)
- when you click Oraclize, some React/Redux stuff logs in the console.
- refactoring plugin - what should be checked here, I couldn't figure it out.

**Compile tab**

- i think compile button is not so nicely visible now, so I would maybe move it inside of the box above so compiler selection and compile button are there together.

And then we can have details and other buttons in the next box. I will be working on fixing the compile tab a bit, so I can take over this part.

yann300:
> I think it would be better to have all the info that is not part of the app in one place and this is documentation.

Right, the idea behind that (put the info in remix / settings directly) was to put these information in a place where user go frequently.

>styling in plugin box is broken. Buttons behave strange and text is not align in the center (is it using style guide styling we use for buttons?)

might be yes

> when you click Oraclize, some React/Redux stuff logs in the console.
that come from the Oraclize code, you can't do so much here

> refactoring plugin - what should be checked here, I couldn't figure it out.

nothing specific. just that now, when you paste a plugin description and click load, it add the plugin to the list of available plugin. before it would just start it, with no way of having a custom plugin in list.

> I will be working on fixing the compile tab a bit, so I can take over this part.

ok, linking the issue https://github.com/ethereum/remix-ide/issues/1227

yann300:
thanks for the comment @ninabreznik 
I have dispatched everything in other thread / issues.

yann300:
merging this, to be continued in next PRs

PR #1482, opened 2018-08-24
Compile tab ui

ninabreznik:
UI fixes in the Compile tab

screenshot_2018-08-24_02-28-24
screenshot_2018-08-24_02-28-08

### DIFF
--- a/src/app/tabs/compile-tab.js
+++ b/src/app/tabs/compile-tab.js
@@ -236,7 +236,7 @@ module.exports = class CompileTab {
         <div class="${css.contractContainer}">
           ${self._view.contractNames}
           <div title="Publish on Swarm" class="${css.publish}" onclick=${publish}>
-            <i class="${css.copyIcon} fa fa-upload" aria-hidden="true"></i>
+            <i class="${css.copyIcon} fa fa-upload" aria-hidden="true"></i><span>Swarm</span>
           </div>
         </div>
         <div class="${css.contractHelperButtons}">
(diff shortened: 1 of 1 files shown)


yann300:
Overall looks better!
some comments though:
 - There's a little glitch when the compiler loading icon switch from `idle` to `compiling`.
 - Isn't the Details button too large?
 - Was not really sure about changing the icon of `publish to Swarm` into this. At first sight that looks not really consistent with all the other actions... well would be great to have another point of view @ryestew

ryestew:
I think the buttons should all be gray - if not they look like one button is in a down state ( well that is if 1 color is darker). So given that - here is what it looks like if the buttons are the same width. 
image

ryestew:
But when the rt panel is a bit narrow the buttons need some help.
image

ryestew:
Though this looks pretty good:
image

ryestew:
Or take out the buttons for the copy icons because the icon already catchy like a button
image

ryestew:
All of these choices minimizes the importance of "Publish to Swarm".  Publish to swarm is really cool even if not that many people currently use it.  When its an icon on the right it may be too understated.  Just a sec...

ryestew:
If we make the pulldown a bit more narrow: 
 image

ryestew:
Here's the whole panel:
image

yann300:
Should the Swarm action have rounded border like for abi and bytecode?

ryestew:
@yann300 - here's what it looks like for me - there is not button for swarm - just the icon.  But maybe I don't follow you:
image

PR #1523, opened 2018-09-17
Various fixes

yann300:
- don't clear input after sending
- add api in terminal "getFile"
- improve UX run tab
- add info icon

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,10 +26,10 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - dep-bundle-21-{{ checksum "package.json" }}
+            - dep-bundle-22-{{ checksum "package.json" }}
       - run: npm install
       - save_cache:
-          key: dep-bundle-21-{{ checksum "package.json" }}
+          key: dep-bundle-22-{{ checksum "package.json" }}
           paths:
             - ~/repo/node_modules
       - run: npm run lint && npm run test && npm run make-mock-compiler && npm run build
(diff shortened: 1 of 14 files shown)


ryestew:
- On the runtab- how is i next to Environment different from ? next to the contract pulldown?  I think it should the ? without the circle is a bit clearer - fa-question and maybe at 13 or 14px.  If they are basically doing the same thing - adding a bit of info & context - then they should both be an "i".

- Were we getting some complaints about confusion between atAddress and deploy?  It does make it kind of conversational and more legible to have the "or".  I tried putting a line separating them instead of the "or" but it looked more distracting.  So lets try the "or".  Maybe add 13px of margin-top above the Deploy:

image

For not clearing the inputs - do users generally not want the inputs cleared after submitting?  I assumed clearing would be better - or is clearing better for methods in the contract and not clearing is better for the constructor?

yann300:
> On the runtab- how is i next to Environment different from ? next to the contract pulldown? I think it should the ? without the circle is a bit clearer - fa-question and maybe at 13 or 14px. If they are basically doing the same thing - adding a bit of info & context - then they should both be an "i".

right ;)

yann300:
> Were we getting some complaints about confusion between atAddress and deploy?

not really but we were discussing that point during a call a few weeks/month ago

yann300:
> For not clearing the inputs - do users generally not want the inputs cleared after submitting? I assumed clearing would be better - or is clearing better for methods in the contract and not clearing is better for the constructor?

I think there's a bug filed from an user for that (can't find the issue though), and after having tried a bit last week function / constructor execution, i think that more convenient to not clear the input. The main reason is that you often need to test the same function and you need to be able to easily copy/paste the inputs.

I think the better would be to be able to have some sort of autocompletion e.g a way to easily retrieve the inputs, so not clearing is like a "work around" for now.

ryestew:
Everything looks fine - except the i seems a bit big - probably it should not have the font-size specified - as I had suggested when it was a ?.    Its a very tiny detail though. We can proceed with or without this change.

ryestew:
1 more comment - the spacing between the Deploy button and the or and then between the or and the AtAddress button is not the same.  The Ataddress needs to have a margin-top of 6px - I think.

PR #1526, opened 2018-09-18
Alert if panic error

yann300:


src/app/tabs/compile-tab.js
```diff
@@ -138,6 +138,13 @@ module.exports = class CompileTab {
       if (data['error']) {
         error = true
         self._deps.renderer.error(data['error'].formattedMessage, self._view.errorContainer, {type: data['error'].severity || 'error'})
+        if (data['error'].mode === 'panic') {
+          return modalDialogCustom.alert(yo`<div><i class="fa fa-exclamation-circle ${css.panicError}" aria-hidden="true"></i><br>
+                                            The compiler returned with the following internal error: <br> <b>${data['error'].formattedMessage}.<br> 
+                                            The compiler might be in a non-sane state, please be careful and do not use further compilation data to deploy to mainnet. 
+                                            It is heavily recommended to use another browser (not affected by this issue) until the issue is finally fixed.</b><br>
```
axic: Well it is not something which can be fixed (that specific instance was workarounded). It is rather a limitation in the browser for Javascript stack depth, could happen in the future with different browsers having different limits.

PR #1534, opened 2018-09-21
Fix sendCompilationResults API call

oogetyboogety:
Some small corrections to fix up the sendCompilationResults API call


yann300:
that feature was pushed in another PR

PR #1536, opened 2018-09-26
Support several compiler

yann300:
Support having a plugin sending compilation artefact to the run tab

### DIFF
--- a/src/app/debugger/debugger.js
+++ b/src/app/debugger/debugger.js
@@ -17,16 +17,14 @@ function Debugger (container, sourceHighlighter, localRegistry) {
   this._deps = {
     offsetToLineColumnConverter: this._components.registry.get('offsettolinecolumnconverter').api,
     editor: this._components.registry.get('editor').api,
-    compiler: this._components.registry.get('compiler').api
+    compiler: this._components.registry.get('compiler').api,
+    compilersArtefacts: this._components.registry.get('compilersartefacts').api
   }
   this.debugger = new Ethdebugger(
     {
       executionContext: executionContext,
       compilationResult: () => {
-        var compilationResult = this._deps.compiler.lastCompilationResult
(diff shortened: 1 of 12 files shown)


src/app/debugger/debugger.js
```diff
@@ -17,16 +17,14 @@ function Debugger (container, sourceHighlighter, localRegistry) {
   this._deps = {
     offsetToLineColumnConverter: this._components.registry.get('offsettolinecolumnconverter').api,
     editor: this._components.registry.get('editor').api,
-    compiler: this._components.registry.get('compiler').api
+    compiler: this._components.registry.get('compiler').api,
+    compilersArtefacts: this._components.registry.get('compilersartefacts').api
```
iurimatias: depending on the order or merges, we need to either add this change to the debugger in other PR or to this one

src/app/debugger/debugger.js
```diff
@@ -17,16 +17,14 @@ function Debugger (container, sourceHighlighter, localRegistry) {
...
   }
   this.debugger = new Ethdebugger(
     {
       executionContext: executionContext,
       compilationResult: () => {
-        var compilationResult = this._deps.compiler.lastCompilationResult
-        if (compilationResult) {
-          return compilationResult.data
-        }
+        if (this._deps.compilersArtefacts['__last']) return this._deps.compilersArtefacts['__last'].getData()
```
iurimatias: *artifacts

src/app/debugger/debugger.js
yann300: I think that s managed below: https://github.com/ethereum/remix-ide/pull/1536/files#diff-23075dc7d1a122a47f6be2a98ee4770cR27

in `compilationResult` instead of returning the solidity compiler result, we return the last one, could be solidity or other one

PR #1542, opened 2018-10-04
Added check/uncheck all cheeckbox to StaticAnalysisTab

LianaHus:


yann300:
could you give a rebase on that one? should fix the test

PR #1546, opened 2018-10-08
move events to their own file as a intermediary step for further refa…

iurimatias:
…ctoring


yann300:
can we post pone this one for a little while?

PR #1547, opened 2018-10-09
made vertical margin invisible

LianaHus:
removed vertical line from editor


src/app/editor/editor.js
```diff
@@ -297,7 +297,7 @@ function Editor (opts = {}, localRegistry) {
 
   // Unmap ctrl-t & ctrl-f
   editor.commands.bindKeys({ 'ctrl-t': null })
-
+  editor.setShowPrintMargin(false)
```
yann300: great ;)

PR #1549, opened 2018-10-09
Add (initial) Context Manager, move execution context of out debugger

iurimatias:


src/app/debugger/debuggerUI.js
```diff
@@ -106,6 +172,7 @@ class DebuggerUI {
   startDebugging (blockNumber, txNumber, tx) {
     const self = this
 
+    this.debugger.debugger.updateWeb3(this.executionContext.web3())
```
yann300: would this be `this.debugger.updateWeb3(...`

src/app/debugger/debuggerUI.js
```diff
@@ -24,18 +30,78 @@ var css = csjs`
...
     this.debugger = new Debugger({
-      executionContext: executionContext,
+      web3: this.contextManager.getWeb3(),
       offsetToLineColumnConverter: this.registry.get('offsettolinecolumnconverter').api,
       compiler: this.registry.get('compiler').api
     })
 
+    this.contextManager.initProviders()
+
+    this.contextManager.event.register('providerChanged', () => {
```
yann300: then `switchProvider` has to be called from somewhere. but don't see where

src/app/debugger/debuggerUI.js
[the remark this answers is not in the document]
iurimatias: yup you're right it should just be  `this.debugger.updateWeb3(getWeb3())`

PR #1550, opened 2018-10-14
Add remix command "setFile"

yann300:


src/lib/cmdInterpreterAPI.js
```diff
@@ -172,6 +173,18 @@ class CmdInterpreterAPI {
       cb('file not found')
     }
   }
+  setFile (path, content, cb) {
+    var provider = this._deps.fileManager.fileProviderOf(path)
+    if (provider) {
+      provider.set(path, content, (error) => {
+        if (error) return cb(error)
+        this._deps.fileManager.syncEditor(path)
+        cb()
```
ryestew: This line probably needs to be if (cb) { cb() }
```suggestion
        if (cb) { cb() }
```

yann300: i put the change above, so that apply to the function

PR #1551, opened 2018-10-15
Contract metadata

yann300:
- add <contractname>.abi & <contractname>.data.json
 - rename <contractname>.json => <contractname>.deploy.json


ryestew:
@yann300 - Its much better with all the data rolled into 1 file.  
This comment might be out of scope of this PR but would changing the color of a json file in the file explorer be an option?  Or at least an auto generated json file:

image

PR #1631, opened 2018-11-30
Update ballot.sol to match the one shown in the IDE

corollari:
The contract in src/app/editor/example-contracts.js (which is the one that appears in the app) was updated on #1612 but this file hasn't been updated so it's inconsistent with the other one, this PR removes this inconsistency.
I've looked around in the codebase and haven't been able to find any code which uses this file, if that's the case the best course of action is to delete it, but I don't know the ins and outs of this project so it's quite possible that I've missed something in my quick search.

### DIFF
--- a/contracts/ballot.sol
+++ b/contracts/ballot.sol
@@ -1,144 +1,65 @@
-pragma solidity ^0.4.0;
-
-/// @title Voting with delegation.
+pragma solidity >=0.4.22 <0.6.0;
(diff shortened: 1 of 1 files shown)


LianaHus:
thanks @corollari

LianaHus:
@yann300  could this be merged?

corollari:
any update here?

PR #1650, opened 2019-01-02
switch to JSVM 2.5.0

yann300:


cdetrio:
any updates on this? I'd like to test Constantinople opcodes in remix.

src/app/debugger/debugger/debugger.js
```diff
@@ -51,9 +51,7 @@ function Debugger (options) {
     self.step_manager.jumpTo(step)
   })
 
-  this.debugger.addProvider('vm', this.executionContext.vm())
-  this.debugger.addProvider('injected', this.executionContext.internalWeb3())
-  this.debugger.addProvider('web3', this.executionContext.internalWeb3())
+  this.debugger.addProvider(this.executionContext.getProvider(), this.executionContext.internalWeb3())
```
cdetrio: this is causing a bug for me.

When I patch remix-debug/src/Ethdebugger.js,

```
Ethdebugger.prototype.addProvider = function (type, obj) {
  console.log('Ethdebugger.prototype.addProvider type:', type)
  console.log('Ethdebugger.prototype.addProvider obj:', obj)
  this.web3Providers.addProvider(type, obj)
```
the type is `vm` but the obj is a web3 object.   later it tries to call `vm.on()`, but the web3 object doesn't have that method. `TypeError: this.vm.on is not a function`

cdetrio:
This is supposed to work with https://github.com/ethereum/remix/pull/1103

cdetrio:
The debugger is broken in this PR. On master the debugger has changed a lot. maybe it will work if rebased...

Issue #1198, opened 2018-03-27
New UX Theme


yann300:
we have light and dark, would be  great to come up with a third one.
https://github.com/ethereum/remix-ide/blob/master/src/app/ui/styles-guide/style-guide.js

byteknacker:
Ok, this looks like something I can understand a little bit. If I add a theme, do I just literally copy the style-guide.js and change the values of the css properties, like colour and icons to get to another theme?

yann300:
yes as a base, you can use the dark or light theme

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 50.0 DAI (50.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/remix-ide/1198/2071).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/remix-ide/1198/2071).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $47,125.86 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


**1) [jeremigendron](https://gitcoin.co/profile/jeremigendron) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ethereum/remix-ide/1198/2071?mutate_worker_action=approve&worker=jeremigendron) | [reject worker](https://gitcoin.co/issue/ethereum/remix-ide/1198/2071?mutate_worker_action=reject&worker=jeremigendron))_.**

I will forward this to @dvdptr. I may submit work on behalf of @dvdptr with his agreement.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/remix-ide/1198/2071).

JeremiGendron:
@yann300 could you review #1657 and share some feedback?

gitcoinbot:
[...]

PR #1657, opened 2019-01-04
 feat(ui): Add support for a new theme

dvdptr:
Fixes: #1198 

Please merge if OK, if you wish for the theme to be called something different or would like the theme to change, feel free to review this PR.

work done with @jeremigendron


JeremiGendron:
image

image

yann300:
thanks for the PR! ;) here some comments:

 - would it be possible to keep border on buttons? (e.g `run tab / at address button` , and `settings tab /  Plugins` )
 - some shared css were changed : righthand-panel.js, file-panel-style.js, terminal-panel-style.js . how would that impact the other themes?

JeremiGendron:
@yann300 Ok, will localize the changes to other CSS files and keep the border buttons! :+1:

yann300:
I can't see the differences on border buttons, could you double check?

src/app/panels/righthand-panel.js
```diff
@@ -141,10 +141,19 @@ const css = csjs`
   }
   .dragbar             {
     position           : absolute;
-    width              : 0.5em;
+    width              : ${
```
yann300: if you really need to have a specific value for this, that's fine no problem. 
But I would push back the original value (`0.5em` in that case) in the 2 others style.js (dark and light) instead of having a `if` here.

yann300:
thanks;),
btw why using `david` for the name, does it refer to something special?

dvdptr:
thx, no it's not a reference let's just change `david` to `cleanWhite_theme`. If you agree, I'm going to change it.

src/app/tabs/settings-tab.js
```diff
@@ -115,7 +115,7 @@ module.exports = class SettingsTab {
         </div>
         <div class="${css.crow}">
           ${self._view.theme.david}
```
yann300: is it also possible to rename underlying property/variable/filename ? (just to avoid confusion)

JeremiGendron: fixing errors and pushing

JeremiGendron: https://github.com/JeremiGendron/remix-ide/tree/patch-03 can't resolve this: changed all occurrences of "david" to clean with proper casing and renamed styleGuideDavid to styleGuideClean, still getting runtime errors.

I'm going to bed now, I am sad :(

yann300: yes but the code still contain `david` :)
could you add your commit https://github.com/JeremiGendron/remix-ide/commit/346ad992afce67f2550ab42838538d5f5b7a4580
to this PR? `git cherry-pick <commit-id>` should work
I can take a closer look too.
thanks

JeremiGendron:
Yes, I think I was still messing around to see how the storage would behave.

PR #1670, opened 2019-01-17
Record timestamps of contracts passed as arguments

LeviBarnes:
When the address of a contract is passed as an argument to a contract function, the recorder records the hard-coded address. On replay, the contracts get new addresses, so the replay fails. 

This change records the timestamp of the contract in place of the contract address. Replays involving a contract address as an argument can then be reliably replayed.


yann300:
thanks ;)

PR #1689, opened 2019-02-12
move location logic to swappanel

yann300:


src/app/components/swap-panel-component.js
```diff
@@ -6,14 +6,29 @@ var csjs = require('csjs-inject')
...
 
-    this.store.event.on('activate', (name) => { })
+    appManager.event.on('pluginNeedsLocation', (profile, domEl) => {
+      if ((profile.prefferedLocation === this.name) || (!profile.prefferedLocation && opt.default)) {
+        this.add(profile.name, domEl)
+      }
+    })
+
+    this.store.event.on('activate', (name) => {
+      let item = this.store.get(name)
```
GrandSchtroumpf: Maybe `const { profile, api } = this.store.get(name)` ?
Or `entry` instead of `item` if you don't want to use destructuration.

src/remixAppManager.js
```diff
@@ -52,24 +39,19 @@ export class RemixAppManager extends AppManagerApi {
     this.store.add(entity.profile.name, entity)
   }
 
+  // this function is only used for iframe plugins
   resolveLocation (profile, domEl) {
-    // if there's an icon, we add to the swap panel
-    // if not we suppose it just need to be put to DOM (that would be)
     if (profile.icon) {
```
GrandSchtroumpf: Maybe check on something else that the icon in the long run.

src/app/components/swap-panel-component.js
yann300: sure

PR #1691, opened 2019-02-12
Move tab logic outside of editor panel

yann300:


src/app/files/fileManager.js
```diff
@@ -101,6 +99,18 @@ class FileManager {
     return self._deps.config.get('currentFile')
   }
 
+  closeFile (name) {
+    var self = this
```
GrandSchtroumpf: I think you can use `this` with confidence here.

src/app/files/fileManager.js
```diff
@@ -101,6 +99,18 @@ class FileManager {
...
+    delete self.openedFiles[name]
+    if (Object.keys(self.openedFiles).length) {
+      self.switchFile(Object.keys(self.openedFiles)[0])
+    } else {
+      self._deps.editor.displayEmptyReadOnlySession()
+      self._deps.config.set('currentFile', '')
+    }
+    this.event.trigger('fileClosed', [name])
```
GrandSchtroumpf: Should you use `this.nodeEvent.emit` instead ?

src/app/panels/tab-proxy.js
```diff
@@ -0,0 +1,176 @@
...
+      }
+    }
+  }
+
+  addHandler (type, fn) {
+    this.handlers[type] = fn
+  }
+
+  renderTabsbar () {
+    var self = this
```
GrandSchtroumpf: You can use `this`

src/app/panels/tab-proxy.js
```diff
@@ -0,0 +1,176 @@
...
+  constructor (fileManager, editor) {
+    this.fileManager = fileManager
+    this.editor = editor
+    this.activeEntity
+    this.entities = {}
+    this.data = {}
+    this._view = {}
+
+    fileManager.event.register('fileRemoved', (name) => {
+      var filesEl = document.querySelector('#files')
```
GrandSchtroumpf: You can use `const`

src/app/panels/tab-proxy.js
```diff
@@ -0,0 +1,176 @@
...
+export class TabProxy {
+  constructor (fileManager, editor) {
+    this.fileManager = fileManager
+    this.editor = editor
+    this.activeEntity
+    this.entities = {}
+    this.data = {}
+    this._view = {}
+
+    fileManager.event.register('fileRemoved', (name) => {
```
GrandSchtroumpf: If used with `nodeEvent` use `on` instead of `register`

src/app/files/fileManager.js
yann300: yeah, fileManager is already using the remix `EventManager` for the `currentFileChanged` event. better to keep using that one for now and not mixing event...

PR #1692, opened 2019-02-13
Display plugin as a Tab

yann300:


yann300:
note that the wording `editorPanel` is not accurate anymore. will change that in subsequent PR

GrandSchtroumpf:
If you :
1 - Search for a plugin
2 - Activate a plugin
Then the search is discarded. I would remain.

yann300:
right, thanks
just added 2 more commits to it

GrandSchtroumpf:
When clicking on `Run Tests` under `Unit Testing` I've got `self.compileTab.importFileCb is not a function`

GrandSchtroumpf:
When using `Vyper` the solidity compiler tries to compile and fails.

GrandSchtroumpf:
Pipeline should open on activation (not current behavior), but vyper should not (current behavior)

GrandSchtroumpf:
When `ethdoc` is activated, an empty iframe pop at the bottom of the page. Maybe it should be `display: none` by default.

GrandSchtroumpf:
When runTab becomes active, it doesn't check if there were a contract already compiled.

yann300:
@GrandSchtroumpf added a new issue for listing all your feedback 
https://github.com/ethereum/remix-ide/issues/1705

GrandSchtroumpf:
Ok, thanks

PR #1702, opened 2019-02-19
refactor/simplify code in the tabs

iurimatias:


yann300:
just rebased

GrandSchtroumpf:
I don't really see the advantage of using more precise dependancy injection in each class, instead of injecting the `registry` only. In both cases we are in O1.

PR #1733, opened 2019-03-01
Plugin localhost

GrandSchtroumpf:
Add a form to create a local plugin

### DIFF
--- a/src/app/components/local-plugin.js
+++ b/src/app/components/local-plugin.js
@@ -1,0 +1,117 @@
+const yo = require('yo-yo')
+const modalDialog = require('../ui/modaldialog')
+
+module.exports = class LocalPlugin {
+
+  constructor() {}
+
+  /**
+   * Open a modal to create a local plugin
+   * @param {{profile: any, api: any}[]} plugins The list of the plugins in the store
+   * @returns {Promise<{api: any, profile: any}>} A promise with the new plugin profile
+   */
+  open(plugins) {
+    this.profile = JSON.parse(localStorage.getItem('plugins/local')) || { notifications: {} }
+    return new Promise((res, rej) => {
+      modalDialog('Local Plugin', this.form(plugins),
+        { fn:() => res(this.create()) },
+        { fn: () => rej() }
+      )
+    }) 
+  }
+
+  /**
+   * Create the object to add to the plugin-list
+   */
+  create() {
+    const profile = {
+      ...this.profile,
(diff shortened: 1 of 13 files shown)


src/app/components/plugin-manager-proxy.js
```diff
@@ -16,12 +16,12 @@ class PluginManagerProxy {
       registry.get('compilersartefacts').api['__last'] = new CompilerAbstract(languageVersion, data, source)
       event.trigger('sendCompilationResult', [file, source, languageVersion, data])
     }
-    instance.event.on('compilationFinished', this._listener)
+    instance.events.on('compilationFinished', this._listener)
```
yann300: if that instance is the solidity compiler, i think it implement `event`

src/app/files/file-explorer.js
```diff
@@ -200,7 +200,8 @@ function fileExplorer (localRegistry, files) {
   }
 
   // register to main app, trigger when the current file in the editor changed
-  self._deps.fileManager.event.register('currentFileChanged', (newFile, explorer) => {
+  self._deps.fileManager.events.on('currentFileChanged', (newFile) => {
```
yann300: usually we used `event` without `s` . Any specific reason for this change? 
in other case perhaps better to keep `event` (even if the `event` interface has changed)

src/app/tabs/compile-tab.js
```diff
@@ -75,13 +75,14 @@ class CompileTab {
...
 
-    this._deps.fileManager.event.register('currentFileChanged', (name) => {
+    this._deps.fileManager.events.on('currentFileChanged', (name) => {
       this.compilerContainer.currentFile = name
     })
     this.compiler.event.register('compilationFinished', (success, data, source) => {
       if (success) {
         // forwarding the event to the appManager infra
-        this.event.emit('compilationFinished', source.target, source, this.data.selectedVersion, data)
+        console.log(source.target, source, this.data.selectedVersion, data)
```
yann300: could that be removed?

src/app/components/local-plugin.js
```diff
@@ -0,0 +1,116 @@
...
+      methods: [],
+      events: []
+    }
+    if (!this.profile.name) throw new Error('Plugin should have a name')
```
yann300: these throws should be catched at some point, right? cause now it seems to silently fails (but i suppose that's fine to merge that one now for an alpha  if we are aware of the left TODOs)

src/app/files/file-explorer.js
GrandSchtroumpf: I switch to `events` because this is the interface expected by the AppManager to listen and broadcast to the other modules and plugins.
`events` is also exposed in the `Plugin` class and matchs the `Profile` interface.
Also it would help making the difference between `event` => `EventManager` and `events` => `EventEmitter`. Might be useful for the migration.

src/app/components/plugin-manager-proxy.js
GrandSchtroumpf: When testing, this instance was the `CompileTab`. I switched to `events` in `CompileTab`. But can it be also the solidity compiler from remix lib ?

src/app/components/local-plugin.js
GrandSchtroumpf: When you throw an unhandle error it goes into a console.error. But it's better if it handled indeed.
I just add handlers into the plugin-manager-component, but just a console.log. Maybe it would be better to have an alert instead of a console.log. Which method should I use for that ?

yann300: just TODO for now is fine. It could either alert or an error in the modal

src/app/components/plugin-manager-proxy.js
yann300: > but can it be also the solidity compiler from remix lib ?

 not it is not the same

src/app/files/file-explorer.js
yann300: ok

PR #1742, opened 2019-03-06
landing page styleing

LianaHus:


LianaHus:
waiting for https://github.com/ethereum/remix-ide/pull/1764

ryestew:
I switched the theme to https://bootswatch.com/4/cerulean/bootstrap.min.css  - cerulean 
I see that the div the says that it is a card ( in the homepage) has these classes:
class="card bg-primary text-success border-success p-3"
text-success is green and in cerulean - bg-primary is blue - so the text isn't legible.

Also the body text of the card is further defined in the csjs and it is:
.link_4m9f1c {
    cursor: pointer;
    background-color: var(--primary);
    color: var(--success);
    font-weight: normal;
    text-decoration: none;
}

This makes all the links green with a background color of blue. 
I think you are defining it too much.  All the stuff in the .link_4m9f1c should be left to the theme - so it isn't needed.

Also the width of the cards is not specified - so they are each incredibly wide.  This might be due to  flex-box setting.  I'm not sure why it doesn't show up in the fresca theme.

PR #1762, opened 2019-03-12
Fix autocomplete

yann300:
Fixes https://github.com/ethereum/remix-ide/issues/1760

TODOs:
 - [x] arrow selection is still buggy
 - [x] modal doesn't hide when no completion to  display <= modal hidden when backspace triggered

### DIFF
--- a/src/app/panels/terminal.js
+++ b/src/app/panels/terminal.js
@@ -13,7 +13,6 @@ var CommandInterpreterAPI = require('../../lib/cmdInterpreterAPI')
 var executionContext = require('../../execution-context')
 var Dropdown = require('../ui/dropdown')
 var AutoCompletePopup = require('../ui/auto-complete-popup')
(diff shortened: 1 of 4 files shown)


src/app/panels/terminal.js
```diff
@@ -406,14 +399,16 @@ class Terminal {
     return self._view.el
 
     function change (event) {
-      handleAutoComplete(event)
+      if (self._components.autoCompletePopup.handleAutoComplete(
+        event,
+        self._view.input.innerText)) return
```
LianaHus: ")) return" should go to the new line here, aligned with if

yann300: well `standard` doesn't seems to complain ;)

yann300: but it do complain if `return` is put in a new line (cause then it needs `{}` )

PR #1763, opened 2019-03-14
Remix plugin api

GrandSchtroumpf:
Use `ApiFactory` to manage plugins.


src/app/tabs/compile-tab.js
```diff
@@ -131,22 +138,8 @@ class CompileTab {
...
-      description: 'compile solidity contracts',
-      kind: 'compile'
-    }
-  }
-
   getCompilationResult () {
```
yann300: should this be tagged as `async` ?

src/app/tabs/compile-tab.js
```diff
@@ -131,22 +138,8 @@ class CompileTab {
...
-    return new Promise((resolve, reject) => {
-      resolve(this.compileTabLogic.compiler.lastCompilationResult)
-    })
+    return this.compileTabLogic.compiler.lastCompilationResult
```
yann300: should this be tagged as async ?

yann300:
for the browser test failure, please check if the following https://github.com/ethereum/remix-ide/pull/1763/files#diff-bd9c9dcd314f2d7df52935b3a6a4d504R505
is ok. This is where we start debugging when user click on `debug`. seems that browser test fails here...

src/app/tabs/compile-tab.js
GrandSchtroumpf: no need. With the new way of handling request (the queue) the `async` is not needed for synchronous methods.

PR #1765, opened 2019-03-15
Swap it key navigation on modal dialog and context menu style fixes

LianaHus:
also updated themes

### DIFF
--- a/src/app/panels/tab-proxy.js
+++ b/src/app/panels/tab-proxy.js
@@ -110,8 +110,8 @@ export class TabProxy {
       <div class="d-flex">
         <div class="m-1">
           <span class="p-1">
-            <i class="m-1 fa fa-plus" onclick=${increase} aria-hidden="true" title="increase editor font size"></i>
-            <i class="m-1 fa fa-minus" onclick=${decrease} aria-hidden="true" title="decrease editor font size"></i>
+            <i class="m-1 fa fa-plus" style="color: var(--dark)" onclick=${increase} aria-hidden="true" title="increase editor font size"></i>
+            <i class="m-1 fa fa-minus" style="color: var(--dark)" onclick=${decrease} aria-hidden="true" title="decrease editor font size"></i>
           </span>
         </div>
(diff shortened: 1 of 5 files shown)


ryestew:
I think our theory that the background color in the context menu would work without defining it is wrong.  Also I only see the old themes here ( the incomplete ones).  When I load 'https://stackpath.bootstrapcdn.com/bootswatch/4.3.1/slate/bootstrap.min.css' ( the slate theme) - the context menu was still transparent.  When I added the class bg-light it was light and not dark - which makes semantic sense but which also doesn't square with our theory.

The modal dialog looks good.

ryestew:
Sorry - I think I gave a not so great comment last time on the context menu:
I put in 2 different stylesheets for checking - I used https://bootswatch.com/4/materia/bootstrap.min.css as a light theme and https://stackpath.bootstrapcdn.com/bootswatch/4.3.1/darkly/bootstrap.min.css for a dark theme - maybe its easiest if you update the theme-chooser with these.

- I added the class bg-light in the container div and that will work with the themes that we will ultimately use.  Which is to say bootswatch- darkly and many of the bootswatch light themes.  
- I took the   `background-color: var(--light);` out of the container styles and put bg-light as an additional class in the container - see below:

`<div class="container_HC4Dy bg-light" style="left: 86px; top: 89px; display: block;"><ul id="menuitems"><li id="menuitemrename" class="liitem_HC4Dy">Rename</li><li id="menuitemdelete" class="liitem_HC4Dy">Delete</li></ul></div>`

For the modal dialog - I have a few comments too:
1. take out the color:var(--primary) - so that the text will be whatever the normal theme text color is 
2. in the modal header the color should be color: var(--light); not primary ( primary will be richly colored and so is the background color ( info).  This will make the X stand out more for closing
3. there is a class called undefined that gets injected into the div container of modal dialog - we should clean this up in this PR or in another. ( It may have nothing to do with this PR)
4.  in the light theme - (materia) - the buttons change the background color - not just the border color.

PR #1768, opened 2019-03-18
Bootstrap theme list

ryestew:
List of bootstrap themes and their URL's in theme-chooser.js for use in the settings tab and to change to the dark or light theme in the editor.

### DIFF
--- a/src/app/tabs/settings-tab.js
+++ b/src/app/tabs/settings-tab.js
@@ -1,10 +1,7 @@
 var yo = require('yo-yo')
-var remixLib = require('remix-lib')
-
 var tooltip = require('../ui/tooltip')
 var copyToClipboard = require('../ui/copy-to-clipboard')
 var styleGuide = require('../ui/styles-guide/theme-chooser')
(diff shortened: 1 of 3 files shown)


yann300:
could you squash it in one commit? `not quite working yet` is not really a commit here.
And remove all commented and unnecessary code

src/app/tabs/settings-tab.js
```diff
@@ -45,6 +37,25 @@ module.exports = class SettingsTab extends ApiFactory {
       location: 'swapPanel'
     }
   }
+  createThemeCheckies () {
+    let themes = styleGuide.getThemes()
+    let self = this
```
ryestew: What is the better way of dealing with the scope here?

yann300:
I think that is good to merge. please @LianaHus would you have a look?
just a quick question: what about the ace editor, do you keep the current behavior, or is this something yet to be done?

ryestew:
1 more fix to do - it is not loading the saved theme from browser storage ( after a refresh).

PR #1769, opened 2019-03-19
fix(@embark-remix): Fix embark-remix plugin

emizzle:
Fix embark-remix plugin to work with the latest version of `remixd`.

Added plugin options to allow for configurable options from `embark.json`.

Added default options. The default options use `http://localhost:8088`, which is a change from `http://localhost:8080`, as this was conflicting with IPFS.

Only start webserver to serve Embark IDE locally if a localhost is specified in the options. If a public URL is specified, it is assumed the Remix IDE is being hosted there and therefore a webserver is not needed. This allows developers to specify connecting to `https://remix.ethereum.org` if desired.


yann300:
@emizzle @iurimatias that's merged and published ;) https://www.npmjs.com/package/embark-remix

PR #1771, opened 2019-03-20
Permission Handler

GrandSchtroumpf:
Create a permission handler for modules


src/persmission-handler.js
```diff
@@ -0,0 +1,126 @@
...
+  justify-content: center;
+  align-item: center;  
+}
+.images img {
+  width: 40px;
+  height: 40px;
+}
+`
+
+export class PermissionHandler {
```
yann300: should that inherit the base class?

GrandSchtroumpf: There is no base class. It's just an example in `remix-plugin`.

Issue #1758, opened 2019-03-11
Make compiler usable from terminal


yann300:
that feature is currently broken in `swap_it`

PR #1774, opened 2019-03-20
Push the `appStore` to the terminal and add autocompletion support

yann300:
fixes https://github.com/ethereum/remix-ide/issues/1758

This PR remove the ability of accessing the compiler from the terminal.
instead of that the appStore is available from there.
at the moment the only way to accessing the API requires using `addRequest`, but that'll change in the future

### DIFF
--- a/src/app/panels/terminal.js
+++ b/src/app/panels/terminal.js
@@ -17,7 +17,7 @@ var AutoCompletePopup = require('../ui/auto-complete-popup')
 var csjs = require('csjs-inject')
 
 var css = require('./styles/terminal-styles')
-import { ApiFactory } from 'remix-plugin'
+import { BaseApi } from 'remix-plugin'
 
 var packageV = require('../../../package.json')
 
@@ -27,9 +27,18 @@ function register (api) { KONSOLES.push(api) }
 
 var ghostbar = yo`<div class=${css.ghostbar} bg-secondary></div>`
 
-class Terminal extends ApiFactory {
+const profile = {
+  displayName: 'Terminal',
+  name: 'terminal',
+  methods: [],
+  events: [],
+  description: ' - ',
+  required: false
+}
+
+class Terminal extends BaseApi {
   constructor (opts, api) {
-    super()
+    super(profile)
     var self = this
     self.event = new EventManager()
     self._api = api
@@ -115,16 +124,6 @@ class Terminal extends ApiFactory {
   }
   focus () {
     if (this._view.input) this._view.input.focus()
-  }
-  get profile () {
-    return {
-      displayName: 'terminal',
-      name: 'terminal',
-      methods: [],
-      events: [],
-      description: ' - ',
-      required: false
-    }
   }
   render () {
     var self = this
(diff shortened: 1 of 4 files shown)


yann300:
with the latest change, it allows calling a plugins/module from the terminal using this syntax:
`remix.call({name: 'solidity', key:'getCompilationResult', payload: []}).then((result) => { console.log(result) }).catch((error) => { console.log(error) })`

src/app/panels/terminal.js
```diff
@@ -17,6 +17,7 @@ var AutoCompletePopup = require('../ui/auto-complete-popup')
 var csjs = require('csjs-inject')
 
 var css = require('./styles/terminal-styles')
+import { ApiFactory } from 'remix-plugin'
```
GrandSchtroumpf: Now it's `BaseApi`

src/app/panels/terminal.js
```diff
@@ -26,8 +27,9 @@ function register (api) { KONSOLES.push(api) }
 
 var ghostbar = yo`<div class=${css.ghostbar} bg-secondary></div>`
 
-class Terminal {
+class Terminal extends ApiFactory {
   constructor (opts, api) {
+    super()
```
GrandSchtroumpf: The profile should be inside `super(profile)`

src/app/panels/terminal.js
```diff
@@ -103,14 +105,27 @@ class Terminal {
...
+  get profile () {
+    return {
+      displayName: 'terminal',
+      name: 'terminal',
+      methods: [],
+      events: [],
+      description: ' - ',
+      required: false
+    }
+  }
```
GrandSchtroumpf: Create const variable outside the class, and add it inside the super() : 
```javascript
const profile = {
      displayName: 'terminal',
      name: 'terminal',
      methods: [],
      events: [],
      description: ' - ',
      required: false
}
....
constructor(opts, api) {
  super(profile)
  ...
}
```

src/app/ui/auto-complete-popup.js
```diff
@@ -161,6 +164,12 @@ class AutoCompletePopup {
           })
         }
       })
+      this.extraCommands.forEach(item => {
+        let command = getKeyOf(item)
```
GrandSchtroumpf: You can use `const` here

src/app/ui/auto-complete-popup.js
```diff
@@ -188,6 +197,21 @@ class AutoCompletePopup {
     this._selectedElement = 0
     yo.update(this._view, this.render())
   }
+
+  extendAutocompletion () {
+    // TODO: this is not using the appManager interface. Terminal should be put as module
+    this.opts.appStore.event.on('activate', (id) => {
+      let keyValue = {}
```
GrandSchtroumpf: you can use `const` because object are mutable.

src/lib/cmdInterpreterAPI.js
```diff
@@ -40,6 +41,9 @@ class CmdInterpreterAPI {
       'remix.debugHelp()': 'Display help message for debugging'
     }
   }
+  async call (message) {
+    return await this._components.terminal.externalApi.request(message)
```
GrandSchtroumpf: if `this._components.terminal.externalApi.request(message)` is a Promise, you don't need `async/await` here. It's redoundant.

src/app/panels/terminal.js
LianaHus: please always use uppercase for displayName. terminal->Terminal

yann300:
updated with some others changes @GrandSchtroumpf

PR #1779, opened 2019-03-22
Bootstrap swapit header

ryestew:
Works / looks best in cerulean & darkly themes. Improves semantic html of header and adds scroll room to bottom of the swap_it panel div.


yann300:
you did this PR against master but I think you meant to do it against `swap_it`  right?

PR #1786, opened 2019-03-27
Style fixes

### LINKED ISSUE #1785: terminal style is not complete
Screenshot from 2019-03-27 12-51-21

LianaHus:
- restyling and implementing of Landing page
- style fixes in terminal
- style fixes in File Explorer

Screenshot from 2019-03-27 14-29-55

### DIFF
--- a/src/app/files/file-explorer.js
+++ b/src/app/files/file-explorer.js
@@ -429,7 +429,7 @@ fileExplorer.prototype.packageFiles = function (filesProvider, callback) {
     else {
       async.eachSeries(Object.keys(files), (path, cb) => {
         filesProvider.get(path, (error, content) => {
-          if (/^\s+$/.test(content)) {
+          if (/^\s+$/.test(content) || !content.length) {
             content = '// this line is added to create a gist. Empty file is not allowed.'
           }
           if (error) cb(error)
--- a/src/app/files/file-explorer.js
+++ b/src/app/files/file-explorer.js
@@ -517,7 +517,7 @@ fileExplorer.prototype.renderMenuItems = function () {
         `
       } else {
         return yo`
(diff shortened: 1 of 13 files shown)


ryestew:
A small addition is: Around the select box - the bounding box is not there - if you add the class form-control - it comes back but then the text is too big - so you'd need to add a font-size.  Also the width is too wide. I'm not sure why its 30ch wide.  I've never used ch before - looks interesting though.

Also maybe the bar should be thicker 5em's instead of 3em.

Here are some comments about the landing page:

- the remix logo isn't working where it is and I think its taking up too much room.  We need a horizontal logo - not a vertical one.  I'm not sure if we even need a logo at all.
- I think all the card in the landing page should be the same height or a minimum height of 210px.
- Also if the swap panel is showing then it scrolls horizontally - which isn't ideal.
- Maybe the workspaces should be on their own line.  Making them green attracts attention but it seems like too big a gesture.  Maybe having them on their own line is enough.

LianaHus:
the style for landing page is not final. it will look quite different. thanks for comments, will take into account!.

will try to change select box style.

src/app/files/file-explorer.js
```diff
@@ -516,14 +517,12 @@ fileExplorer.prototype.renderMenuItems = function () {
         `
       } else {
         return yo`
-        <span onclick=${(event) => { event.stopPropagation(); this[ action ]() }} class="newFile ${css.newFile}" title=${title}>
-          <i class=${icon}></i>
-        </span>
+        <span onclick=${(event) => { event.stopPropagation(); this[ action ]() }} class="${icon} ${css.newFile}" title=${title}></span>
```
yann300: for the sake of browser test :) we should keep the class here `newFile`. will do the edit

PR #1803, opened 2019-04-04
test, static analysis tab && default minwidth for panel

yann300:
- improve test tab && static analysis tab
 - decrease minwidth for swap panel
 - decrease minwidth for terminal
 - focus on terminal at load

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -170,7 +170,7 @@ class App extends ApiFactory {
 
   init () {
     var self = this
-    self._components.resizeFeature = new PanelsResize('#swap-panel', '#editor-container', { 'minWidth': 400, x: 450 })
+    self._components.resizeFeature = new PanelsResize('#swap-panel', '#editor-container', { 'minWidth': 300, x: 300 })
     run.apply(self)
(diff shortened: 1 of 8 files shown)


ryestew:
In static analysis - the div.buttons - should probably be a nav element styled like the Active / Inactive in the plugins module.

in the static analysis -there can be 3 scroll bars - which is confusing. The problem is that the max-height for the containing divs is too high - too high that is for my screen.  When both are max-height:300px and not 400px - there are only 2 scrollbars and I don't panic. ( and no side by side scrollbars).

The only way to make this more custom than making this work for my screen is by grabbing the window height and doing some math...I think.

Also the checkboxes need to have a look to separate them from above and below - the above will probably be fixed by the nav ( see above) and the below - a border or making this a card or a some other bootstrap component.  

**But for the moment - just make the max-heights shorter and we can style further later.**

LianaHus:
Swap panel will be open. We have the change for that but it comes with my pr

yann300:
ok I will change the `max-height`

PR #1821, opened 2019-04-09
Un-mapping brace shortcut Cmd-L which clobbers with browser shortcut 

### LINKED ISSUE #76: Editor: cmd + L clobbers browser shortcut
On OSX + Chrome, go-to-line is mapped to `Cmd + L` which is a much-used shortcut in the browser to move the cursor to the address bar. I have seen most editors avoid overriding this shortcut and instead using something like `Cmd + Ctrl + G` or `Cmd + Alt + G` (for **go** to line).

ckshei:
Fixes #76

### DIFF
--- a/src/app/editor/editor.js
+++ b/src/app/editor/editor.js
@@ -99,8 +99,12 @@ class Editor {
     this.editor = ace.edit(el)
     ace.acequire('ace/ext/language_tools')
 
-    // Unmap ctrl-t & ctrl-f
-    this.editor.commands.bindKeys({ 'ctrl-t': null })
+    // Unmap ctrl-t & ctrl-f & ctrl-l & cmd-l
+    this.editor.commands.bindKeys({
+      'ctrl-t': null,
+      'ctrl-L': null,
+      'Command-L': null
+    })
 
     // shortcuts for "Ctrl-"" and "Ctrl+"" to increase/decrease font size of the editor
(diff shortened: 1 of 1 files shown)


src/app/editor/editor.js
```diff
@@ -99,8 +99,12 @@ class Editor {
     this.editor = ace.edit(el)
     ace.acequire('ace/ext/language_tools')
 
-    // Unmap ctrl-t & ctrl-f
-    this.editor.commands.bindKeys({ 'ctrl-t': null })
+    // Unmap ctrl-t & ctrl-f & ctrl-l & cmd-l
+    this.editor.commands.bindKeys({
+      'ctrl-t': null,
```
LianaHus: could you please remove ctrl-t from the list. Currently, it's not being used by Ace editor. thanks

LianaHus: @ckshei

ckshei: @LianaHus - Yup I can remove ctrl-t. On brace it's mapped as "transposeletters" not sure what that does but assuming that it's fine? I'll also update the comment 

```    name: "transposeletters",
    bindKey: bindKey("Ctrl-T", "Ctrl-T"),
    exec: function(editor) { editor.transposeLetters(); },
    multiSelectAction: function(editor) {editor.transposeSelections(1); },
    scrollIntoView: "cursor"```

PR #1837, opened 2019-04-09
Small fixes on receiving new block

yann300:


src/app/execution/txLogger.js
```diff
@@ -178,7 +178,7 @@ class TxLogger {
     })
 
     this._deps.txListener.event.register('newBlock', (block) => {
-      if (!block.transactions.length) {
+      if (!block.transactions || block.transactions && !block.transactions.length) {
```
GrandSchtroumpf: Could you add parenthesis for better readibility : `if (!block.transactions || (block.transactions && !block.transactions.length))`

yann300: are you sure? `console.log(!![])` return true

GrandSchtroumpf: Sorry, you're right...
To have only one boolean operation you would have to do : 
```javascript
if (block.transactions && block.transactions.length) return
this.logEmptyBlock({ block })
```

Issue #1835, opened 2019-04-09
Implement remix-plugin@0.0.2-alpha.2


GrandSchtroumpf:
- [x] Change `ApiFactory` to `BaseApi`
- [x] Change `SourceHighlighter` to `Editor`
- [x] Remove `TxListenerModule` and add it inside `UdappModule`
- [x] Split `App` classes into two classes : `App` and `NetworkModule`

Update methods/events names for : 
- [x] `Network` 
- [x] `Editor`
- [x] `Compiler`
- [x] `FileSystem`
- [x] `Udapp`
- [x] `Theme`

yann300:
@GrandSchtroumpf i think that can be closed isn't it?

GrandSchtroumpf:
Yes thx

PR #1838, opened 2019-04-09
Issue#1835 remix plugin 0.0.2

GrandSchtroumpf:
**Ready to merge**
Implement remix-plugin 0.0.2.
See : https://github.com/ethereum/remix-ide/issues/1835


src/app/editor/SourceHighlighters.js
```diff
@@ -1,24 +1,22 @@
 'use strict'
 const SourceHighlighter = require('./sourceHighlighter')
 
-import { ApiFactory } from 'remix-plugin'
+import { EditorApi } from 'remix-plugin'
 
-class SourceHighlighters extends ApiFactory {
+const profile = {
+  displayName: 'source highlighters',
```
yann300: `displayName` and `name` are not necessary right? they will be overridden by the `super` of `EditorApi`

GrandSchtroumpf: No, only the methods, events and kind are going to be overrided.

yann300: oh ok, then I think we should have a default naming for API shared among different platform. That does not need to be completely typescripted but at least well documented.
i.e if a file system is usually accessible under `call('fileSystem', '<function>', [])` , others platform have to provide it under the same name `fileSystem`.

in a further enhancement, we could add a function for retrieving modules of a specific `kind`.

GrandSchtroumpf: Sounds fair, but how you would deal with `compiler` ? You'll have the same `kind` with the same `methods` and `events`, but the names are different (`solidity`, `vyper`, `lll`, ...).

yann300: That would only apply with the core API like file manager, editor, context, udapp

yann300: That would only apply with the core API like file manager, editor, context, udapp

yann300: That would only apply with the core API like file manager, editor, context, udapp

GrandSchtroumpf: So the core API would be the one that are unique ?
Like : 
- core : `fileSystem`, `editor`, `network`, `udapp`.
- extension : `compiler`, `test`, `analysis`, `vm` (?)

yann300: Why 'vm' ? 
Yes, The core API provides infrastructure for module/plugin to interact with.

GrandSchtroumpf: Well, I guess at some point we would like to support eWASM Virtual Machine for example.

Issue #1828, opened 2019-04-09
file explorers


ryestew:
The file explorers should have a padding-left ( or margin-left) of 15px or so to left align under the title of the swap panel.  There is also a div in the file explorers with a  class of undefined.

yann300:
@ryestew @LianaHus can that be closed?

LianaHus:
yes this is already fixed

Issue #1827, opened 2019-04-09
Solidity Icon


ryestew:
Needs to not be so thin and tall - so it is legible at 28x28px

Issue #1826, opened 2019-04-09
Icon Panel Icons & Background


ryestew:
Icon Panel needs to be either be:
-light so the icons can stay dark
or
- the svg's need to change to an appropriate color if the background changes.  This is the more complicated option.  Also check if the color change can work with icons with multiple colors or with gradients ( like Viper).

PR #1845, opened 2019-04-11
Feedback fixes

LianaHus:
https://github.com/ethereum/remix-ide/issues/1827
https://github.com/ethereum/remix-ide/issues/1826
https://github.com/ethereum/remix-ide/issues/1824
https://github.com/ethereum/remix-ide/issues/1828
partially https://github.com/ethereum/remix-ide/issues/1849 (only for Deploy tab)
https://github.com/ethereum/remix-ide/issues/1850


src/app.js
```diff
@@ -505,6 +505,19 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
     }
   }, false)
 
+  window.addEventListener('keypress', (e) => {
```
yann300: That code should be put in a separate handler (class), so we don't spam `app.js`.
can be in another PR though

src/app/components/vertical-icons-component.js
```diff
@@ -57,7 +58,10 @@ class VerticalIconComponent {
    * @param {ModuleProfile} profile The profile of the module
    */
   addIcon ({kind, name, icon, displayName}) {
-    this.icons[name] = yo`<div class="${css.icon}" onclick="${(e) => { this._iconClick(name) }}" plugin="${name}" title="${displayName || name}" ><img src="${icon}" alt="${name}" /></div>`
+    this.icons[name] = yo`
+      <div class="${css.icon}" onclick="${(e) => { this._iconClick(name) }}" plugin="${name}" title="${displayName}" >
```
yann300: on the diff, the previous version is `title="${displayName || name}"`

src/app/components/vertical-icons-component.js
```diff
@@ -69,15 +73,25 @@ class VerticalIconComponent {
...
       let key = helper.checkSpecialChars(status.key) ? '' : status.key
       let type = helper.checkSpecialChars(status.type) ? '' : status.type
       let title = helper.checkSpecialChars(status.title) ? '' : status.title
-      el.appendChild(yo`<i title="${title}" class="fa fa-${key} ${css.status} text-${type}" aria-hidden="true"></i>`)
+      el.appendChild(yo`<span title="${title}" class="fa fa-${key} ${css.status} text-${type}" aria-hidden="true"></span>`)
+
+      // el.classList = "" doesn't work on all browser use instead
+      var classList = el.classList
+      while (classList.length > 0) {
+        classList.remove(classList.item(0))
```
yann300: i guess you should remove all the classes here right?
`border-${type}` is not removed here

src/app/components/vertical-icons-component.js
```diff
@@ -206,4 +228,15 @@ const css = csjs`
     bottom: 0;
     right: 0;
   }
+  .statusWithBG
```
yann300: that one seems not used

src/app.js
LianaHus: ok, will do

src/app/components/vertical-icons-component.js
LianaHus: I know. I thought we should not show name. The description should be provided, if not we don't show anything. no?

src/app/components/vertical-icons-component.js
LianaHus: I know. I have it, for now, we will discuss this in a call. this is the other way to show status. The Circle with the number of errors/warning instead of border and icon. I'll remove either one of them

src/app/components/vertical-icons-component.js
yann300: I think we have to default to something

src/app/components/vertical-icons-component.js
yann300: hmm ok sorry for that please scratch that comment ;)

LianaHus: :)

ryestew:
It's cool you found border-warning etc.  The border outline looks nice and elegant.  But having 2 selected is confusing.  Either just have 1 for the selected or don't use them at all and use a rectangle like in gitter would be better for identifying the active.  Although that will involve restructuring the html of the icon panel.  

The use of filters is great.  They need a bit of tweaking to work well in both light and dark themes.  Maybe use opacity... or maybe grab the dark or light theme: quality ( or in the html - style="--theme:dark;" or (light) to make the active or even the inactive icon pop more relative to the current theme.
Almost there...almost

LianaHus:
thanks for review!
you are right about the status, it should be shown with circle as discussed today, but this is another PR
I didn't get about filter. opacity will not work. you want to have more difference between active and not active one?

src/app/components/vertical-icons-component.js
LianaHus: ok

ryestew:
The contrast was a problem in some themes and not others.

LianaHus:
> The contrast was a problem in some themes and not others.

currently, it is 0.5. the problem is not all dark themes have light enough color for icon panel. we 1) either remove part of them(which I guess not good idea because currently, we have just a couple of dark themes) 2) have a fixed color of icon panel for dark and light themes 3) or leave it like now. 

please feel free to share other ideas if you have. maybe some corrections for some theme... or more details. we can have a call for this.

ryestew:
The compile tab has 2 scroll bars - side by side.  Well they are side by side at the top of the panel but the inside one is only for the checkboxes.  The side by side scroll bars-in general- are a sign of a UX problem.  In this case it isn't the worst.  So the current is an OK temporary solution.
Here are a couple of solutions:
1. Make the top div with the checkboxes smaller with a max-height: of 200 or 150px and grab the height of the screen with JS and do the math so that max-height of the lower results is the remainder of the screen size - and make that the max-height.
2. Make the checkboxes or both for that matter with a bootstrap collapse -so that by default you only see the results - but you could view the checkboxes by clicking on a button.  The idea is that you'd only see one or the other.

LianaHus:
I guess the last comment was for another issue. I'll copy it there

Issue #1799, opened 2019-04-03
font awesome 5


ryestew:
Load font-awesome 5 icons - assuming that the new icons won't mess up the font-awesome 4 icons.

PR #1855, opened 2019-04-16
Fa5

ryestew:
fixes https://github.com/ethereum/remix-ide/issues/1799
Adding fa5 icons.  Need to be checked for disappearing fa4 fonts that disappear.  I think I caught them all.  The fa4 shim ( to make any left over fa4 icons appear) has not been loaded because I'd like to catch or at least try to catch all the fa4 icons.  Fa4 icons won't appear if their name has been changed like fa-close is now fas-times.


yann300:
that looks good ;)
tests are failing but i am not sure about the reason ...
some icon are not displayed, so far I could identify
`fas fa-files-o` and `fas fa-github`

yann300:
and `fa-pencil-square-o` too

yann300:
and `fa-floppy-o` , that's why browser test are currently failing

LianaHus:
FE: publish to gist is broken
FE: copy All files is missing
Home: See all plugins icon is still fa fa-plug is this how it should be?
Deploy and run: save transaction is missing

LianaHus:
what about  far?
far fa-copy
far fa-folder-open

PR #1858, opened 2019-04-16
Remove "listen on network" dropdown list

yann300:


LianaHus:
what about improving search?

LianaHus:
https://github.com/ethereum/remix-ide/issues/1871

yann300:
> what about improving search?

should already work

PR #1879, opened 2019-04-18
error msg after loading gist when input has no gist id

LianaHus:
fixes https://github.com/ethereum/remix-ide/issues/1875
fixes https://github.com/ethereum/remix-ide/issues/1868

### DIFF
--- a/src/lib/gist-handler.js
+++ b/src/lib/gist-handler.js
@@ -12,11 +12,13 @@ function GistHandler (_window) {
     var gistId
     if (params['gist'] === '') {
       loadingFromGist = true
-      modalDialogCustom.prompt(null, 'Enter the ID of the Gist you would like to load.', null, (target) => {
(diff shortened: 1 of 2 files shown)


src/lib/gist-handler.js
```diff
@@ -12,11 +12,13 @@ function GistHandler (_window) {
     var gistId
     if (params['gist'] === '') {
       loadingFromGist = true
-      modalDialogCustom.prompt(null, 'Enter the ID of the Gist you would like to load.', null, (target) => {
+      modalDialogCustom.prompt(null, 'Enter the ID of the Gist or URL you would like to load.', null, (target) => {
```
yann300: not sure `URL` works in that case, are you sure of this change?

LianaHus: works for me

Issue #1849, opened 2019-04-12
Swap panel inner layout


LianaHus:
all tabs should fit in min-width of swap panel.
should use % to resize with swap's width

currently Deploy and Viper look bad in min-width of Swap panel

LianaHus:
https://github.com/ethereum/remix-ide/pull/1845

LianaHus:
@ryestew could you please link with you pr so we can close this after merge. TIA!

yann300:
@ryestew @LianaHus  could you link the PR?

yann300:
@LianaHus @ryestew  can we close it?

LianaHus:
if Rob is done with it I guess yes.

PR #1886, opened 2019-04-23
Bootstrap runtab aliscious

ryestew:


ryestew:
Still to be done:
Maybe update the alert for a card 
and...

yann300:
Could we close the topic run tab UX for now or are there some other changes to apply?

ryestew:
A few small updates to this PR coming in a half hour.

ryestew:
Last update to accommodate long input names in the multi-param manager ( the accordion to input 1 param at a time).  The deployed instance is still an alert - not a card - so it has a background color.  Maybe it should be a card.  Inside the multi-param-manager - the bg-color is light.

ryestew:
this is associated with #1849

Issue #1878, opened 2019-04-18
On run/deploy tab - add a line


ryestew:
Take out the info about the provider on the right side of the Environment pull down menu and put it in its own line labeled: Provider.  Also remove the plugin icon.  The provider info will not be editable so it won't go in an input box.

PR #1888, opened 2019-04-25
remove the network info from Environment pulldown

ryestew:
Addresses issue: On run/deploy tab - add a line #1878 
I'm not sure if the word "network" is necessary.

### DIFF
--- a/src/app/tabs/runTab/settings.js
+++ b/src/app/tabs/runTab/settings.js
@@ -7,18 +7,25 @@ const copyToClipboard = require('../../ui/copy-to-clipboard')
 const modalDialogCustom = require('../../ui/modal-dialog-custom')
 const addTooltip = require('../../ui/tooltip')
 const helper = require('../../../lib/helper.js')
+const globalRegistry = require('../../../global/registry')
 
 class SettingsUI {
 
   constructor (settings) {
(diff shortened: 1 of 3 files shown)


src/app/tabs/runTab/settings.js
```diff
@@ -68,7 +67,15 @@ class SettingsUI {
         </div>
       </div>
     `
-
+    var networkEl = yo`
```
GrandSchtroumpf: You can use `const` here, and for the next elements.

src/app/tabs/runTab/settings.js
```diff
@@ -231,7 +246,8 @@ class SettingsUI {
         this.netUI.innerHTML = 'can\'t detect network '
         return
       }
-      this.netUI.innerHTML = `<i class="${css.networkItem} fas fa-plug" aria-hidden="true"></i> ${name} (${id || '-'})`
+      let network = this._deps.networkModule.getNetworkProvider
```
yann300: shouldn't this be `this._deps.networkModule.getNetworkProvider()`

yann300: oh ok i see,  you are using `network()`

PR #1891, opened 2019-04-26
new status icons

LianaHus:


src/app/components/vertical-icons-component.js
```diff
@@ -52,7 +52,14 @@ class VerticalIconComponent {
   listenOnStatus (api) {
     if (!api.events) return
     const fn = (status) => {
-      this.setIconStatus(api.profile.name, status)
+      if (
```
yann300: why do we need to test the inputs before calling `setIconStatus` ?

LianaHus: to limit status types

PR #1899, opened 2019-04-30
Fix the context view

yann300:
Fix the panel showing info about the code the user is currently focusing on


LianaHus:
badge badge-info font-weight-bold
for context view

PR #1902, opened 2019-05-02
moving wei over - but not far enough

ryestew:


ryestew:
This PR is not quite ready for merging - the wei pulldown needs to be moved a bit closer to the value input.

ryestew:
Latest update has the wei pulldown in the right place - but the scaling is back to 100%.  Which gets rid of the question why it was small in the first place.

PR #1906, opened 2019-05-03
Update package.json

yann300:
workaround for https://github.com/indexzero/http-server/pull/523


yann300:
we'll swicth back to `http-server` when it's fixed

PR #1916, opened 2019-05-03
Improve solidity context view

yann300:
after discussing with @LianaHus @GrandSchtroumpf


LianaHus:
looks very good. I would suggest to always show it and add a column number for the current position. what do you think?

PR #1919, opened 2019-05-06
Fix config

yann300:
that will allow switching 0.8 / 0.7.x without loosing configuration data

### DIFF
--- a/src/app/tabs/settings-tab.js
+++ b/src/app/tabs/settings-tab.js
@@ -35,11 +35,6 @@ module.exports = class SettingsTab extends BaseApi {
       }
     } /* eslint-enable */
     this.event = new EventManager()
-    this.initTheme()
-  }
-
-  initTheme () {
-    this.currentTheme = this._deps.themeModule.currentTheme()
   }
 
   createThemeCheckies () {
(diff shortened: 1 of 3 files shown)


src/app/tabs/settings-tab.js
```diff
@@ -54,7 +54,7 @@ module.exports = class SettingsTab extends BaseApi {
           <input type="radio" onchange=${event => { onswitchTheme(event, aTheme.name) }} class="align-middle form-check-input" name="theme" id="${aTheme.name}"   >
           <label class="form-check-label" for="${aTheme.name}">${aTheme.name} (${aTheme.quality})</label>
         </div>`
-          if (this.currentTheme === aTheme.name) el.querySelector('input').setAttribute('checked', 'checked')
+          if (this.currentTheme.name === aTheme.name) el.querySelector('input').setAttribute('checked', 'checked')
```
GrandSchtroumpf: You don't use other things than the name in `currentTheme`. Maybe you can directly do : 
```
if (this._deps.themeModule.active === aTheme.name)
```
Like that you don't need the `currentTheme` anymore in the setting tab.

yann300: ok

PR #1921, opened 2019-05-07
[WIP] Home Icon

LianaHus:


src/app/ui/landing-page/landing-page.js
```diff
@@ -77,8 +77,9 @@ const profile = {
...
+  location: 'mainPanel',
+  kind: 'home'
```
yann300: we don't need anymore this right?

LianaHus: no

LianaHus: removed

PR #1933, opened 2019-05-07
Fix vyper plugin

yann300:
- use remix-plugin 0.0.2-alpha.6
 - add events for local plugin


src/app/components/local-plugin.js
```diff
@@ -37,9 +37,9 @@ module.exports = class LocalPlugin {
...
       methods: [],
-      events: [],
       hash: `local-${this.profile.name}`,
-      location: 'swapPanel'
+      location: 'swapPanel',
+      events: ['compilationFinished']
```
GrandSchtroumpf: This should not be hardcoded. It's only useful for the vyper compiler.

src/app/components/local-plugin.js
```diff
@@ -117,6 +124,10 @@ module.exports = class LocalPlugin {
         <label for="plugin-url">Url <small>(required)</small></label>
         <input class="form-control" onchange="${e => this.updateUrl(e)}" value="${url}" id="plugin-url" placeholder="ex: https://localhost:8000">
       </div>
+      <div class="form-group">
+        <label>Events <small>(comma separated)</small></label>
+        <input class="form-control" onchange="${e => this.updateEvents(e)}" value="${events}" id="plugin-events" placeholder="ex: compilationFinished,configurationChanged">
+      </div>
```
GrandSchtroumpf: We should use a list of input instead. With a button to add a new input field.

GrandSchtroumpf: Something like that : 
```javascript
const eventsForm = (events) => events.map((event, i) => {
  yo`<input class="form-control" onchange="${e => this.updateEvents(e, i)}" value="${event}" />
})
const eventsEl = eventsForm(this.profile.events)
const pushEvent= () => {
  this.profile.events.push('')
  yo.update(eventsEl, eventsForm(this.profile.events)
} 
const addEvent = yo`<button onclick="${() => pushEvent()}">Add an event</button>`
```

PR #1950, opened 2019-05-09
update local plugin form for location

ryestew:
The form does not have validation errors (though the errors will appear in the console).  If the location is none - then the location is location:none. ( it works but I'm not sure if this is OK).


GrandSchtroumpf:
I guess "none" is ok because it's not "mainPanel", "swapPanel". But it would be better to have real support for "none" at some point I guess

PR #1957, opened 2019-05-10
ask permission to injected provider

yann300:


yann300:
this commit is meant to be deployed in next intermediate release 0.7.7

Issue #1934, opened 2019-05-07
Check if plugin name is not used yet


GrandSchtroumpf:
fix https://github.com/ethereum/remix-ide/pull/1969

When using a local plugin, we should check if the name is not used yet.
Either we check it [here](https://github.com/ethereum/remix-ide/blob/swap_it/src/app/components/local-plugin.js#L68) and display an error under the input.
Or/And [here](https://github.com/ethereum/remix-ide/blob/swap_it/src/app/components/plugin-manager-component.js#L110) and we display a toaster error.
To check if the name is already used do : 
```javascript
appStore.ids.includes(name)
```

PR #1969, opened 2019-05-14
Plugin put in

ryestew:
update to set location with local storage and a fix if no local storage is set for the plugin


ryestew:
wait to merge this until I fix the issue https://github.com/ethereum/remix-ide/issues/1934

src/app/components/local-plugin.js
```diff
@@ -34,11 +34,13 @@ module.exports = class LocalPlugin {
...
       methods: [],
+      location: 'swapPanel',
+      ...this.profile,
       hash: `local-${this.profile.name}`
     }
+    profile.events = profile.events || []
     profile.events = profile.events.filter((item) => { return item !== '' })
```
GrandSchtroumpf: You can do : 
```javascript
profile.events = (profile.events || []).filter(item => item !== '')
```

src/app/components/local-plugin.js
```diff
@@ -120,6 +122,20 @@ module.exports = class LocalPlugin {
         return yo`<input class="form-control" onchange="${e => this.updateEvents(e, i)}" value="${event}" />`
       })}</div>`
     }
+    const radioLocations = (label, displayN) => {
+      const radioButton = (this.profile.location === label)
```
GrandSchtroumpf: You can just do : 
```javascript
return (this.profile.location === label) ? ... : ...
```

ryestew:
Check if plugin name is not used yet #1934

Issue #1970, opened 2019-05-14
Allow switching to 0.7.7


yann300:
fix https://github.com/ethereum/remix-ide/pull/1971

PR #1971, opened 2019-05-14
switch to previous version

yann300:
and from the previous version which back to the latest
fix https://github.com/ethereum/remix-ide/issues/1970


index.html
```diff
@@ -35,6 +35,36 @@
...
+					let versions = {
+						'0.7.7': 'assets/js/0.7.7/app.js',
+						'0.8.0': 'build/app.js'
+					}
+					let version = urlParams().appVersion ? urlParams().appVersion : '0.7.7'
+					let app = document.createElement('script')
+					app.setAttribute('src', versions[version])
+					document.body.appendChild(app)
+				}				
+		</script>
```
LianaHus: you have tabs instead of spaces

ryestew:
It works like magic!  I'd recommend taking out the remix logo in the homepage and changing the padding in thisJumbotron: 2.5rem 0rem;  

And the buttons will be on the same level as the headline "The new layout ..."

yann300:
the rest is going to be done in another PR

PR #1980, opened 2019-05-16
Fix highlight reference

yann300:


src/app/editor/contextualListener.js
```diff
@@ -115,10 +116,16 @@ class ContextualListener {
     let lastCompilationResult = this._deps.compilersArtefacts['__last']
     if (lastCompilationResult && lastCompilationResult.languageversion.indexOf('soljson') === 0) {
       let lineColumn = this._deps.offsetToLineColumnConverter.offsetToLineColumn(position, position.file, lastCompilationResult.getSourceCode().sources, lastCompilationResult.getAsts())
-      let css = 'highlightreference'
+      const css = csjs`
+        .highlightref_fullLine {
+          position:absolute;
+          z-index:20;
```
GrandSchtroumpf: You should use a "real" value like "1" or "2" in some rare cases.

src/app/editor/contextualListener.js
```diff
@@ -115,10 +116,16 @@ class ContextualListener {
     let lastCompilationResult = this._deps.compilersArtefacts['__last']
     if (lastCompilationResult && lastCompilationResult.languageversion.indexOf('soljson') === 0) {
       let lineColumn = this._deps.offsetToLineColumnConverter.offsetToLineColumn(position, position.file, lastCompilationResult.getSourceCode().sources, lastCompilationResult.getAsts())
-      let css = 'highlightreference'
+      const css = csjs`
+        .highlightref_fullLine {
+          position:absolute;
```
GrandSchtroumpf: When using absolute, most of the time you want to specify the top / left. At least to make it explicit.

src/app/editor/contextualListener.js
```diff
@@ -132,7 +139,7 @@ class ContextualListener {
       }
       const fileName = lastCompilationResult.getSourceName(position.file)
       if (fileName) {
-        return this.editor.addMarker(lineColumn, fileName, css)
+        return this.editor.addMarker(lineColumn, fileName, css.highlightref_fullLine.className)
```
GrandSchtroumpf: Why not just writing "highlightref_fullLine" ?

src/app/editor/contextualListener.js
yann300: did the changes. except this one, anyway ace editor is handling these properties

PR #1983, opened 2019-05-17
learn more button goes to docs & css adjustments to jumbotron

ryestew:
Do you think it is better to have a space between the buttons?  The link in learn more doesn't go to the master docs but to a branch.  This URL needs to get updated.


src/app/ui/landing-page/landing-page.js
```diff
@@ -183,15 +182,12 @@ export class LandingPage extends BaseApi {
...
-          <div class="${css.logoContainer}">
-            <img id="remixLogo" src="${logo}" alt="remix logo">
-          </div>
           <div class="${css.headlineContainer}">
             <h2 class="">The new layout has arrived</h2>
           </div>
           <div class="${css.jumboBtnContainer}">
             <button class="btn btn-primary btn-lg" href="#" onclick=${() => { learnMore() }} role="button">Learn more</button>
-            <button class="btn btn-secondary btn-lg" onclick=${() => { switchToPreviousVersion() }}>Switch to the previous version</button>
+            <button class="btn btn-secondary btn-lg" onclick=${() => { switchToPreviousVersion() }}>Use old layout</button>
```
GrandSchtroumpf: I'm not that found of "Use old Layout". For me it sounds like "use the very old ugly layout if you have no taste". Also this new version is not only about layout.

ryestew: To make a list of choices for the text:
1. Use previous version
2. Use old version
3. Load previous version
4. Load version for those who don't like change
5. Load "remix classic"
6. Revert to previous version.
7. Revert to old Remix

yann300: +1 for 1.

PR #1984, opened 2019-05-17
Update Vyper Plugin, and enables switch file for FileManager

### LINKED ISSUE #1978: FileManager should expose a "setCurrentFile" method
Right now we can getFile, setFile and getCurrentFile, but we should also be able to set the current file like that plugin could focus the attention of the user when they "setFile".

GrandSchtroumpf:
fix #1978

### DIFF
--- a/src/app/files/fileManager.js
+++ b/src/app/files/fileManager.js
@@ -22,7 +22,7 @@ const profile = {
 
 // File System profile
 // - events: ['currentFileChanged']
-// - methods: ['getFolder', 'getCurrentFile', 'getFile', 'setFile']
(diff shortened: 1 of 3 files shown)


src/app/files/fileManager.js
```diff
@@ -114,10 +114,7 @@ class FileManager extends FileSystemApi {
   }
 
   async getCurrentFile () {
-    const path = this.currentFile()
-    if (!path) throw new Error('no file selected')
```
yann300: have you checked if catching the error was not handled somewhere? removing this might change the behavior of who is using `getCurrentFile`

GrandSchtroumpf: oups, I forget to add it back. It's corrected now.

PR #1985, opened 2019-05-17
Link2docs

ryestew:
This PR has to do with: https://github.com/ethereum/remix-ide/issues/1876
*NOTE:*  I separated the root of the docs from the page of the docs.  
https://github.com/ethereum/remix-ide/pull/1985/files#diff-7fcd623a8b34472b25b227bda0cd7960R69
This is only a temporary solution so that we can point the link to a working branch of the docs.  Once the new docs get merged - then separating the page from the URL of the folder should be removed because external plugins won't be in our docs.


yann300:
Awesome! getting closer ;)

yann300:
i'd see more `fas fa-info` for the icon and using the standard cursor pointer

Issue #1951, opened 2019-05-09
MythX integration


yann300:
This issue aim to move https://github.com/ethereum/remix-ide/pull/1883
to an infrastructure using remix plugin infrastructure

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 400.0 DAI (400.0 USD @ $1.0/DAI)  attached to it as part of the Ethereum Foundation fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/remix-ide/1951/2977).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/remix-ide/1951/2977).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $98,191.43 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [iamonuwa](https://gitcoin.co/profile/iamonuwa) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ethereum/remix-ide/1951/2977?mutate_worker_action=approve&worker=iamonuwa) | [reject worker](https://gitcoin.co/issue/ethereum/remix-ide/1951/2977?mutate_worker_action=reject&worker=iamonuwa))_.**

Am ready to pull this off in 2 days. I understand the requirements and having gone through the integration documentation of MythX, I would love to work on this issue.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/remix-ide/1951/2977).

iamonuwa:
@gitcoinbot work in progress

yann300:
@aquiladev already did a lot, could you team up on this @iamonuwa ?

yann300:
@iamonuwa @aquiladev
From what I see, the code for requesting MythX has already been done by @aquiladev .
[...]

PR #1999, opened 2019-05-20
MythX plugin

aquiladev:
Repo: https://github.com/aquiladev/remix-mythx-plugin
Issue: https://github.com/ethereum/remix-ide/issues/1951


src/remixAppManager.js
```diff
@@ -84,9 +84,23 @@ export class RemixAppManager extends AppManagerApi {
...
+      name: 'remythx',
+      displayName: 'MythX',
+      description: 'Security analysis of smart contracts by using MythX',
+      events: [],
+      methods: [],
+      notifications: {
+        'solidity': ['compilationFinished']
+      },
+      url: 'http://remythx.xyz',
+      icon: 'http://remythx.xyz/logo.png',
```
yann300: could we use `https` for both url and icon?

aquiladev: yes

b-mueller:
@aquiladev what are the steps to get run an analysis?

I started your remix branch and compiled a couple of contracts, but the MythX tab keeps showing the message "You need to compile your contract first!". Not sure if I'm doing it wrong.

Screen Shot 2019-05-22 at 21 02 16

aquiladev:
@b-mueller Yeah, there is an issue with throwing events on the local environment. I'm figuring out it.
You can open http://remix-alpha.ethereum.org and `Connect to Local Plugin` in order to test it

image

b-mueller:
@aquiladev works pretty nicely! Some feedback on the reporting:

1. Left-clicking on the SWC link opens the code editor (in Safari at least). 
2. It's not really clear that clicking on the issue title leads to code highlighting. How about putting that link only on the "line:column" part (but with link style)?

E.g. 
```
[1:0] A floating pragma is set [SWC-1xx]
^ Link to code                           ^ Link to SWC
```

3. How about displaying the long description on the report page as well?

Gtonizuka:
looks good to me. @aquiladev are you planning implement more features from mythx?

aquiladev:
@b-mueller I'll take a look. There is an issue with SWC link, coz the plugin is rendered in an iframe I think.
@Gtonizuka what features? I open to discuss

b-mueller:
@aquiladev I think feature-wise this is pretty complete. What is still missing is some tool tips and explanation what the analysis actually does and how to obtain the API credentials. Probably also a hint that the analysis takes around 2 minutes to complete.

Not sure where to best put all of this design-wise. I'll also look over the texts in the [Remix-plugin](https://github.com/aquiladev/remix-mythx-plugin) repo.

b-mueller:
@aquiladev One feature that would be nice to have would be a possibility to view full issue descriptions (which are returned by the API in the 'tail' field), or even view/export a detailed report that contains the full descriptions. 

We are currently also working on returning call traces that show how a particular issue/exception is triggered which will be help users to determine the cause of an issue, so some space in the UI will be needed to display that additional info.

We are planning to offer a reports dashboard on the MythX website itself, but this is still a few months away :(

aquiladev:
> @aquiladev One feature that would be nice to have would be a possibility to view full issue descriptions (which are returned by the API in the 'tail' field), or even view/export a detailed report that contains the full descriptions.

`tail` is displaying on hover of issue. it is not obvious, but when I display it fully it is too messy. I need to think about UX

aquiladev:
> @aquiladev I think feature-wise this is pretty complete. What is still missing is some tool tips and explanation what the analysis actually does and how to obtain the API credentials. Probably also a hint that the analysis takes around 2 minutes to complete.

I've added info tooltips

b-mueller:
I tested the plugin pretty extensively and it's quite awesome. The only real drawback from the user's view is that there's no option to view a full security report that also contains the detailed descriptions and the soon-to-be-added call traces. In some cases this will mean that the results are not particularly useful (for example when we detect a reachable exception but don't output the transaction sequence / inputs that trigger it).

This is also a question for @yann300 - is it possible to to add a pop/overlay that displays a full report on request? Possibly we can set up an additional bounty for this. Also tagging @Gtonizuka @aquiladev

yann300:
>  is it possible to to add a pop/overlay that displays a full report on request?
@b-mueller 
not at the moment, but we are thinking of it. It is also possible to build a `mainPanel` plugin, that will be displayed as a tab. That why the view has more space for displaying more detailed informations.
cc @GrandSchtroumpf

yann300:
btw you can also provide a `documentation` property which should be an URL pointing to the doc if you have one.

aquiladev:
@yann300 if `readme.md` is enough I can add it

yann300:
yes I suppose that should be fine for the moment

PR #2006, opened 2019-05-21
fixes publish Swarm

LianaHus:


src/app/tabs/compile-tab.js
```diff
@@ -220,6 +223,10 @@ class CompileTab extends CompilerApi {
     : yo`<section class="${css.container} clearfix"><article class="${css.compilerArticle}">
       <span class="alert alert-warning" role="alert">No Contract Compiled Yet</span>
     </article></section>`
+
+    if (0 !== contractList.length)
```
yann300: better check with `if (!contractList.length)`

yann300: could we reset `this.selectedContract` if not contract is compiled

PR #2025, opened 2019-05-25
Fix 'ballot' and 'compiling' browser tests under nightwatch_local

scottt:
The `testSignature()` browser test in `generalTests.js` would
occasionally fail due to not obtaining the correct `hash` and
`signature` values from the browser, resulting in:
```
✖ generalTests
   - Simple Contract (40.954s)
   Failed [equal]: ('' == '0: address: 0xCA35b7d915458EF540aDe6068dFe2F44E8fa733c')  - expected "0: address: 0xCA35b7d915458EF540aDe6068dFe2F44E8fa733c" but got: ""
       at Object.<anonymous> (/home/scottt/work/remix-0.8/remix-ide/test-browser/helpers/contracts.js:129:22)
```
failures.

This patch fixes it by waiting for the DOM elements with
`waitForElementPresent` which implicitly retries and obtaining
the correct contract instance address instead of hard coding it.

### DIFF
--- a/src/app/tabs/runTab/settings.js
+++ b/src/app/tabs/runTab/settings.js
@@ -97,7 +97,7 @@ class SettingsUI {
         <div class=${css.account}>
           <select name="txorigin" class="form-control ${css.select}" id="txorigin"></select>
           ${copyToClipboard(() => document.querySelector('#runTabView #txorigin').value)}
-          <i id="remixRunSignMsg" class="fas fa-edit ${css.icon}" aria-hiden="true" onclick=${this.signMessage.bind(this)} title="Sign a message using this account key"></i>
+          <i id="remixRunSignMsg" class="fas fa-edit ${css.icon}" aria-hidden="true" onclick=${this.signMessage.bind(this)} title="Sign a message using this account key"></i>
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


yann300:
Thanks for the PR,
We have fixed the issue and test should pass now, we'd prefer not to modify the browser test
But that said, it makes sense to include your PR because it contains some nice improvement.
Could you remove the calls to  `perform(contractHelper.waitForValue('#txorigin', 10))` and keep the rest, we then merge it.
cc @LianaHus 

will have a look on the `evmVersion` shortly

scottt:
> Could you remove the calls to `perform(contractHelper.waitForValue('#txorigin', 10))` and keep the rest, we then merge it.

Done! :)

> We have fixed the issue and test should pass now, we'd prefer not to modify the browser test

I believe you're referring to:
https://github.com/ethereum/remix-ide/commit/0351eb19b6418d962e320f4a582a0c5e839f16ac
which makes `waitForValue('#txorigin')` no longer necessary by changing the way the account list is filled?

> will have a look on the `evmVersion` shortly

I really appreciate the patch reviews and the work you guys put into this open source project.

Issue #2024, opened 2019-05-24
Viper plugin


LianaHus:
Create file->strange empty toaster is created

PR #2028, opened 2019-05-27
makes tooltip work with DOM element as an inpput

LianaHus:
fixes https://github.com/ethereum/remix-ide/issues/2024
fixes https://github.com/ethereum/remix-ide/issues/2023


LianaHus:
still need to adjust the style

PR #2051, opened 2019-05-29
Fix spelling of GitHub

fulldecent:


fulldecent:
This fix is wrong.

Correct fix is included in #2052

Issue #1992, opened 2019-05-20
Make documentation links absolute


ryestew:
Links from the icon in the title of the swap panel should be absolute.  Update the swap-panel.js and all the profiles that use the swap panel.

PR #2056, opened 2019-05-31
Add Provable plugin to app manager

D-Nice:
Adds latest remix-alpha compatible Provable plugin to the app manager.


yann300:
thanks for this PR ;) 
Do you have any documentation page? I think it would be helpful to put a documentation link cause it is not obvious how to use it at the moment. You can add `documentation` property containing an URL in the profile.
Also did the plugin work for all the env or only for JavaScript VM? I recall that it was only for JavaScript VM but the state does not change when injected is selected... might be an issue with the API please let us know.

D-Nice:
@yann300 I thought we had some documentation on it, but double checking we have the old browser-solidity fork documented instead of the plugin, so we'll look to rectify that and then add a commit with the property.

Yes, plugin is still only for JavaScript VM, and should appropriately display the warning. Unfortunately, I had trouble getting a local build of remix working in general, so I simply tested using the currently up version of remix alpha at the URL, and then importing the plugin via connection option and URL https://remix-plugin.provable.xyz/, it should be equivalent to importing via this app manager I believe. The only recommendation I have is that before when I tested, remix would allow you to choose injected, even though there was no injected environment. Testing it today I no longer see that option, it forced me to have metamask or the like up, so worked in all cases for me: 

2019-06-04-100501_423x442_scrot

D-Nice:
added documentation prop as requested with link containing the pertinent info.

yann300:
thanks for the link, where can we find examples of using the Provable API?

D-Nice:
We have a dedicated pinned repository with examples. DieselPrice or KrakenPriceTicker are some good starting points and what I used to test the plugin: https://github.com/oraclize/ethereum-examples/tree/master/solidity

yann300:
@D-Nice thanks, we'll merge the PR just after https://github.com/ethereum/remix-ide/issues/1992
would it make sense to put https://github.com/oraclize/ethereum-examples/tree/master/solidity in the documentation link put the link in the doc itself. That would give good `how to get started` for first users.

D-Nice:
Yes, it makes sense and thanks for the feedback on that, will do this by end of today

D-Nice:
added link to examples in documentation

PR #2071, opened 2019-06-07
Adding the remixd command to the remixd modal

ryestew:


yann300:
tests are failing... perhaps change `<br/>` to `<br>`

yann300:
just tried, the command is not really visible like that, that should perhaps be wrapped to look as a command line

ryestew:
I'll wrap it differently - but I don't think that is the cause of the failure because all my tests are failing at this point: ERROR: Unable to locate element: "#icon-panel div[plugin="solidity"]" using: css selector 
( in other PR's)

PR #2076, opened 2019-06-11
Link2docs

ryestew:
don't merge this yet


ryestew:
Part of the next release:
Make documentation links absolute #1992

ryestew:
the absolute links should be pointing to the docs folder in remix-ide - which isn't there yet but will be when this PR is merged #2077

PR #2087, opened 2019-06-17
Select Yul compiler

yann300:
using https://gist.githubusercontent.com/chriseth/4402022ba55f18bde768fc1f8bcaed29/raw/c5736f40a12051901e81581c1f49bcd514659b53/ERC20.yul

### DIFF
--- a/src/app/tabs/compileTab/compileTab.js
+++ b/src/app/tabs/compileTab/compileTab.js
@@ -46,14 +46,16 @@ class CompileTab {
     this.compiler.setEvmVersion(this.evmVersion)
   }
 
+  setLanguage (lang) {
+    this.compiler.setLanguage(lang)
+  }
+
   /**
    * Compile a specific file of the file manager
    * @param {string} target the path to the file to compile
    */
   compileFile (target) {
     if (!target) throw new Error('No target provided for compiliation')
-    if (!/\.sol/.exec(target)) throw new Error(`${target} is not a solidity file. It cannot be compiled with solidity compiler`)
-    // only compile *.sol file.
     const provider = this.fileManager.fileProviderOf(target)
     if (!provider) throw new Error(`cannot compile ${target}. Does not belong to any explorer`)
     return new Promise((resolve, reject) => {
--- a/src/app/tabs/compileTab/compileTab.js
+++ b/src/app/tabs/compileTab/compileTab.js
@@ -45,7 +45,11 @@ class CompileTab {
     this.queryParams.update({ evmVersion: this.evmVersion })
     this.compiler.setEvmVersion(this.evmVersion)
   }
-
+  
+  /**
+   * Set the compiler to using Solidity or Yul (default to Solidity)
+   * @params lang {'Solidity' | 'Yul'} ...
+   */
   setLanguage (lang) {
     this.compiler.setLanguage(lang)
   }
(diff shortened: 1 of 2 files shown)


src/app/tabs/compileTab/compileTab.js
```diff
@@ -46,14 +46,16 @@ class CompileTab {
...
+  }
+
   /**
    * Compile a specific file of the file manager
    * @param {string} target the path to the file to compile
    */
   compileFile (target) {
     if (!target) throw new Error('No target provided for compiliation')
-    if (!/\.sol/.exec(target)) throw new Error(`${target} is not a solidity file. It cannot be compiled with solidity compiler`)
-    // only compile *.sol file.
```
LianaHus: should we check for .sol or .yul?

yann300: There is a case (when we load a swarm hash) where the extension is not specified, but the content is still solidity. So it kinda make sense to try compiling it.
That basically why I removed that one.

src/app/tabs/compileTab/compileTab.js
```diff
@@ -46,14 +46,16 @@ class CompileTab {
     this.compiler.setEvmVersion(this.evmVersion)
   }
 
+  setLanguage (lang) {
```
GrandSchtroumpf: What litteral can be `lang` ? "Solidity" or "solidity" or "sol" ? You may want to have a comment here : 
```
/**
 * ...
 * @params lang {'solidity' | 'yul' | 'vyper'} ...
 */
setLanguage (lang) 
```

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -227,6 +238,11 @@ class CompilerContainer {
     this.compileTabLogic.runCompiler()
   }
 
+  onchangeLanguage (event) {
```
GrandSchtroumpf: Can it be "selectLangage" or "changeLangage". This is the action that is triggered on change from select.

src/app/tabs/compileTab/compileTab.js
yann300: that can be `Solidity` or `Yul`.
@chriseth is there other valid input for the `language` property?

src/app/tabs/compileTab/compilerContainer.js
yann300: i'd keep it like that for the moment as the format `onSomething` is already used in this file

src/app/tabs/compileTab/compileTab.js
chriseth: Not for now.

LianaHus:
@yann300 when I compile with Yul I have "Only "Solidity" is supported as a language." error. Is this expected behavior?

yann300:
Yes I see, let's do that in a subsequent PR, The error message is i guess clear enough to let this PR in,
and at load, the default compiler is still "Solidity" (is it?)
beside that instead of forcing Solidity, we should just remove Yul from the list.
Please create an issue for that.
@chriseth since which version Yul is accepted as a keyword for `language`?

LianaHus:
ok as you want. but I would not release without this fix. I'll create an Issue for that

chriseth:
@yann300 since version 0.5.7

LianaHus:
https://github.com/ethereum/remix-ide/issues/2100

Issue #2100, opened 2019-06-20
Compiler tab: language selector


LianaHus:
The language selector should be disabled for compiler versions which don't support Yul

PR #2108, opened 2019-06-24
compiler tab style fixes. language selector ui behaviour 

LianaHus:
fixes https://github.com/ethereum/remix-ide/issues/2100

### DIFF
--- a/src/app/tabs/compileTab/compilerContainer.js
+++ b/src/app/tabs/compileTab/compilerContainer.js
@@ -110,7 +110,7 @@ class CompilerContainer {
     return el
   }
 
-  _retriveVersion () {
+  _retrieveVersion () {
     let version = this._view.versionSelector.value
     return version.substring(9, 14)
   }
@@ -306,7 +306,7 @@ class CompilerContainer {
   }
 
   _updateLanguageSelector () {
-    if (this._retriveVersion() < '0.5.7') {
+    if (this._retrieveVersion() < '0.5.7') {
       this._view.languageSelector.setAttribute('disabled', '')
       this._view.languageSelector.value = 'Solidity'
       this.compileTabLogic.setLanguage('Solidity')
--- a/src/app/tabs/compileTab/compilerContainer.js
+++ b/src/app/tabs/compileTab/compilerContainer.js
@@ -166,10 +166,10 @@ class CompilerContainer {
       <section>
         <!-- Select Compiler Version -->
         <article>
-          <header class="navbar navbar-light bg-light">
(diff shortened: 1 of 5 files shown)


LianaHus:
https://github.com/ethereum/remix-ide/pull/2087

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -294,6 +305,16 @@ class CompilerContainer {
     }
   }
 
+  _updateLanguageSelector () {
+    if (this._retriveVersion() < '0.5.7') {
```
yann300: how this can work? you should probably use semver here.

yann300:
some infos about semver https://www.npmjs.com/package/semver

src/app/tabs/compileTab/compilerContainer.js
LianaHus: :) it works. ok I'll look at it

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -240,7 +248,9 @@ class CompilerContainer {
 
   onchangeLanguage (event) {
     this.compileTabLogic.setLanguage(event.target.value)
-    this.compile()
```
yann300: all the other `on*` method, just trigger a new compilation. Might be more consistent to to keep the same behavior here

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -110,6 +110,11 @@ class CompilerContainer {
     return el
   }
 
+  _retriveVersion () {
```
yann300: typo

src/app/tabs/compileTab/compilerContainer.js
LianaHus: but there was a issue here. build shouldn't be triggered if there is no autocompile an selected file

LianaHus: or I can put the check in the compile() function

yann300: > but there was a issue here. build shouldn't be triggered if there is no autocompile an selected file

This should done in another PR and all the `on*` function should be updated.

LianaHus: ok will create a new PR

yann300:
Click on `Solidity` env (from the front page), the selected compiler is `0.5.1`, the dropdown list is still enabled. I guess it should not be

Issue #2113, opened 2019-06-25
Cannot write in the Terminal


yann300:
The html input is broken
This is not in remix.ethereum.org so it has been introduced very recently.

PR #2125, opened 2019-06-27
Fix setting input property

yann300:
That is probably due to the use of babel 7...
This is ok like this but we have to find a way to make this work without using `setAttribute`
fix https://github.com/ethereum/remix-ide/issues/2113

### DIFF
--- a/src/app/panels/terminal.js
+++ b/src/app/panels/terminal.js
@@ -111,8 +111,11 @@ class Terminal extends BaseApi {
     if (self._view.el) return self._view.el
     self._view.journal = yo`<div class=${css.journal}></div>`
     self._view.input = yo`
-      <span class=${css.input} spellcheck="false" onload=${() => { this.focus() }} contenteditable="true" onpaste=${paste} onkeydown=${change}></span>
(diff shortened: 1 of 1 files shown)


src/app/panels/terminal.js
```diff
@@ -405,24 +408,22 @@ class Terminal extends BaseApi {
...
-            <li><a target="_blank" href="https://www.npmjs.com/package/swarmgw">swarmgw</a> </li>
-            <li>remix (run remix.help() for more info)</li>
-          </ul>
-        </li>
-        <li>Executing common command to interact with the Remix interface (see list of commands above). Note that these commands can also be included and run from a JavaScript script.</li>
-        <li>Use exports/.register(key, obj)/.remove(key)/.clear() to register and reuse object across script executions.</li>
-      </ul>
-      </div>
-    `
+    var intro = yo`<div><div> - Welcome to Remix ${packageV.version} - </div><br>
```
LianaHus: why?

yann300: that's reformatting issue, will rollback that

PR #2137, opened 2019-07-01
Update tests to include 1 for checking if the console works

ryestew:


ryestew:
This PR has 3 returns in it - it really needed 2 to work.  But as soon as the terminal is fixed it will only need 1 return.

test-browser/tests/console.js
```diff
@@ -0,0 +1,18 @@
...
+
+module.exports = {
+  before: function (browser, done) {
+    init(browser, done)
+  },
+  'SimpleExecutionConsole': function (browser) {
+    browser
+    .waitForElementVisible('#terminalCli', 10000)
+    .executeScript(['\uE01B','\uE025','\uE01B','\uE006','\uE006','\uE006'])
+    .waitForElementVisible('#journal div:last-child span.text-info', 10000)
```
yann300: `waitForElementVisible` should perhaps be in `journalLastChild` (this is the thing you are waiting to appear)

test-browser/tests/console.js
```diff
@@ -0,0 +1,18 @@
...
+var sauce = require('./sauce')
+
+module.exports = {
+  before: function (browser, done) {
+    init(browser, done)
+  },
+  'SimpleExecutionConsole': function (browser) {
+    browser
+    .waitForElementVisible('#terminalCli', 10000)
+    .executeScript(['\uE01B','\uE025','\uE01B','\uE006','\uE006','\uE006'])
```
yann300: does that include hitting `return`?

test-browser/commands/journalLastChild.js
```diff
@@ -0,0 +1,13 @@
+const EventEmitter = require('events')
+
+class JournalLastChild extends EventEmitter {
+  command (val) {
+    this.api.waitForElementVisible('#journal div:last-child span.text-info', 10000)
```
yann300: I think you have to queue the calls here

yann300: I think you have to queue the call here

test-browser/tests/console.js
```diff
@@ -0,0 +1,17 @@
...
+var sauce = require('./sauce')
+
+module.exports = {
+  before: function (browser, done) {
+    init(browser, done)
+  },
+  'SimpleExecutionConsole': function (browser) {
+    browser
+    .waitForElementVisible('#terminalCli', 10000)
+    .executeScript(['1+1'])
```
yann300: this should work without the input being an array

test-browser/commands/journalLastChild.js
GrandSchtroumpf: Maybe use only ids

PR #2140, opened 2019-07-01
Adding package-lock.json to repository

edisinovcic:
Using package-lock.json so dependency installation is usually faster (npm ci https://docs.npmjs.com/cli/ci.html) and dependencies are fixed so adding it should be beneficial.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@ reports
 soljson.js
 soljson.js.*
 npm-debug.log*
-package-lock.json
 remix
 .DS_Store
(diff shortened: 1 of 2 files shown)


yann300:
thanks for this (actually we were discussion that issue this morning)
we should also remove it from https://github.com/ethereum/remix-ide/blob/master/.gitignore#L10

edisinovcic:
Definitelly will update .gitignore too.

edisinovcic:
Updated.

PR #2145, opened 2019-07-02
added closure functions for onclick events

LianaHus:


src/app/files/fileManager.js
```diff
@@ -206,6 +201,20 @@ class FileManager extends FileSystemApi {
...
   }
 
+  _saveAsCopy (path, content) {
+    const fileProvider = this.fileProviderOf(path)
+    if (fileProvider) {
+      helper.createNonClashingNameWithPrefix(path, fileProvider, '', (error, copyName) => {
+        if (error) {
+          console.log('createNonClashingNameWithPrefix', error)
+          copyName = path + '.' + this.currentRequest.from
+        }
```
GrandSchtroumpf: Use `else` here, or if there is an error, it'll still go through that part.

LianaHus: right! thanks

PR #2147, opened 2019-07-02
remove cleanup of plugins before new workspace load

LianaHus:


yann300:
the commit is missing something I think

LianaHus:
> the commit is missing something I think

haven't save :) will push now

src/app/ui/landing-page/landing-page.js
```diff
@@ -22,13 +22,9 @@ let css = csjs`
     user-select:none;
   }
   .jumbotronContainer {
```
GrandSchtroumpf: Could you remove the class if it's empty ?

src/app/ui/landing-page/landing-page.js
```diff
@@ -127,21 +123,14 @@ export class LandingPage extends BaseApi {
 
     let learnMore = () => { window.open('https://remix.readthedocs.io/en/latest/layout.html', '_blank') }
 
-    let closeAll = () => {
-      this.appStore.getActives()
-      .filter(({profile}) => !profile.required)
-      .forEach((profile) => { this.appManager.deactivateOne(profile.name) })
-    }
     let startSolidity = () => {
```
GrandSchtroumpf: I think it would be good practice to change all `var` and `let` into `const` if applicable.

src/app/ui/landing-page/landing-page.js
LianaHus: I don't know why are those here. @ryestew can we remove those?

LianaHus: I don't know why are those here. @ryestew can we remove those?

LianaHus:
Sure. But don't think it's relevant to this PR. We can do that as a separate PR

src/app/ui/landing-page/landing-page.js
ryestew: Yes we can remove that class definition.

PR #2180, opened 2019-07-10
fixed swapPanel height

LianaHus:


src/app/components/side-panel.js
```diff
@@ -29,8 +31,9 @@ const css = csjs`
     cursor: pointer;
   }
   .pluginsContainer {
-    height: calc(100% - 35px);
-    overflow: auto;
+    height: 100%;
```
GrandSchtroumpf: remove height here. flex: 1 should be enough.

src/app/components/side-panel.js
```diff
@@ -29,8 +31,9 @@ const css = csjs`
...
+    flex: 1;
+    padding-bottom: 40px;
```
GrandSchtroumpf: Why ?

src/app/components/side-panel.js
```diff
@@ -5,7 +5,9 @@ const yo = require('yo-yo')
 const css = csjs`
   .panel {
     height: 100%;
-    overflow-y: hidden; 
+    overflow-y: hidden;
```
GrandSchtroumpf: remove this. There is no reason to hide it now.

src/app/components/side-panel.js
LianaHus: no it's not

src/app/components/side-panel.js
LianaHus: because otherwise the top is shifted when it's not an iframe

PR #2182, opened 2019-07-11
Fix modal dialog callback being called multiple times

yann300:
When the modal closes, we need to remove the DOM element. 
If not, we create a new EventListener associated with the modal each time the modal shows up.
And basically when the user `Cancel` the modal, all the previous callback where the modal was used get called.
That's pretty critical


yann300:
for reproducing:
 - run remixd, accept the connection.
 - right click on any file (from the browser explorer for instance) for deletion
 - when the modal shows up, click outside
 - result: the remixd connection get closed

PR #2190, opened 2019-07-16
Remove verbose logs

yann300:


src/app/panels/terminal.js
```diff
@@ -201,9 +201,7 @@ class Terminal extends Plugin {
     setInterval(async () => {
       try {
         self._view.pendingTxCount.innerHTML = await self.call('udapp', 'pendingTransactionsCount')
-      } catch (err) {
-        console.log('Terminal try to call "udapp".')
-      }
+      } catch (err) {}
```
GrandSchtroumpf: Maybe we should add a TODO + issue to remove this try and create an event instead

yann300: issue https://github.com/ethereum/remix-ide/issues/2191

Issue #2014, opened 2019-05-22
close tab(plugin)


LianaHus:
Steps to reproduce:
open remix-ide
home->pipeline
close pipeline tab
expected: home page content
actual: empty editor

PR #2195, opened 2019-07-16
Close tab issue

LianaHus:
fixes https://github.com/ethereum/remix-ide/issues/2014

### DIFF
--- a/src/app/panels/tab-proxy.js
+++ b/src/app/panels/tab-proxy.js
@@ -20,6 +20,7 @@ export class TabProxy {
 
     fileManager.events.on('fileClosed', (name) => {
       this.removeTab(name)
+      this.switchPreviousTab()
     })
 
     fileManager.events.on('currentFileChanged', (file) => {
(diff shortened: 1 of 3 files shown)


src/app/panels/tab-proxy.js
```diff
@@ -96,7 +97,6 @@ export class TabProxy {
       const handlers = Object.keys(this._handlers)
       let i = handlers.indexOf(active)
       if (i >= 0) {
-        i = handlers[i - 1] ? i - 1 : handlers.length - 1
```
yann300: this line should not be removed cause it completely changes the behavior of this function. then it is no longer `switching to Previous Tab`

src/app/panels/tab-proxy.js
```diff
@@ -20,6 +20,7 @@ export class TabProxy {
 
     fileManager.events.on('fileClosed', (name) => {
       this.removeTab(name)
+      this.switchPreviousTab()
```
yann300: if you want to switch to home when no tab are displayed, that's not the right solution.

PR #2212, opened 2019-07-22
Improve Plugin permissions + Use emit function from plugin

GrandSchtroumpf:
Expose all the events to external plugins.

### DIFF
--- a/src/app/ui/persmission-handler.js
+++ b/src/app/ui/persmission-handler.js
@@ -78,7 +78,7 @@ export class PermissionHandler {
               }
               this.persistPermissions()
             }
-            resolve()
+            resolve(true)
           }
         },
         {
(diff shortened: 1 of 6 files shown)


src/framingService.js
```diff
@@ -10,6 +10,7 @@ export class FramingService {
 
   start () {
     this.sidePanel.events.on('toggle', () => {
+      console.log('Toggled')
```
yann300: should be remove I guess

src/app/ui/persmission-handler.js
```diff
@@ -105,19 +105,23 @@ export class PermissionHandler {
...
+      if (!allow) {
+        const warning = notAllowWarning(from, to)
+        addTooltip(warning)
+        throw new Error(warning)
+      }
+      return hash === from.hash
+        ? true  // Allow
+        : this.openPermission(from, to)  // New version of a plugin
+    } catch (err) {
+      return false
```
yann300: shall we ` throw new Error(err)` here, as for the line 117

PR #2219, opened 2019-07-24
adding `retrieveTrace` to the remix-plugin API

yann300:
This add a new function to the debugger API available from the plugin infra : "retrieveTrace" .

### DIFF
--- a/src/app/tabs/debugger-tab.js
+++ b/src/app/tabs/debugger-tab.js
@@ -9,7 +9,7 @@ import * as packageJson from '../../../package.json'
 const profile = {
   name: 'debugger',
   displayName: 'Debugger',
-  methods: ['debug'],
+  methods: ['debug', 'retrieveTrace'],
   events: [],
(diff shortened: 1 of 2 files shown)


src/app/tabs/debugger-tab.js
```diff
@@ -38,7 +38,11 @@ class DebuggerTab extends ViewPlugin {
   }
 
   debug (hash) {
-    if (this.debugger) this.debuggerUI.debug(hash)
+    if (this.debuggerUI) this.debuggerUI.debug(hash)
+  }
+
+  async retrieveTrace (hash) {
```
GrandSchtroumpf: Maybe "getTrace" would be a better name for an external method ?

src/app/tabs/debugger/debuggerUI.js
```diff
@@ -103,38 +103,65 @@ class DebuggerUI {
     return this.isActive
   }
 
-  startDebugging (blockNumber, txNumber, tx) {
+  async getDebugWeb3 () {
```
GrandSchtroumpf: You don't need async as you're already returning a promise.

src/app/tabs/debugger/debuggerUI.js
```diff
@@ -103,38 +103,65 @@ class DebuggerUI {
...
+    return new Promise((resolve, reject) => {
+      executionContext.detectNetwork((error, network) => {
+        let web3
+        if (error || !network) {
+          web3 = init.web3DebugNode(executionContext.web3())
+        } else {
+          var webDebugNode = init.web3DebugNode(network.name)
```
GrandSchtroumpf: You can use `const` here

src/app/tabs/debugger/debuggerUI.js
```diff
@@ -103,38 +103,65 @@ class DebuggerUI {
...
+          web3 = (!webDebugNode ? executionContext.web3() : webDebugNode)
```
GrandSchtroumpf: Would this work ? `web3 = webDebugNode || executionContext.web3()`

src/app/tabs/debugger/debuggerUI.js
```diff
@@ -103,38 +103,65 @@ class DebuggerUI {
...
+    let web3 = await this.getDebugWeb3()
+    this.debugger = new Debugger({
+      web3,
+      offsetToLineColumnConverter: this.registry.get('offsettolinecolumnconverter').api,
+      compiler: { lastCompilationResult }
+    })
+
+    this.listenToEvents()
+    this.debugger.debug(blockNumber, txNumber, tx, () => {
+      self.stepManager = new StepManagerUI(this.debugger.step_manager)
```
GrandSchtroumpf: You don't need `self` here, because you're using an array function.

src/app/tabs/debugger/debuggerUI.js
```diff
@@ -103,38 +103,65 @@ class DebuggerUI {
...
+      self.vmDebugger = new VmDebugger(this.debugger.vmDebuggerLogic)
+      self.txBrowser.setState({ blockNumber, txNumber, debugging: true })
+      self.renderDebugger()
+    }).catch((error) => {
+      toaster(error)
+      this.unLoad()
+    })
+  }
+
+  async retrieveTrace (hash) {
```
GrandSchtroumpf: You don't need async here

src/app/tabs/debugger/debuggerUI.js
```diff
@@ -103,38 +103,65 @@ class DebuggerUI {
...
+    return new Promise(async (resolve, reject) => {
+      let compilers = this.registry.get('compilersartefacts').api
```
GrandSchtroumpf: You can use `const` here as you're not reasigning it later (just mutate it)
Same for web3 and debug

src/app/tabs/debugger/debuggerUI.js
yann300: yeah.. I have a standard error if i do the other way around

src/app/tabs/debugger-tab.js
```diff
@@ -38,7 +38,11 @@ class DebuggerTab extends ViewPlugin {
   }
 
   debug (hash) {
-    if (this.debugger) this.debuggerUI.debug(hash)
+    if (this.debuggerUI) this.debuggerUI.debug(hash)
+  }
+
+  async getTrace (hash) {
```
GrandSchtroumpf: You don't need async and await here. Sorry I didn't see this one.

PR #2232, opened 2019-07-27
add the debug plugin

pldespaigne:
Hello ! I made this PR to add my debug plugin, it will help plugin devs debugging their plugins, and remix devs testing the plugin-api !
Also sources are available [here](https://github.com/pldespaigne/remix-debug-plugin) if you want to have a look.


src/remixAppManager.js
```diff
@@ -179,6 +179,17 @@ export class RemixAppManager extends PluginEngine {
...
+    const debugPlugin = {
+      name: 'debugPlugin',
+      displayName: 'Debug Tools for Remix plugins',
+      description: 'Easily test and debug your plugins !',
+      methods: ['sayHello', 'sayMyName', 'sayOurNames'], // test calls with 0, 1, and 2 args
+      events: [],
+      version: '0.1.0-alpha',
+      url: 'https://remix-debug-a.surge.sh',
+      icon: 'https://remix-debug-a.surge.sh/icon.png',
+      location: 'sidePanel'
```
GrandSchtroumpf: Wouldn't it be better in the mainpanel ? As most of the plugins are inside the sidepanel, you can have both of them on the same page.

pldespaigne: That make sense ! I will test the ui in the main panel then

src/remixAppManager.js
```diff
@@ -179,6 +179,17 @@ export class RemixAppManager extends PluginEngine {
...
     }
+    const debugPlugin = {
+      name: 'debugPlugin',
+      displayName: 'Debug Tools for Remix plugins',
+      description: 'Easily test and debug your plugins !',
+      methods: ['sayHello', 'sayMyName', 'sayOurNames'], // test calls with 0, 1, and 2 args
+      events: [],
+      version: '0.1.0-alpha',
+      url: 'https://remix-debug-a.surge.sh',
+      icon: 'https://remix-debug-a.surge.sh/icon.png',
```
LianaHus: we use a very similar icon for our native Debug plugin. both will be shown on Vertical panel when activated.

pldespaigne: Yeah I wasn't sure about my icon, I try to mix the debug icon with the plugin manager icon, I can change that without changing the icon's url. Also I'm open to any icon suggestion

LianaHus: https://fontawesome.com/icons/spider?style=solid ? :) I'm not sure either, but it should be different

pldespaigne: I'll do it this evening

pldespaigne: Hi ! I have changed the icon, what do you think now ?
debug plugin icon

LianaHus: I like it :))

LianaHus:
@pldespaigne could you please resolve the conflict so we can merge?

Issue #2250, opened 2019-08-09
Run tab: danger color for payable functions


LianaHus:
currently it is warning

PR #2251, opened 2019-08-09
danger color for payable functions

LianaHus:
fixes https://github.com/ethereum/remix-ide/issues/2250

### DIFF
--- a/src/app/ui/multiParamManager.js
+++ b/src/app/ui/multiParamManager.js
@@ -210,6 +210,7 @@ class MultiParamManager {
       let funButton = this.contractActionsContainerSingle.querySelector('button')
       funButton.setAttribute('title', (title + ' - transact (payable)'))
       funButton.classList.add('btn-danger')
+      button.classList.add('btn-danger')
     }
 
     if (!this.lookupOnly && this.funABI.payable === false) {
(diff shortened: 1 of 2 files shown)


src/app/ui/multiParamManager.js
```diff
@@ -207,7 +207,10 @@ class MultiParamManager {
     if (this.funABI.payable === true) {
```
ryestew: Does this line need to check if mutable function returns payable and not check this.funABI.payable?

src/app/ui/multiParamManager.js
```diff
@@ -207,7 +207,10 @@ class MultiParamManager {
...
       contractProperty.classList.add(css.payable)
       button.setAttribute('title', (title + ' - transact (payable)'))
-      this.contractActionsContainerSingle.querySelector('button').setAttribute('title', (title + ' - transact (payable)'))
+      let funButton = this.contractActionsContainerSingle.querySelector('button')
+      funButton.setAttribute('title', (title + ' - transact (payable)'))
+      funButton.classList.add('btn-danger')
```
ryestew: With this code the btn-danger is getting added to the css - but the btn-warning is still there - it should get removed or really it should get put on there in the 1st place.  The code works as is - but its a bit messy because both classes are there - and the last class which was added - btn-danger wins - because it was the last added.

src/app/ui/multiParamManager.js
LianaHus: sure, 10x

src/app/ui/multiParamManager.js
yann300: @LianaHus @ryestew it looks like that the comment was addressed

Issue #2271, opened 2019-08-19
Expose getToken(apiName) from settings


GrandSchtroumpf:
- [x] Create the method `getToken(apiName)` in settings
- [x] Expose the method in the profile
- [x] Set "true" on profile permission

PR #2273, opened 2019-08-19
getGithubAccessToken API

yann300:
fix https://github.com/ethereum/remix-ide/issues/2271

### DIFF
--- a/src/app/tabs/settings-tab.js
+++ b/src/app/tabs/settings-tab.js
@@ -10,14 +10,15 @@ import * as packageJson from '../../../package.json'
 const profile = {
   name: 'settings',
   displayName: 'Settings',
-  methods: ['getConfigByKey'],
+  methods: ['getGithubAccessToken'],
   events: [],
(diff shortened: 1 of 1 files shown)


src/app/tabs/settings-tab.js
```diff
@@ -10,14 +10,15 @@ import * as packageJson from '../../../package.json'
 const profile = {
   name: 'settings',
   displayName: 'Settings',
-  methods: [],
+  methods: ['getGithubAccessToken'],
```
GrandSchtroumpf: Why "Access" ?

src/app/tabs/settings-tab.js
```diff
@@ -162,4 +163,8 @@ module.exports = class SettingsTab extends ViewPlugin {
     this._deps.themeModule.switchTheme()
     return this._view.el
   }
+
+  getGithubAccessToken () {
+    return new Promise((resolve, reject) => resolve(this.config.get('settings/gist-access-token')))
```
GrandSchtroumpf: You don't need to return a promise : 
```javascript
return this.config.get('settings/gist-access-token')
```

src/app/tabs/settings-tab.js
yann300: that's `Personal access tokens` in https://github.com/settings/tokens

PR #2287, opened 2019-08-22
Use registry for getting the access token

yann300:


src/app/compiler/compiler-imports.js
```diff
@@ -1,4 +1,5 @@
 'use strict'
+let globalRegistry = require('../../global/registry')
```
GrandSchtroumpf: You can use const here

src/app/compiler/compiler-imports.js
```diff
@@ -14,16 +15,16 @@ const profile = {
...
-    this.githubAccessToken = githubAccessToken || (() => {})
     this.previouslyHandled = {} // cache import so we don't make the request at each compilation.
   }
 
   handleGithubCall (root, path, cb) {
     let param = '?'
-
-    param += this.githubAccessToken() ? 'access_token=' + this.githubAccessToken() : ''
+    // const token = await this.call('settings', 'getGithubAccessToken')
+    const token = globalRegistry.get('config').api.get('settings/gist-access-token') // TODO replace with the plugin call above
```
GrandSchtroumpf: Can you create an issue for that and reference it in the TODO statement ?

src/app/compiler/compiler-imports.js
```diff
@@ -14,16 +15,16 @@ const profile = {
...
+    param += token ? 'access_token=' + token : ''
```
GrandSchtroumpf: Maybe more readable to do : 
```typescript
const param = token ? `?access_token=${token}` : '?' 
```
(or "let param = ..." if you're mutating the param later.

PR #2307, opened 2019-09-04
docs: fix url for a workshop from docs & gramatical fix

ryestew:


docs/contract_metadata.md
```diff
@@ -1,8 +1,8 @@
 Build Artifact
 ==============
 
-As compilation succeeds Remix creates a JSON file for each compiled contract.
-These JSON files contains several metadata
+When a compilation succeeds, Remix creates a JSON file for each compiled contract.
+The JSON file contains the compilation's metadata.
```
LianaHus: Probably would be better to have "artifacts" instead of "metadata"  here

PR #2318, opened 2019-09-07
`_deps` is not used anymore in udapp

yann300:
This bug has been introduced after we moved udapp to remix repository


yann300:
we should address https://github.com/ethereum/remix-ide/issues/2319 at some point @LianaHus

PR #2342, opened 2019-09-18
Add plugins from the plugin directory

yann300:
`this.pluginsDirectory` should be replaced by the url to the built list of plugins

### DIFF
--- a/src/remixAppManager.js
+++ b/src/remixAppManager.js
@@ -1,4 +1,4 @@
-/* global localStorage */
+/* global localStorage, fetch */
 import { PluginEngine, IframePlugin } from '@remixproject/engine'
 import { EventEmitter } from 'events'
 import { PermissionHandler } from './app/ui/persmission-handler'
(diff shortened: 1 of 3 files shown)


src/remixAppManager.js
```diff
@@ -262,6 +263,7 @@ export class RemixAppManager extends PluginEngine {
       icon: 'https://zokrates.blockchain-it.hr/zokrates.svg',
       location: 'sidePanel'
     }
+    const plugins = JSON.parse(await fetch(this.pluginsDirectory))
```
GrandSchtroumpf: `await fetch(path)` return a response object. You need to tranform it into a JSON object : 

```javascript
const res = await fetch(path);
const profiles = await res.json();
```

src/remixAppManager.js
```diff
@@ -277,7 +279,8 @@ export class RemixAppManager extends PluginEngine {
       new IframePlugin(gasProfiler),
       new IframePlugin(flattener),
       new IframePlugin(ethpm),
-      new IframePlugin(zokrates)
+      new IframePlugin(zokrates),
+      ...plugins
```
GrandSchtroumpf: I think you need to do: `plugins.map(plugin => new IframePlugin(plugin))`

PR #2352, opened 2019-09-19
firefox restriction removed

Aniket-Engg:
Now asynchronous loading of solidity compiler using web worker works for each browser which is compatible with web worker.

Related to https://github.com/ethereum/remix/issues/1239

### DIFF
--- a/src/app/tabs/compileTab/compilerContainer.js
+++ b/src/app/tabs/compileTab/compilerContainer.js
@@ -319,8 +319,8 @@ class CompilerContainer {
       }
       url = `${this.data.baseurl}/${this.data.selectedVersion}`
     }
-    const isFirefox = typeof InstallTrigger !== 'undefined'
-    if (document.location.protocol !== 'file:' && Worker !== undefined && isFirefox) {
+    // Check if browser is compatible with web worker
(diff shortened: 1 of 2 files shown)


src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -319,8 +319,8 @@ class CompilerContainer {
       }
       url = `${this.data.baseurl}/${this.data.selectedVersion}`
     }
-    const isFirefox = typeof InstallTrigger !== 'undefined'
-    if (document.location.protocol !== 'file:' && Worker !== undefined && isFirefox) {
+    // Check if browser is compatible with web worker
+    if (document.location.protocol !== 'file:' && Worker !== undefined) {
```
GrandSchtroumpf: Maybe you can create a function for that. Like that code is self-explained, and maybe we'll need it somewhere else.
```javascript
function browserSupportWorker() {
  return document.location.protocol !== 'file:' && Worker !== undefined;
}
```

Aniket-Engg: okay sure. will add.

PR #2353, opened 2019-09-19
listen on rootFolderChanged

yann300:
That happen when remixd is started and change its target folder: the explorer need to be warned and update accordingly: reload the root folder.
EDIT: this is used when we run remix IDE with electron

### DIFF
--- a/src/app/files/remixDProvider.js
+++ b/src/app/files/remixDProvider.js
@@ -42,6 +42,9 @@ module.exports = class RemixDProvider {
               this.event.trigger('fileExternallyChanged', [path, content])
             }
           })
+        } else if (data.name === 'rootFolderChanged') {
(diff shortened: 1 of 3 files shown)


src/app/files/remixDProvider.js
```diff
@@ -42,6 +42,9 @@ module.exports = class RemixDProvider {
               this.event.trigger('fileExternallyChanged', [path, content])
             }
           })
+        } else if (data.name === 'rootFolderChanged') {
+          // new path has been set, we should reset
+          this.event.trigger('folderAdded', [this.type + '/'])
```
LianaHus: should we remove the old one?

yann300: i think that 's ok

PR #2359, opened 2019-09-23
Push on several ipfs nodes

yann300:
This allow to push on several ipfs gateway

### DIFF
--- a/src/lib/publishOnIpfs.js
+++ b/src/lib/publishOnIpfs.js
@@ -2,8 +2,11 @@
 
 const async = require('async')
 const IpfsClient = require('ipfs-mini')
-const host = 'ipfs.komputing.org'
-const ipfs = new IpfsClient({ host, port: 443, protocol: 'https' })
+
+const ipfsNodes = [
+  new IpfsClient({ host: 'ipfs.komputing.org', port: 443, protocol: 'https' }),
+  new IpfsClient({ host: 'ipfs.infura.io', port: 5001, protocol: 'https' })
+]
 
 module.exports = (contract, fileManager, cb, ipfsVerifiedPublishCallBack) => {
   // gather list of files to publish
@@ -89,7 +92,7 @@ module.exports = (contract, fileManager, cb, ipfsVerifiedPublishCallBack) => {
 
 async function ipfsVerifiedPublish (content, expectedHash, cb) {
   try {
-    const results = await ipfs.add(content)
+    const results = await severalGatewaysPush(content)
     if (results !== expectedHash) {
       cb(null, { message: 'hash mismatch between solidity bytecode and uploaded content.', url: 'dweb:/ipfs/' + results, hash: results })
     } else {
@@ -98,4 +101,10 @@ async function ipfsVerifiedPublish (content, expectedHash, cb) {
   } catch (error) {
     cb(error)
   }
+}
+
+function severalGatewaysPush (content) {
+  const invert = p => new Promise((resolve, reject) => p.then(reject).catch(resolve)) // Invert res and rej
+  const promises = ipfsNodes.map((node) => invert(node.add(content)))
+  return invert(Promise.all(promises))
 }


src/lib/publishOnIpfs.js
```diff
@@ -99,3 +102,28 @@ async function ipfsVerifiedPublish (content, expectedHash, cb) {
     cb(error)
   }
 }
+
+async function tryPushes (content) {
+  return new Promise((resolve, reject) => {
```
GrandSchtroumpf: As you're using `async` you don't need to create a promise. 
```javascript
await function goodFunctionName(content) {
  let error;  // Do that only if you want to throw the last error after every node have failed
  for (const node of ipfsNodes) {
    try {
      const result = await node.add(content)
      return result  // will stop when first works
    } catch (err) {
      error = err
    }
  }
  throw new Error(error);
}
```

src/lib/publishOnIpfs.js
```diff
@@ -99,3 +102,28 @@ async function ipfsVerifiedPublish (content, expectedHash, cb) {
...
+    let failedCount = 0
+    let alreadyReturned = false
+    const failed = (error) => {
+      console.log(error)
+      if (alreadyReturned) return
+      failedCount++
+      if (failedCount >= ipfsNodes.length) {
```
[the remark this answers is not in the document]
yann300: There is a `forEach`, but we don't `await` for the response

src/lib/publishOnIpfs.js
yann300: in the code above, we shouldn't `await`

yann300: removed the `async` keyword, just keeping the `Promise`

GrandSchtroumpf: Why we shouldn't await ?
From what I understand you're trading **readibility** for a small leap of performance (your array is very small).

src/lib/publishOnIpfs.js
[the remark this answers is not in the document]
GrandSchtroumpf: But you trigger the `failed` function maximum one time per item in the array, so there is no way the `failedCount` is above the length.

src/lib/publishOnIpfs.js
yann300: We are sending requests in parallel and waiting for response:
if one response is succeed, we are good. If we got X errors (X being the amount of IPFS nodes), we fail.

GrandSchtroumpf: I understand that, but the current implementation is very difficult to understand.
If you really want to use concurrency, you can do Promise.all : 
```javascript
const requests = ipfsNodes.map(node => node.add(content).catch(e => e));
const results = await Promise.all(requests);  // Wait for all requests to arrive
const success = results.find(result => !(result instanceof Error));  // Find the first working
success
  ? return success
  : throw new Error(results[0]);
```

yann300: That would be sweet to use `Promise.all`, but that will not work by using it the classic way: 

`Promise.all is rejected if any of the elements are rejected. For example, if you pass in four promises that resolve after a timeout and one promise that rejects immediately, then Promise.all will reject immediately.`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#Promise.all_fail-fast_behaviour

checking if we could tweak it

GrandSchtroumpf: That's why I do `.catch(e => e)`. It feel the result with errors

yann300: Ok thanks will have a look shortly. 
Just to be sure: does this handle the fact that even if one request fail, `Promise.all` still wait for the other request to finish before resolving

GrandSchtroumpf: Yes it should. See second answers of this stakeoverflow question: https://stackoverflow.com/questions/30362733/handling-errors-in-promise-all

GrandSchtroumpf: Again, I think you are readibility for a small amount of performance. I had to try to rewrite the code to understand what it was doing. And I think this is bad practice, espacially for a team like remix.

yann300: Well not really, the infura gateway can reply in 10s / 20s sometimes (even more sometimes), whether our node reply in a few ms.
That the function misses some comment is another thing.

PR #2363, opened 2019-09-23
- fixes import with github, http,resolver-engine

LianaHus:
- fixes tooltip for publish to gist
- add readonly mode(is not finished yet)


src/app/files/file-explorer.js
```diff
@@ -350,7 +350,7 @@ fileExplorer.prototype.init = function () {
 fileExplorer.prototype.publishToGist = function () {
   modalDialogCustom.confirm(
     'Create a public gist',
-    'Are you sure you want to publish all your files anonymously as a public gist on github.com?',
+    'Are you sure you want to publish all your files in browser directory anonymously as a public gist on github.com?',
```
yann300: we should say that this will not include folder

src/app/files/fileProvider.js
```diff
@@ -63,15 +66,23 @@ class FileProvider {
     return true
   }
 
-  addReadOnly (path, content) {
+  addReadOnly (path, content, url) {
+    this.readonlyItems.push(path)
+    if (url !== undefined) this.normalizedNames[url] = path
```
yann300: `if (!url)` will work

src/app/files/fileProvider.js
```diff
@@ -63,15 +66,23 @@ class FileProvider {
     return true
   }
 
-  addReadOnly (path, content) {
+  addReadOnly (path, content, url) {
+    this.readonlyItems.push(path)
```
yann300: we should keep this in the local storage. Perhaps by having a `.remix-system` file.

src/app/files/fileProvider.js
```diff
@@ -63,15 +66,23 @@ class FileProvider {
...
+    if (url !== undefined) this.normalizedNames[url] = path
     return this.set(path, content)
   }
 
   isReadOnly (path) {
-    return false
+    return !this.readonlyItems.includes(path)
```
yann300: `return this.readonlyItems.includes(path)` ?

src/app/files/fileProvider.js
yann300: next PR for that?

src/app/files/file-explorer.js
LianaHus: I would say it's fine. we would implement this anyway, oder?
Wouldn't this be too long with folder stuff?

src/app/files/fileProvider.js
LianaHus: yes

src/app/files/file-explorer.js
yann300: well I don't really think we can implement that easily, gists does not support having folder.
Your sentence is ok but it should explicitly state that folders are not uploaded

PR #2366, opened 2019-09-25
add browser tests for testing basic use of solidity tests tab

yann300:


test-browser/tests/solidityUnittests.js
```diff
@@ -0,0 +1,34 @@
...
+  },
+  tearDown: sauce
+}
+
+function runTests (browser) {
+  browser
+    .waitForElementVisible('#icon-panel', 10000)
+    .click('#icon-panel div[plugin="pluginManager"]')
+    .scrollAndClick('#pluginManager article[id="remixPluginManagerListItem_solidityUnitTesting"] button')
+    .clickLaunchIcon('fileExplorers')
```
LianaHus: shouldn't you activate solidity plugin and compile a file?

LianaHus: to run tests

yann300: it is already activated when tests loads. see init.js

yann300: and anyway remix-tests take care of compiling, so we only need the compile tab for getting the compiler version, not directly for compiling

Issue #2246, opened 2019-08-08
Update text on readthedocs run


ryestew:
The text on this page https://remix-ide.readthedocs.io/en/latest/run.html needs some editing.

LianaHus:
@ryestew should i keep this?

PR #2367, opened 2019-09-25
docs: FAQ, update run, update import

ryestew:
addresses: https://github.com/ethereum/remix-ide/issues/2246
& the issue ( that I think was closed) about outdated images.
& and adds to the FAQ


docs/FAQ.md
```diff
@@ -17,6 +17,10 @@ Please change the compiler version in Solidity Plugin to the newer one or use an
 
 **A:**  Try a different browser or a newer solidity compiler version.
 
+**Q:** How to verify a contract that imports other contracts?
+
+**A:**  Use the flattener plugin.
```
yann300: perhaps should we explain what it does and the reason why we should use the flattener plugin

docs/import.md
```diff
@@ -25,15 +30,45 @@ This method uses **remixd** - the remix daemon.  Please go to the [remixd tutori
...
+import 'bzz-raw://5766400e5d6d822f2029b827331b354c41e0b61f73440851dd0d06f603dd91e5';
+```
+
+Importing from IPFS
+--------------------
+
+Files can be imported from IPFS. 
+
+```
+import 'dweb:/ipfs/Qmdyq9ZmWcaryd1mgGZ4PttRNctLGUSAMpPqufsk6uRMKh';
```
yann300: i think `ipfs://Qmdyq9ZmWcaryd1mgGZ4PttRNctLGUSAMpPqufsk6uRMKh` will work but not sure about
`dweb:/ipfs/Qmdyq9ZmWcaryd1mgGZ4PttRNctLGUSAMpPqufsk6uRMKh`

LianaHus: right. 
ipfs://<ipfs_hash> is the right pattern for now. although we should add <ipfs_hash> as well. Same in swarm

PR #2384, opened 2019-10-01
fixed editing gists from last session

LianaHus:


src/app/files/file-explorer.js
```diff
@@ -441,35 +441,42 @@ fileExplorer.prototype.toGist = function (id) {
     }
   }
 
+  async function getOriginalFiles (id) {
```
GrandSchtroumpf: Maybe add a comment with jsdoc synthax so we can understand what this function is doing : 
```
/**
 * This function is awesome
 * @params id The if of....
 */
async function getOriginalFiles (id) { ... }
```

src/app/files/file-explorer.js
```diff
@@ -441,35 +441,42 @@ fileExplorer.prototype.toGist = function (id) {
...
-          token: tokenAccess
-        })
+        const description = 'Created using remix-ide: Realtime Ethereum Contract Compiler and Runtime. \n Load this file by pasting this gists URL or ID at https://remix.ethereum.org/#version=' +
+          queryParams.get().version + '&optimize=' + queryParams.get().optimize + '&gist='
+        const gists = new Gists({ token: tokenAccess })
+
         if (id) {
-          const fileList = Object.keys(this.files.origGistFiles)
-          const updatedFileList = Object.keys(packaged)
+          let originalFileList = getOriginalFiles(id)
```
GrandSchtroumpf: You're ne reassigning this value later. You can use const

src/app/files/file-explorer.js
```diff
@@ -441,35 +441,42 @@ fileExplorer.prototype.toGist = function (id) {
...
           // Telling the GIST API to remove files
-          const allItems = fileList
+          const updatedFileList = Object.keys(packaged)
+          const allItems = Object.keys(originalFileList)
             .filter(fileName => updatedFileList.indexOf(fileName) === -1)
             .reduce((acc, deleteFileName) => ({
               ...acc,
               [deleteFileName]: null
-            }), this.files.origGistFiles)
+            }), originalFileList)
```
GrandSchtroumpf: I've read an article that says that it's actually not a good idea to use spead operator (...) inside a reduce as it's increase the complexity of the function from On to On² (spread operator iterate on each element). Here, the amount of elements is small so it's ok.
You could do that to improve it : 
```typescript
reduce((acc, deleteFileName) => {
  delete acc[deleteFileName]
  return acc
}, originalFileList)
```

PR #2389, opened 2019-10-02
added istanbul to evm versions

LianaHus:


src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -151,6 +151,7 @@ class CompilerContainer {
       <select onchange="${this.onchangeEvmVersion.bind(this)}" class="custom-select" id="evmVersionSelector">
         <option value="default">compiler default</option>
         <option>petersburg</option>
+        <option>istanbul</option>
```
axic: IMHO it should be above `petersburg` because its never.

yann300:
can we merge this now @LianaHus ?

src/app/tabs/compileTab/compilerContainer.js
LianaHus: good point

PR #2398, opened 2019-10-08
pickup compiler version from pragma

### LINKED ISSUE #2402: Auto pickup compiler version based on pragma solidity spec
Hi All, 

Would it be possible to automatically select the compiler at it's stable version when the compiler is specified at the top of the code? e.g. pragma solidity ^0.4.25;

Annoying to have to constantly adjust it on start up. 

Cheers

Aniket-Engg:
fixes #2402

### DIFF
--- a/src/app/tabs/compileTab/compilerContainer.js
+++ b/src/app/tabs/compileTab/compilerContainer.js
@@ -32,6 +32,9 @@ class CompilerContainer {
    * Update the compilation button with the name of the current file
    */
   set currentFile (name = '') {
+    if (name && name !== '') {
+      this._setCompilerVersionFromPragma(name)
+    }
     if (!this._view.compilationButton) return
     const button = this.compilationButton(name.split('/').pop())
     yo.update(this._view.compilationButton, button)
@@ -111,6 +114,23 @@ class CompilerContainer {
       el.setAttribute('disabled', 'true')
     }
     return el
+  }
+
+  _setCompilerVersionFromPragma (filename) {
+    this.compileTabLogic.fileManager.getFile(filename).then(data => {
+      const pragmaArr = data.match(/(pragma solidity (.+?);)/g)
+      if (pragmaArr && pragmaArr.length === 1) {
+        const pragmaStr = pragmaArr[0].replace('pragma solidity', '').trim()
+        const pragma = pragmaStr.substring(0, pragmaStr.length - 1)
+        const fixedVersions = this.data.allversions.filter(obj => !obj.prerelease).map(obj => obj.version)
+        const compilerToLoad = semver.maxSatisfying(fixedVersions, pragma)
+        const compilerPath = this.data.allversions.filter(obj => !obj.prerelease && obj.version === compilerToLoad)[0].path
+        if (this.data.selectedVersion !== compilerPath) {
+          this.data.selectedVersion = compilerPath
+          this._updateVersionSelector()
+        }
+      }
+    })
   }
 
   _retrieveVersion () {
@@ -243,6 +263,9 @@ class CompilerContainer {
 
   compile (event) {
     if (this.config.get('currentFile')) {
+      if (!this.data.selectedVersion.includes('nightly')) {
+        this._setCompilerVersionFromPragma(this.config.get('currentFile'))
+      }
       this.compileTabLogic.runCompiler()
     }
   }


src/app/tabs/compile-tab.js
```diff
@@ -103,7 +103,21 @@ class CompileTab extends ViewPlugin {
     })
 
     this.fileManager.events.on('currentFileChanged', (name) => {
-      this.compilerContainer.currentFile = name
+      this.fileManager.events.on('currentFileChanged', (name) => {
+        this.fileManager.getFile(name).then(data => {
+          let pragmaArr = data.match(/(pragma solidity (.+?);)/g)
+          if (pragmaArr && pragmaArr.length === 1) {
+            let pragma = pragmaArr[0].replace('pragma solidity', '').trim()
```
yann300: I don't think this is handling the following case : `pragma solidity >=0.4.0 <0.6.0;`

Aniket-Engg: Currently, for a range in pragma, I am thinking to use default version as we are not sure which version to load. Looking for the suggestions @yann300

yann300: i think we should select the latest available from the given range

Aniket-Engg:
I checked it with various pragma locally on IDE, it worked like this:

- Pragma: >0.4.0<=0.5.0,   Loaded compiler version: 0.5.0
- Pragma: >0.4.0<0.5.0,   Loaded compiler version: 0.4.26
- Pragma: >0.4.0<=0.6.0 ,  Loaded compiler version: 0.5.12
- Pragma: >0.4.0<0.6.0,   Loaded compiler version: 0.5.12
- Pragma: >0.4.26<0.5.0,   Loaded compiler version: already loaded version reloaded
- Pragma: >=0.4.26<0.5.0,   Loaded compiler version: 0.4.26
- Pragma: >0.4.26<=0.5.0,   Loaded compiler version: 0.5.0
- Pragma: >0.5.0,   Loaded compiler version: 0.5.12
- Pragma: >=0.5.0,   Loaded compiler version: 0.5.12
- Pragma: ^0.4.18,   Loaded compiler version: 0.4.18
- Pragma: 0.5.3,   Loaded compiler version: 0.5.3

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -289,11 +289,37 @@ class CompilerContainer {
...
   }
 
-  _updateVersionSelector () {
+  _updateVersionSelector (pragma = null) {
     // update selectedversion of previous one got filtered out
     if (!this._shouldBeAdded(this.data.selectedVersion)) {
       this.data.selectedVersion = this.data.defaultVersion
     }
+    if (pragma && pragma.start) {
+      let allversions = this.data.allversions.filter(build => (!build.prerelease))
```
GrandSchtroumpf: use `const` as your variable is not going to change.

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -289,11 +289,37 @@ class CompilerContainer {
            (version.includes('nightly') && this._view.includeNightlies.checked)
   }
 
-  _updateVersionSelector () {
+  _updateVersionSelector (pragma = null) {
```
GrandSchtroumpf: Maybe add a comment with the type of object you want to get : 
```javascript
/**
 * Update ***
 * @params pragma {Object} Version of solidity compiler
*/
```
Here is a reference to [@param tag](https://devdocs.io/jsdoc/tags-param)
Even better would be to use the [@type tag](https://devdocs.io/jsdoc/tags-type) so we know how the object should look like.

src/app/tabs/compileTab/compilerContainer.js
GrandSchtroumpf: Same for `startIndex`

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -289,11 +326,54 @@ class CompilerContainer {
...
+                endIndex = -1
+              }
+              maxIndex = endIndex
+            }
+          }
+          if (maxIndex > -1) {
+            selectedVersion = allversions[maxIndex].path
+          }
+        }
+        if (this.data.selectedVersion === selectedVersion) {
```
LianaHus: maybe it's not selected version by this logic but also fine. for example for a range from 0.3.0 to 0.6.0 you can pick 0.5.12 but 0.5.0 is also fine.

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -113,6 +116,39 @@ class CompilerContainer {
     return el
   }
 
+  _setCompilerVersionFromPragma (filename) {
```
LianaHus: this is not working for nightly builds.

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -113,6 +116,39 @@ class CompilerContainer {
...
+    this.compileTabLogic.fileManager.getFile(filename).then(data => {
```
GrandSchtroumpf: It feel very "manual" couldn't we use the AST to get the information instead of manipulating strings ?

LianaHus: +1 for AST and I would suggest using semver we already use it in the project

Aniket-Engg: okay, will do that

src/app/tabs/compileTab/compilerContainer.js
Aniket-Engg: selected version is here something which will be selected from compiler dropdown

src/app/tabs/compileTab/compilerContainer.js
Aniket-Engg: Thanks fro semver suggestion @LianaHus , Now it works in less than half code

Aniket-Engg:
> I checked it with various pragma locally on IDE, it worked like this:
> 
> * Pragma: >0.4.0<=0.5.0,   Loaded compiler version: 0.5.0
> * Pragma: >0.4.0<0.5.0,   Loaded compiler version: 0.4.26
> * Pragma: >0.4.0<=0.6.0 ,  Loaded compiler version: 0.5.12
> * Pragma: >0.4.0<0.6.0,   Loaded compiler version: 0.5.12
> * Pragma: >0.4.26<0.5.0,   Loaded compiler version: already loaded version reloaded
> * Pragma: >=0.4.26<0.5.0,   Loaded compiler version: 0.4.26
> * Pragma: >0.4.26<=0.5.0,   Loaded compiler version: 0.5.0
> * Pragma: >0.5.0,   Loaded compiler version: 0.5.12
> * Pragma: >=0.5.0,   Loaded compiler version: 0.5.12
> * Pragma: ^0.4.18,   Loaded compiler version: 0.4.18
> * Pragma: 0.5.3,   Loaded compiler version: 0.5.3

Just to add, now  Pragma: ^0.4.18, Loaded compiler version: 0.5.12

PR #2403, opened 2019-10-14
use white-space:pre-wrap for spaces make effect

helderjnpinto:
The white spaces are not rendered: 
The normal behavior for the display of whitespaces is to compress them into a single one, which is then displayed.

There are two exceptions from that:

The pre tag, which keeps the whitespaces as entered.
Setting the CSS property white-space: pre; (respectively pre-wrap or pre-line)

Ex: 
image

in the code we can add a simple style for childs with white-space style.

Result in to this render: 

image


LianaHus:
@helderjnpinto thanks for the PR

PR #2407, opened 2019-10-19
Increase time waiting for element visible in unit testing `deploy ballot`

icegotcha:
From 10 sec to 30 sec


LianaHus:
@icegotcha thanks for PR
could you please change it to 15 000 (15sec)?

PR #2426, opened 2019-11-06
Dark theme update

alexcherman:


alexcherman:
> text-dark in the theme is .text-dark {
> color: #2a2c3f !important;
> }
> But this should be a light color in a dark theme. ( and a dark color in a light theme).

@ryestew  I thought we need to switch .text-dark and .text-light colors in dark theme. But if you want I can revert this change

src/app/components/plugin-manager-component.js
```diff
@@ -196,7 +196,7 @@ class PluginManagerComponent extends ViewPlugin {
       <div id='pluginManager'>
         <header class="form-group ${css.pluginSearch} plugins-header py-3 px-4 border-bottom">
           <input onkeyup="${e => this.filterPlugins(e)}" class="${css.pluginSearchInput} form-control" placeholder="Search">
-          <button onclick="${_ => this.openLocalPlugin()}" class="${css.pluginSearchButton} btn bg-transparent text-dark border-0 mt-2 text-underline">
+          <button onclick="${_ => this.openLocalPlugin()}" class="${css.pluginSearchButton} btn bg-transparent text-dark border-0 mt-2 text-underline plugins-connect-btn">
```
ryestew: I don't think you need to add this class - but to use the color of plugins-connect-btn as the color of text-dark.

ryestew:
text-dark - is still too dark - please use #babbcc

alexcherman:
> text-dark - is still too dark - please use #babbcc

ok, but "Connect to a Local Plugin" button has #A2A3BD color

alexcherman:
@ryestew Updated text-dark color

ryestew:
Actually there is an issue the alpha and the beta tags are different colors - alpha is missing a class that is in beta.

alexcherman:
@ryestew I'll check it

Issue #2386, opened 2019-10-02
FE: issues with context menu


LianaHus:
- [ ] add *Create Folder* and *Create File* to the context menu of **Browser**
- [ ] Check if file name already exists and use *createNonClashingNameWithPrefix()* of helper
- [ ] check if folder name already exists and add a error or suggest versions
- [ ] *rename* of gist id folder should be prohibitted
- [ ] *create folder* doesn't work when current folder is gist id, openes *create file* modal instead

LianaHus:
https://github.com/ethereum/remix-ide/issues/2460

PR #2427, opened 2019-11-08
Fixing bugs of FE #2386

LianaHus:


src/app/files/file-explorer.js
```diff
@@ -579,16 +584,15 @@ fileExplorer.prototype.copyFiles = function () {
...
 
-fileExplorer.prototype.createNewFile = function (parentFolder) {
+fileExplorer.prototype.createNewFile = function (parentFolder = 'browser') {
   let self = this
   modalDialogCustom.prompt('Create new file', 'File Name (e.g Untitled.sol)', 'Untitled.sol', (input) => {
-    helper.createNonClashingName(input, self.files, (error, newName) => {
-      if (error) return modalDialogCustom.alert('Failed to create file ' + newName + ' ' + error)
-      const currentPath = !parentFolder ? self._deps.fileManager.currentPath() : parentFolder
-      newName = currentPath ? currentPath + '/' + newName : self.files.type + '/' + newName
+    if (input === '') input = 'New file'
```
GrandSchtroumpf: I would rather do `if (input)` like that you check for null & undefined also.

src/app/files/file-explorer.js
```diff
@@ -601,14 +605,24 @@ fileExplorer.prototype.createNewFile = function (parentFolder) {
 
 fileExplorer.prototype.createNewFolder = function (parentFolder) {
   let self = this
-  modalDialogCustom.prompt('Create new folder', '', '', (input) => {
+  modalDialogCustom.prompt('Create new folder', '', 'New folder', (input) => {
+    if (input === '') {
```
GrandSchtroumpf: Same as above

src/app/files/file-explorer.js
```diff
@@ -601,14 +605,24 @@ fileExplorer.prototype.createNewFile = function (parentFolder) {
...
+      tooltip('Failed to create folder. The name can not be empty')
+      return false
```
GrandSchtroumpf: If you're using `return` to stop the process of the scope I would rath do `return tooltip(...)`. Returning `false` feels like it's used somewhere. If this is the case then forget this comment.

src/app/files/fileProvider.js
```diff
@@ -131,24 +133,43 @@ class FileProvider {
...
-      return false
-    }
-    const stat = window.remixFileSystem.statSync(unprefixedpath)
-    try {
-      if (stat.isDirectory()) {
-        window.remixFileSystem.rmdirSync(unprefixedpath, console.log)
-      } else {
-        window.remixFileSystem.unlinkSync(unprefixedpath, console.log)
+    path = this.removePrefix(path)
+    if (window.remixFileSystem.existsSync(path)) {
```
GrandSchtroumpf: What ??? We inject the FileSystem as a global value in the file system o.O"

src/app/files/fileProvider.js
```diff
@@ -131,24 +133,43 @@ class FileProvider {
...
+      const stat = window.remixFileSystem.statSync(path)
+      try {
+        if (!stat.isDirectory()) {
+          window.remixFileSystem.unlinkSync(path, console.log)
+          this.event.trigger('fileRemoved', [this._normalizePath(path)])
+          return true
+        } else {
+          let items = window.remixFileSystem.readdirSync(path)
```
GrandSchtroumpf: You're not reassigning it. So you can use `const`

src/app/files/fileProvider.js
```diff
@@ -131,24 +133,43 @@ class FileProvider {
...
+          if (items.length !== 0) {
+            items.forEach((item, index) => {
+              let curPath = path + '/' + item
```
GrandSchtroumpf: Maybe : 
```javascript
const currPath = `${path}/${item}`
```

LianaHus: why it is better?

GrandSchtroumpf: You don't reassign it so you can you `const`. 
Personally I find the `${a}divder${b}` writing more readable than using concatenation (just suggesting here).

PR #2443, opened 2019-11-15
[WIP] Adding remixd git plugin

edisinovcic:


src/remixAppManager.js
```diff
@@ -214,6 +214,18 @@ export class RemixAppManager extends PluginEngine {
       version: '0.1.0-beta',
       location: 'sidePanel'
     }
+    const remixdgit = {
```
LianaHus: Could you please use camel case here?

src/remixAppManager.js
```diff
@@ -214,6 +214,18 @@ export class RemixAppManager extends PluginEngine {
...
+      name: 'Remixd.git',
+      displayName: 'Remixd git interface',
+      description: 'All git related stuff',
+      events: [],
+      methods: [],
+      url: 'remixd-git.surge.sh',
+      icon: 'https://git-scm.com/images/logos/1color-lightbg@2x.png',
```
LianaHus: We need a quadratic icon. This is to show it on vertical icons panel
Something like https://images.app.goo.gl/gTQyYP7jhCVBSuL49

src/remixAppManager.js
```diff
@@ -214,6 +214,18 @@ export class RemixAppManager extends PluginEngine {
       version: '0.1.0-beta',
       location: 'sidePanel'
     }
+    const remixdgit = {
+      name: 'Remixd.git',
```
LianaHus: I'm not sure the bame can contain '.'
@GrandSchtroumpf ?

edisinovcic: Will remove it completely :)

src/remixAppManager.js
edisinovcic: Sure

PR #2457, opened 2019-11-22
Do not run with workers

yann300:


LianaHus:
https://github.com/ethereum/remix-ide/issues/2461

PR #2459, opened 2019-11-26
unit testing plugin info updated

Aniket-Engg:


src/app/tabs/test-tab.js
```diff
@@ -242,15 +242,14 @@ module.exports = class TestTab extends ViewPlugin {
...
-          For more details, see
-          How to test smart contracts guide in our documentation.
-          <br/>
+          Test your smart contract in Solidity.<br/><br/>
+          <ol>
+            <li> To get started, click on "Generate test file" button</li>
+            <li> To write tests, visit our <a href="https://remix-ide.readthedocs.io/en/latest/unittesting.html#write-tests" target="_blank"> documentation </a></li>
+            <li> To run tests, select file(s) and click on "Run Tests" button</li>
+          </ol>
+          To run unit tests in your Continuous Integration & as CLI, use the stand alone NPM module <a href="https://www.npmjs.com/package/remix-tests" target="_blank">remix-tests</a>.<br/>
```
ryestew: I'd like that the next line "To get support ..." is in a p tag - so the spacing is bigger - so remove the br and wrap it in a p tag

ryestew: Also here's a very small edit on this line:
To run unit tests in your Continuous Integration and as a CLI, use the stand alone NPM module

ryestew: Or to save space - put the 2 sentences that start with To... in the same paragraph.

PR #2462, opened 2019-11-28
Style deploy&run module in the dark theme

alexcherman:


ryestew:
Here are some comments in the graphic:

annotated-runtab-issues

ryestew:
The lables are still too dark - in the figma I have they a lighter color - check the trello card for the details.  Also the file assets/js/0.7.7/app.js  and I'm not sure why.

Also the PR is now 12 files it seems to now include the settings module.  And there are some issues there too with the input on that tab - it is not visible as an input.

image

the input for the token is not visible.

src/app/tabs/runTab/settings.js
```diff
@@ -127,7 +127,7 @@ class SettingsUI {
     `
 
     const el = yo`
-      <div class="${css.settings}">
+      <div class="${css.settings} module-section">
```
LianaHus: what is module-section ?

src/app/tabs/runTab/settings.js
```diff
@@ -106,17 +106,17 @@ class SettingsUI {
...
     `
 
     const valueEl = yo`
       <div class="${css.crow}">
-        <div class="${css.col1_1}">Value</div>
+        <div class="${css.settingsLabel} form-control-label">Value</div>
         <div class="${css.gasValueContainer}">
           <input type="text" class="form-control ${css.gasNval} ${css.col2}" id="value" value="0" title="Enter the value and choose the unit">
-          <select name="unit" class="form-control p-1 ${css.gasNvalUnit} ${css.col2_2}" id="unit">
+          <select name="unit" class="form-control p-1 ${css.gasNvalUnit} ${css.col2_2} run-unit-input" id="unit">
```
LianaHus: "run-unit-input" is this from theme again?

src/app/tabs/runTab/settings.js
alexcherman: This class necessary to style elements only in dark and themes

src/app/tabs/runTab/settings.js
alexcherman: The same as above

alexcherman:
@ryestew I left comment about label color in the trello card. Could you check it please?

alexcherman:
@ryestew I changed styles in settings tab because this file had styles for input and it affected the inputs in the whole project. It shouldn't be like that. Because of that I wasn't able to align inputs in the Deploy&Run tab. I'll make input in the settings tab visible in the next PR. 

As for the changes in the app.js, I'll revert these changes.

ryestew:
the label is still the wrong color:
.form-control-label {
    color: #747B90;
Should be 959BAD
And maybe more that were the darker should be the lighter color.

alexcherman:
> the label is still the wrong color:
> .form-control-label {
> color: #747B90;
> Should be 959BAD
> And maybe more that were the darker should be the lighter color.

@ryestew  Check the trello card please that you created. I left comment there.

ryestew:
In the run & deploy module - the transact button looks like it has a larger font than the delegate or vote button.  

Screen Shot 2019-12-06 at 3 46 15 PM

src/app/tabs/runTab/settings.js
LianaHus: why can't it be in all of them?

src/app/tabs/runTab/contractDropdown.js
```diff
@@ -51,10 +51,11 @@ class ContractDropdownUI {
     this.atAddressButtonInput = yo`<input class="${css.input} ${css.ataddressinput} ataddressinput form-control" placeholder="Load contract from Address" title="address of contract" oninput=${this.atAddressChanged.bind(this)} />`
     this.selectContractNames = yo`<select class="${css.contractNames} custom-select" disabled></select>`
 
-    this.createPanel = yo`<div class="${css.button}"></div>`
+    this.createPanel = yo`<div class="${css.deployDropdown}"></div>`
     this.orLabel = yo`<div class="${css.orLabel}">or</div>`
     let el = yo`
-      <div class="${css.container}">
+      <div class="${css.container} module-section">
+        <div class="${css.settingsLabel} form-control-label">Contract</div>
```
LianaHus: is form-control-label boostrap class?

src/app/tabs/runTab/settings.js
LianaHus: why can't it be in all of them?

src/app/tabs/runTab/contractDropdown.js
alexcherman: .form-control-label class is not bootstrap class but my custom class. Bootstrap styles label using tag selector. I added this because you didn't use <label> tag where it should be used. Sometime you used just <div> tag instead of <label> tag.

ryestew: When we have used a div tag instead of label - it was bad form on our side.  The tags need to updated to be <label> tags.

Issue #2461, opened 2019-11-27
Solidity compiler: "browser maximum call stack size exceeded"


LianaHus:
**Currently working versions:**
OSX + chrome are fine for every version
Win + chrome are fine for > 0.5.2 + 0.4.26
Ubuntu + chrome are fine for 0.4.26 + 0.5.3 + 0.5.4 + 0.5.5
Firefox is fine

**Solutions**:
To fix the problem with chrome in certain browsers we should load blacklisted versions without web worker. Keep in touch with the solidity team to remove from the list rereleased versions later.

Also, we can cash versions already loaded once by user locally to make them load fast in a future.

Aniket-Engg:
Closed #2463

PR #2463, opened 2019-11-29
workers should be used only with whitelisted versions for chrome

LianaHus:
fixing https://github.com/ethereum/remix-ide/issues/2461

### DIFF
--- a/src/app/tabs/compileTab/compilerContainer.js
+++ b/src/app/tabs/compileTab/compilerContainer.js
@@ -400,7 +400,7 @@ class CompilerContainer {
     // Workers cannot load js on "file:"-URLs and we get a
     // "Uncaught RangeError: Maximum call stack size exceeded" error on Chromium,
     // resort to non-worker version in that case.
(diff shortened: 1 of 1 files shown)


LianaHus:
ok for OSX, Linux
not ok for Win

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -389,6 +409,14 @@ class CompilerContainer {
     }
   }
 
+  _retrieveOS () {
+    let osName = 'Unknown OS'
+    if (navigator.platform.indexOf('Win') !== -1) osName = 'Windows'
+    if (navigator.platform.indexOf('Mac') !== -1) osName = 'MacOS'
```
Aniket-Engg: Not sure if we should check for MacOS as all versions are working fine with it. Also I would prefer to use `if - else if - else`

LianaHus: updated

PR #2469, opened 2019-12-05
disable Compile button on loading

LianaHus:


src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -37,6 +37,7 @@ class CompilerContainer {
     }
     if (!this._view.compilationButton) return
     const button = this.compilationButton(name.split('/').pop())
+    this._disableCompileBtn(!name || name === '')
```
GrandSchtroumpf: if `name === '' // true` then `!name // true`. So you don't need `|| name === ''`.
In Javascript : 
`!0 // true`
`!"" // true`
`!null // true`
`!undefined // true`
BUT
`![] // false`
`!{} // false`

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -103,19 +106,31 @@ class CompilerContainer {
    */
   compilationButton (name) {
```
GrandSchtroumpf: You can add a default value directly inside the parameter : 
```javascript
comilationButton (name = '') {
  const displayed = name || '<no file selected>'
  const disabled = name ? '' : 'disabled'
  ...
}
```

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -103,19 +106,31 @@ class CompilerContainer {
...
+    const compileBtn = yo`
+      <button id="compileBtn" class="btn btn-primary btn-block ${disabled}" title="Compile" onclick="${this.compile.bind(this)}">
         <span>${this._view.compileIcon} Compile ${displayed}</span>
       </button>
-    </div>`
-    if (name === '') {
-      el.setAttribute('disabled', 'true')
-    }
+    `
+    let el = yo`
```
GrandSchtroumpf: To make it more "simple" you can also do : 
```javascript
return yo`<div class="px-2 mt-2 pb-0 d-flex">${compileBtn}</div>`
```

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -103,19 +106,31 @@ class CompilerContainer {
...
+      <div class="px-2 mt-2 pb-0 d-flex">
+        ${compileBtn}
+      </div>
+    `
     return el
   }
 
+  _disableCompileBtn (disable) {
```
GrandSchtroumpf: Can you add a comment on the method ? Also as the param is a boolean, I would recommend `isDisabled` or `shouldDisable`
```javascript
/**
 * Toggle the compile button
 * @param {boolean} isDisabled Should disable or enable
 */
```

src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -103,19 +106,31 @@ class CompilerContainer {
...
+    let btn = document.querySelector("button[id='compileBtn']")
```
GrandSchtroumpf: Id are supposed to be unique, better to do: `document.getElementById('compileBtn')` (it's much faster)

src/app/tabs/compileTab/compilerContainer.js
LianaHus: I don't think it needs comment, the name is clear enough.

PR #2475, opened 2019-12-06
use Web3 utils instead of execution context web3

iurimatias:
part of moving to remix as web3 provider


src/app/tabs/runTab/model/dropdownlogic.js
```diff
@@ -7,6 +7,7 @@ var typeConversion = remixLib.execution.typeConversion
 var txExecution = remixLib.execution.txExecution
 var CompilerAbstract = require('../../../compiler/compiler-abstract')
 var EventManager = remixLib.EventManager
+var Web3 = require('web3')
```
LianaHus: could you please always use const if possible?

GrandSchtroumpf: It's usually better for tree shaking to be more precise about the import: `const { utils } from 'web3'` or with require `const { utils } = require('web3')`

PR #2478, opened 2019-12-09
bump version to 0.9.2

LianaHus:


package.json
```diff
@@ -22,6 +22,7 @@
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-plugin-yo-yoify": "^2.0.0",
     "babelify": "^10.0.0",
+    "bignumber.js": "git+https://github.com/frozeman/bignumber.js-nolookahead.git#57692b3ecfc98bbdd6b3a516cb2353652ea49934",
```
GrandSchtroumpf: I don't like the idea of adding bignumber.js as a npm dependency. Especially when it's not event a node module. If web3.js is using it, let's import it itself, but I would rather not define it in remix-ide

PR #2480, opened 2019-12-11
Fix package remixd inclusion as both a dev and production dependency

ziogaschr:


ziogaschr:
Good point @yann300, though there are no changes as the package remains in package.json under dependencies. I just removed the duplicate from devDependencies

PR #2490, opened 2019-12-13
check and if possible use worker for unit tests compilation

LianaHus:


src/app/compiler/compiler-utils.js
```diff
@@ -0,0 +1,43 @@
...
+  // Following restrictions should be deleted when Solidity will release fixed versions of compilers.
+  // See https://github.com/ethereum/remix-ide/issues/2461
+  const isChrome = !!window.chrome
+  const os = retrieveOS()
+  // define a whitelist for Linux
+  const linuxWL = ['0.4.26', '0.5.3', '0.5.4', '0.5.5']
+  const version = semver.coerce(selectedVersion)
+  // defining whitelist for chrome
+  let isFromWhiteList = false
+  switch (os) {
```
GrandSchtroumpf: What about macOS, Android, iOS ?

src/app/compiler/compiler-utils.js
```diff
@@ -0,0 +1,43 @@
...
+      isFromWhiteList = true
+  }
+  return browserSupportWorker() && (!isChrome || (isChrome && isFromWhiteList))
+}
+
+function browserSupportWorker () {
+  return document.location.protocol !== 'file:' && Worker !== undefined
+}
+
+function retrieveOS () {
```
GrandSchtroumpf: same as above

src/app/compiler/compiler-utils.js
```diff
@@ -0,0 +1,43 @@
...
+  let osName = 'Unknown OS'
+  if (navigator.platform.indexOf('Win') !== -1) {
+    osName = 'Windows'
+  } else if (navigator.platform.indexOf('Linux') !== -1) {
+    osName = 'Linux'
+  }
```
GrandSchtroumpf: You can write it like that directly : 
```typescript
if (navigator.platform.indexOf('Win') !== -1) {
  return 'Windows'
} else if (navigator.platform.indexOf('Linux') !== -1) {
  return 'Linux'
} else {
  return 'Unknow OS'
}
```

src/app/tabs/test-tab.js
```diff
@@ -169,14 +169,20 @@ module.exports = class TestTab extends ViewPlugin {
...
-        this.loading.hidden = true
-        callback(error)
-      }, (url, cb) => {
-        return this.compileTab.compileTabLogic.importFileCb(url, cb)
-      })
+      remixTests.runTestSources(
+        runningTest,
+        currentCompilerUrl,
+        canUseWorker(this.compileTab.getCurrentVersion()),
+        (result) => { this.testCallback(result) },
```
GrandSchtroumpf: You can remove the `{}` :
```typescript
(result) => this.testCallback(result),
```

src/app/compiler/compiler-utils.js
LianaHus: Rob said macOS is fine. why should we check for android?

src/app/compiler/compiler-utils.js
LianaHus: I opt for readability here

PR #2491, opened 2019-12-15
refactor execution context

iurimatias:
stops relying on global execution context and instead pass it as parameter

### DIFF
--- a/src/app/files/compiler-metadata.js
+++ b/src/app/files/compiler-metadata.js
@@ -1,5 +1,4 @@
 'use strict'
-var executionContext = require('../../execution-context')
 var CompilerAbstract = require('../compiler/compiler-abstract')
 import { Plugin } from '@remixproject/engine'
 import * as packageJson from '../../../package.json'
@@ -12,9 +11,10 @@ const profile = {
 }
 
 class CompilerMetadata extends Plugin {
-  constructor (fileManager, config) {
+  constructor (executionContext, fileManager, config) {
     super(profile)
     var self = this
+    self.executionContext = executionContext
     self.fileManager = fileManager
     self.config = config
     self.networks = ['VM:-', 'main:1', 'ropsten:3', 'rinkeby:4', 'kovan:42', 'görli:5', 'Custom']
@@ -97,7 +97,7 @@ class CompilerMetadata extends Plugin {
       var provider = self.fileManager.currentFileProvider()
       var path = self.fileManager.currentPath()
       if (provider && path) {
-        executionContext.detectNetwork((err, { id, name } = {}) => {
+        self.executionContext.detectNetwork((err, { id, name } = {}) => {
           if (err) {
             console.log(err)
           } else {
--- a/src/app/tabs/runTab/model/dropdownlogic.js
+++ b/src/app/tabs/runTab/model/dropdownlogic.js
@@ -7,6 +7,7 @@ var typeConversion = remixLib.execution.typeConversion
 var txExecution = remixLib.execution.txExecution
 var CompilerAbstract = require('../../../compiler/compiler-abstract')
 var EventManager = remixLib.EventManager
+var Web3 = require('web3')
 
 class DropdownLogic {
   constructor (fileManager, compilersArtefacts, config, editor, udapp, filePanel, runView) {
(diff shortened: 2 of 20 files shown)


src/app/files/compiler-metadata.js
```diff
@@ -97,7 +97,7 @@ class CompilerMetadata extends Plugin {
       var provider = self.fileManager.currentFileProvider()
       var path = self.fileManager.currentPath()
       if (provider && path) {
-        executionContext.detectNetwork((err, { id, name } = {}) => {
+        self.executionContext.detectNetwork((err, { id, name } = {}) => {
```
GrandSchtroumpf: you should be able to use `this` here I think

src/app/tabs/runTab/model/dropdownlogic.js
```diff
@@ -106,25 +107,25 @@ class DropdownLogic {
 
   fromWei (value, doTypeConversion, unit) {
     if (doTypeConversion) {
-      return executionContext.web3().fromWei(typeConversion.toInt(value), unit || 'ether')
+      return Web3.utils.fromWei(typeConversion.toInt(value), unit || 'ether')
```
GrandSchtroumpf: Why not using ethers.js instead as we're moving out of Web3 to ethers.js ?

src/app/ui/txLogger.js
```diff
@@ -205,7 +204,7 @@ function log (self, tx, receipt) {
   }
 }
 
-function renderKnownTransaction (self, data) {
+function renderKnownTransaction (self, data, executionContext) {
```
GrandSchtroumpf: Why not using `this` or `self` for `executionContext` as you're doing in the other files ?

src/app/ui/universal-dapp-ui.js
```diff
@@ -180,13 +180,13 @@ UniversalDAppUI.prototype.getCallButton = function (args) {
           cb(txFeeText, priceStatus)
         },
         (cb) => {
-          executionContext.web3().eth.getGasPrice((error, gasPrice) => {
+          self.executionContext.web3().eth.getGasPrice((error, gasPrice) => {
```
GrandSchtroumpf: Can you not use `this` here ?

src/app/ui/universal-dapp-ui.js
```diff
@@ -264,7 +264,7 @@ UniversalDAppUI.prototype.getCallButton = function (args) {
               }
             }
             if (lookupOnly) {
-              var decoded = decodeResponseToTreeView(executionContext.isVM() ? txResult.result.execResult.returnValue : ethJSUtil.toBuffer(txResult.result), args.funABI)
+              var decoded = decodeResponseToTreeView(self.executionContext.isVM() ? txResult.result.execResult.returnValue : ethJSUtil.toBuffer(txResult.result), args.funABI)
```
GrandSchtroumpf: Can you not use `this` here instead of `self` ?

src/lib/cmdInterpreterAPI.js
```diff
@@ -62,14 +62,14 @@ class CmdInterpreterAPI {
   debug (hash, cb) {
     var self = this
     delete self.d
-    executionContext.web3().eth.getTransaction(hash, (error, tx) => {
+    self.executionContext.web3().eth.getTransaction(hash, (error, tx) => {
```
GrandSchtroumpf: Use `this` instead

src/lib/cmdInterpreterAPI.js
```diff
@@ -62,14 +62,14 @@ class CmdInterpreterAPI {
...
       if (error) return cb(error)
       var debugSession = new RemixDebug({
         compilationResult: () => {
           return self._deps.compilersArtefacts['__last'].getData()
         }
       })
-      debugSession.addProvider('web3', executionContext.web3())
+      debugSession.addProvider('web3', self.executionContext.web3())
```
GrandSchtroumpf: `this`

src/app/tabs/runTab/model/dropdownlogic.js
iurimatias: I just wanted to get rid of the using executionContext here avoiding breaking changes as maximum as possible. I agree later we can refactor this to either use web3-utils directly or something else like etherjs, it's quite possible this whole method might disapear though so optimizing now could be premature

src/app/ui/txLogger.js
iurimatias: keeping the same behaviour for now

iurimatias:
@LianaHus it's fixed now

iurimatias:
closing since the changes in this PR were included in #2506

PR #2493, opened 2019-12-17
recorder fixed

### LINKED ISSUE #2489: Remix Recorder doesn't use the abi
When using the recorder, the `to` of a transaction generate `created{undefined}` instead of `created{timestamp}`.
Also, when using a transaction with `to` is triggered by the recorder, the ABI of the targeted contract is not retrieved and cause the recorder to stop.

Aniket-Engg:
fixes #2489

### DIFF
--- a/src/app/tabs/runTab/model/recorder.js
+++ b/src/app/tabs/runTab/model/recorder.js
@@ -73,10 +73,10 @@ class Recorder {
       if (error) return console.log(error)
       if (call) return
 
-      var address = executionContext.isVM() ? txResult.result.createdAddress : txResult.result.contractAddress
+      const  rawAddress = executionContext.isVM() ? txResult.result.createdAddress : txResult.result.contractAddress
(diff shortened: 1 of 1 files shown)


src/app/tabs/runTab/model/recorder.js
```diff
@@ -76,6 +76,7 @@ class Recorder {
       var address = executionContext.isVM() ? txResult.result.createdAddress : txResult.result.contractAddress
       if (!address) return // not a contract creation
       address = this.addressToString(address)
+      address = ethutil.toChecksumAddress(address)
```
GrandSchtroumpf: Better not to reassign the same string. You could have something like : 
```typescript
const rawAddress = executionContext.isVM() ? txResult.result.createdAddress : txResult.result.contractAddress
if (!rawAddress) return
const stringAddress = this.addressToString(address)
const address = ethutil.toChecksumAddress(stringAddress)
```

Issue #2454, opened 2019-11-22
Default files in file explorer


Aniket-Engg:
At present, we show two default contract files on load: `ballot.sol` & `ballot_test.sol`. I think we should remove `ballot.sol` as it is not a right point for someone to start. It has a for & while loop and even our own solidity analyser shows 3 warnings for it. We should not present something at first place regarding which we warn users. Rather, we should show 3 default contracts with increasing complexity. 

Corresponding test contract (_test.sol) should be loaded only when `Solidity Unit Testing` plugin is enabled.

Aniket-Engg:
Closed by #2507

PR #2507, opened 2019-12-26
example contracts updated

Aniket-Engg:
- New example contracts added
- ballot & ballot_test updated to compile fine with solidity 0.6.1
- Default solidity version set to 0.6.1
Related to #2454


src/app/editor/example-contracts.js
```diff
@@ -1,96 +1,261 @@
 'use strict'
 
-var ballot = `pragma solidity >=0.4.22 <0.6.0;
-contract Ballot {
+var storage = `pragma solidity >=0.4.22 <0.7.0;
```
GrandSchtroumpf: can you use `const` ;)

src/app/editor/example-contracts.js
```diff
@@ -1,96 +1,261 @@
...
-        address delegate;
+        uint weight; // weight is accumulated by delegation
+        bool voted;  // if true, that person already voted
+        address delegate; // person delegated to
+        uint vote;   // index of the voted proposal
     }
+
     struct Proposal {
-        uint voteCount;
+        bytes32 name;   // short name (up to 32 bytes)
```
GrandSchtroumpf: maybe explain why it's better to use bytes here instead of string.

src/app/editor/example-contracts.js
```diff
@@ -1,96 +1,261 @@
...
+        }
+    }
+
+    /** 
+     * @dev Calls winningProposal() function to get the index of the winner contained in the proposals array and then
+     * @return winnerName_ the name of the winner
+     */
+     
+    function winnerName() public view
+            returns (bytes32 winnerName_)
```
GrandSchtroumpf: I think it makes it hard to read to put it on two lines.
Is it a convention to use _ at the end of a output variable ?

Aniket-Engg: Yes, It is used to avoid name collisions. See: https://solidity.readthedocs.io/en/v0.6.0/style-guide.html#avoiding-naming-collisions

Aniket-Engg:
Default Solidity version is now 0.6.1

src/app/editor/example-contracts.js
```diff
@@ -1,96 +1,263 @@
...
-        return ballotToTest.winningProposal() == 1;
+        return ballotToTest.winningProposal() == 0;
     }
 }
 `
 
 module.exports = {
-  ballot: { name: 'ballot.sol', content: ballot },
-  ballot_test: { name: 'ballot_test.sol', content: ballotTest }
+  storage: { name: '1_Storage.sol', content: storage },
```
yann300: do we really need to prefix with "1_"

GrandSchtroumpf: I asked him to add it to relfect the level of difficulty. It helps beginner to know where to start

yann300: perhaps should we write directly in the code the level of difficulty.
that's not really obvious that "x_" means that

GrandSchtroumpf: Ok, I thought it was more obvious with a prefix. But a comment in the code is good for me too.

src/app/editor/example-contracts.js
```diff
@@ -1,96 +1,263 @@
...
+
+contract Storage {
+
+    uint256 number;
+
+    /**
+     * @dev Store value in variable
+     * @param num value to store
+     */
+
```
LianaHus: we can remove empty lines after comments

src/app/editor/example-contracts.js
```diff
@@ -1,96 +1,263 @@
...
+    
+    function store(uint256 num) public {
+        number = num;
+    }
+
+    /**
+     * @dev Return value 
+     * @return value of 'number'
+     */
+
```
LianaHus: same here

src/app/editor/example-contracts.js
```diff
@@ -1,96 +1,263 @@
...
+    }
+}`
+
+const owner = `pragma solidity >=0.4.22 <0.7.0;
+
+/**
+ * @title Owner
+ * @dev Set & change owner
+ */
+
```
LianaHus: here

LianaHus:
https://github.com/ethereum/remix-ide/issues/2515

PR #2510, opened 2019-12-30
Refactor executionContext and udapp, into common class

iurimatias:


yann300:
we should wait the release before merging this though
cc @LianaHus

PR #2514, opened 2020-01-06
Load plugins from parameters

yann300:
This PR:
 - refactor the way plugins are loaded at page load.
 - load from query param if provided, from local storage if not.


src/remixAppManager.js
```diff
@@ -232,3 +231,40 @@ export class RemixAppManager extends PluginEngine {
     ]
   }
 }
+
+class PluginLoader {
```
GrandSchtroumpf: Add comment to explain what this class is about.

PR #2517, opened 2020-01-07
moved Workshops to featured plugins section in home page

LianaHus:


yann300:
by the way, i just tried 10 min ago and 3box is actually loaded upon click on `workshops`

PR #2520, opened 2020-01-07
Tiny typo

yann300:
"plugin" is a bit redundant

### DIFF
--- a/src/app/ui/persmission-handler.js
+++ b/src/app/ui/persmission-handler.js
@@ -143,8 +143,8 @@ export class PermissionHandler {
       ? yo`<input type="checkbox" onchange="${switchMode}" checkbox class="custom-control-input" id="remember">`
       : yo`<input type="checkbox" onchange="${switchMode}" class="custom-control-input" id="remember">`
     const message = remember
-      ? `${fromName} has changed and would like to access the plugin ${toName}.`
-      : `${fromName} would like to access plugin ${toName}.`
+      ? `${fromName} has changed and would like to access ${toName}.`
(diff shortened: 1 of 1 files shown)


src/app/ui/persmission-handler.js
```diff
@@ -143,8 +143,8 @@ export class PermissionHandler {
       ? yo`<input type="checkbox" onchange="${switchMode}" checkbox class="custom-control-input" id="remember">`
       : yo`<input type="checkbox" onchange="${switchMode}" class="custom-control-input" id="remember">`
     const message = remember
-      ? `${fromName} has changed and would like to access the plugin ${toName}.`
-      : `${fromName} would like to access plugin ${toName}.`
+      ? `${fromName} has changed and would like to access ${toName}.`
+      : `${fromName} would like to access ${toName}.`
```
ryestew: With your change it reads,"FLATTENER WOULD LIKE TO ACCESS PLUGIN FILE MANAGER. :"  
So we should remove the "." or the ":" or both.   Because it is a headline - I think you should remove both. But either way ( just not both)

yann300: oh ok I will remove '.'

LianaHus: it is still a bit not smooth. maybe "Flattener" would like to access "File Manager" plugin?

yann300: I don't want to use the word plugin here. do you like to have double quote?

LianaHus: then at least double quotes, yes.

PR #2522, opened 2020-01-09
Export to Gist

yann300:
If "id" is not defined, it is not a gist update but a creation. we have to take the file from the browser explorer.
How could we add browser test for this? (cause we are testing against an external provider) @LianaHus @Aniket-Engg

### DIFF
--- a/src/app/files/file-explorer.js
+++ b/src/app/files/file-explorer.js
@@ -470,7 +470,8 @@ fileExplorer.prototype.toGist = function (id) {
     return data.files || []
   }
 
-  this.packageFiles(this.files, 'browser/gists/' + id, (error, packaged) => {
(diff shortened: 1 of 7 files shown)


src/app/files/file-explorer.js
```diff
@@ -470,7 +470,8 @@ fileExplorer.prototype.toGist = function (id) {
     return data.files || []
   }
 
-  this.packageFiles(this.files, 'browser/gists/' + id, (error, packaged) => {
+  const folder = id ? 'browser/gists/' + id : 'browser/'
```
LianaHus: would be nice to have the comment here

LianaHus: "If 'id' is not defined, it is not a gist update but a creation. we have to take the file from the browser explorer."

PR #2531, opened 2020-01-15
Receive/fallback functions in UI

LianaHus:


src/app/ui/multiParamManager.js
```diff
@@ -204,12 +204,15 @@ class MultiParamManager {
...
       contractProperty.classList.add(css.hasArgs)
-    } else if (this.funABI.type === 'fallback') {
-      contractProperty.classList.add(css.hasArgs)
-      this.contractActionsContainerSingle.querySelector('i').style.visibility = 'hidden'
     } else {
       this.contractActionsContainerSingle.querySelector('i').style.visibility = 'hidden'
-      this.basicInputField.style.visibility = 'hidden'
+      if (this.funABI.type === 'fallback' || this.funABI.type === 'receive') {
+        contractProperty.classList.add(css.hasArgs)
+        this.basicInputField.setAttribute('title', `'(${this.funABI.type}')`) // probably should pass name instead
```
LianaHus: this seems wrong I'll fix

src/app/ui/universal-dapp-ui.js
```diff
@@ -131,172 +115,160 @@ UniversalDAppUI.prototype.renderInstanceFromABI = function (contractABI, address
...
+    <label id="deployAndRunLLTxError" class="text-danger"></label>
+  `
+  // constract LLInteractions elements
+  const lowLevelInteracions = yo`
+    <div class="d-flex flex-column">
+      <div class="d-flex flex-row justify-content-between mt-2">
+        <label class="pt-2 border-top d-flex justify-content-start flex-grow-1">
+          Low level interactions with contract
+        </label>
+        <i aria-hidden="true"
```
yann300: can we have a link to doc instead of this info?

src/app/ui/universal-dapp-ui.js
```diff
@@ -131,172 +115,160 @@ UniversalDAppUI.prototype.renderInstanceFromABI = function (contractABI, address
...
+      if (calldata.length > 3 && calldata.substr(0, 2) === '0x') {
+        if (!helper.isHexadecimal(calldata.substr(2, calldata.length))) {
+          setLLIError('the calldata should be a valid hexadecimal value.')
+        }
+      }
+      if (!fallback) {
+        setLLIError("'fallback' function is not defined")
+      }
+    }
+    if ((calldata || amount !== '0') && !error) self.runTransaction(false, args, null, calldata, null)
```
yann300: why `amount !== '0'`

can't you just do `return setLLIError(...)`  when check fails and `self.runTransaction(false, args, null, calldata, null)` at the end?

src/app/ui/universal-dapp-ui.js
LianaHus: ah, sorry about this. totally forgot.

src/app/ui/universal-dapp-ui.js
LianaHus: and send an 'empty' transaction?

yann300: what do you mean empty?
even if there is no data and no value, that's still a valid tx. @LianaHus

LianaHus: lets do this in the second pr

yann300: please include this now (that's more readable)

PR #2536, opened 2020-01-16
Debugger tabs header font consistency

Aniket-Engg:


src/app/tabs/debugger/debuggerUI/vmDebugger/MemoryPanel.js
```diff
@@ -15,7 +15,7 @@ MemoryPanel.prototype.update = function (calldata) {
 }
 
 MemoryPanel.prototype.render = function () {
-  return yo`<div id="memorypanel" class="text-monospace small">${this.basicPanel.render()}</div>`
```
yann300: is the panel well formatted when removing `text-monospace` ? I remenber using font `monospace` for this specific reason?

Aniket-Engg: debug_2

Tabs looks like this

PR #2551, opened 2020-01-22
added missing `reject`

yann300:


src/app/files/compiler-metadata.js
```diff
@@ -99,6 +99,7 @@ class CompilerMetadata extends Plugin {
       if (provider && path) {
         self.executionContext.detectNetwork((err, { id, name } = {}) => {
           if (err) {
+            reject(err)
```
Aniket-Engg: I think `reject` should be called after `console`

PR #2555, opened 2020-01-24
pass compiler configuration to remix-tests

Aniket-Engg:
related to https://github.com/ethereum/remix/pull/1410


src/app/tabs/test-tab.js
```diff
@@ -166,13 +173,19 @@ module.exports = class TestTab extends ViewPlugin {
   runTest (testFilePath, callback) {
     this.loading.hidden = false
     this.fileManager.getFile(testFilePath).then((content) => {
-      var runningTest = {}
+      let runningTest = {}
```
GrandSchtroumpf: you can use `const`

src/app/tabs/test-tab.js
```diff
@@ -153,8 +153,15 @@ module.exports = class TestTab extends ViewPlugin {
...
-      remixTests.runTestSources(runningTest, currentCompilerUrl, canUseWorker(this.compileTab.getCurrentVersion()), () => {}, () => {}, (error, result) => {
+      const {currentVersion, evmVersion, optimize} = this.compileTab.getCurrentCompilerConfig()
+      const currentCompilerUrl = this.baseurl + '/' + currentVersion
+      const compilerConfig = {
+        currentCompilerUrl,
+        evmVersion,
+        optimize,
+        usingWorker: canUseWorker(currentVersion)
+      }
+      remixTests.runTestSources(runningTest, compilerConfig, () => {}, () => {}, (error, result) => {
```
GrandSchtroumpf: this is exacly for this kind of case that I would recommend an object argument instead of a list of ordered arguments (see comment [here](https://github.com/ethereum/remix/pull/1410#discussion_r371176679)).

PR #2562, opened 2020-01-30
Receive bug and more tests

LianaHus:


src/app/ui/universal-dapp-ui.js
```diff
@@ -161,7 +161,7 @@ UniversalDAppUI.prototype.renderInstanceFromABI = function (contractABI, address
     const fallback = self.udapp.getFallbackInterface(contractABI)
     const receive = self.udapp.getReceiveInterface(contractABI)
     const args = {
-      funABI: fallback,
+      funABI: (fallback ? fallback : receive),
```
yann300: and if both are defined?

LianaHus: then it's fine. fallback is default one

src/app/ui/universal-dapp-ui.js
```diff
@@ -177,16 +177,22 @@ UniversalDAppUI.prototype.renderInstanceFromABI = function (contractABI, address
       }
     }
     if (calldata) {
-      if (calldata.length > 3 && calldata.substr(0, 2) === '0x') {
+      if (calldata.length < 2) setLLIError('the calldata should be a valid hexadecimal value with size of at least one byte.')
+      if (calldata.length < 4 && calldata.substr(0, 2) === '0x') {
```
yann300: this is more than that, the size should be even

LianaHus: :) thanks

src/app/ui/universal-dapp-ui.js
```diff
@@ -191,7 +191,17 @@ UniversalDAppUI.prototype.renderInstanceFromABI = function (contractABI, address
...
     }
-    if (!receive && !fallback) setLLIError("Both 'receive' and 'fallback' functions are not defined")
+
+    if (!receive && !fallback) setLLIError(`Both 'receive' and 'fallback' functions are not defined`)
+
+    // we have to put the right function ABI:
+    // if receive is defined and that there is no calldata => receive function is called
+    // if fallback is defined => fallback function is called
+    if (receive && !calldata) args.funABI = receive
+    if (fallback) args.funABI = fallback
```
LianaHus: this will rewrite your previous change

yann300: indeed , changed it

PR #2565, opened 2020-01-31
Receive/Fallback better textmsg + test

LianaHus:


src/app/ui/universal-dapp-ui.js
```diff
@@ -176,13 +180,13 @@ UniversalDAppUI.prototype.renderInstanceFromABI = function (contractABI, address
     let calldata = calldataInput.value
     if (calldata) {
       if (calldata.length < 2 || calldata.length < 4 && helper.is0XPrefixed(calldata)) {
-        return setLLIError('the calldata should be a valid hexadecimal value with size of at least one byte.')
+        return setLLIError('The calldata should be a valid hexadecimal value with size of at least one byte.')
```
yann300: `with size of at least one byte.` should be removed to

Aniket-Engg: This message is fine here according to `calldata.length` check

yann300: ok

PR #2574, opened 2020-02-04
updating docs with about Low level interactions with a contract

ryestew:


docs/udapp.md
```diff
@@ -24,12 +24,27 @@ You will see the functions in the contract.  The functions buttons can have diff
...
+
+With Solidity 0.6 comes the **receive() function**.  
+
+- If you are executing a plain Ether transfer to a contract, you need to have the receive() function in your contract.  If you contract has been deployed and you want to send it funds you would input the amount of Ether or Wei etc., and then input **NOTHING** in the calldata field of **Low level interactions** and click the Transact button.
+
+- If you are sending calldata to your contract with Ether, then you need to use the fallback() function and have it with the state mutability of **payable**.  
+
+- If you are not sending ether to the contract but **are** sending call data then you need to use the fallback() function.
+
+- If you break the rules when using the **Low level interactions** you will be slapped with a warning.
```
Aniket-Engg: I like this line

docs/udapp.md
```diff
@@ -24,12 +24,27 @@ You will see the functions in the contract.  The functions buttons can have diff
...
+
+Please see the [solidity docs](https://solidity.readthedocs.io/en/v0.6.2/contracts.html#receive-ether-function) for more specifics about using the **fallback** and **receive** functions.
```
Aniket-Engg: Its better to use same version link in all solidity links including above ones : https://solidity.readthedocs.io/en/latest/

yann300: @ryestew could you update that URL?

ryestew: Would it be OK to have a link to both the root of the solidity docs ( like the one you are suggesting and also to have a like to the specific issue here - which is about **receive() & fallback()** or is it better to just to have to link to the Solidity docs and have the reader find the appropriate section on their own?

ryestew: Also on line 27 (of udapp.md) there is a link to another specific part of the solidity docs - the modifiers section.  What is the advantage of not pointing people to a specific part of the solidity docs?

PR #2577, opened 2020-02-04
function call fixed

Aniket-Engg:


yann300:
I think we should update all the function to use the `function` in that case, because we'll have the same bug if we introduce `this` is these other function.
And perhaps renaming the property to `udappUI`

PR #2581, opened 2020-02-05
Enviroment variables for running tests locally

ioedeveloper:


.env
```diff
@@ -0,0 +1 @@
+gist_token = 81b0b2c67a98fcca4827e2c8899db913829c7e04
```
yann300: can we have something like
`gist_token = <token>`
then each will have to provide his own (that should be documented in the readme too).
Cause i can't commit that this token will always be working

test-browser/tests/gist.js
```diff
@@ -1,4 +1,6 @@
 'use strict'
+require('dotenv').config()
```
yann300: can this be added to `init.js` so we are sure `.env` is applied for all the tests, cause we might need env variable for other cases.

.env
ioedeveloper: Yes, okay. I'll make the change.

test-browser/tests/gist.js
ioedeveloper: Okay.

yann300:
Thanks for the update, can the last commit title be changed to something like "move environment declaration to init phase & remove token", in order to describe more what is the change about.

ioedeveloper:
> Thanks for the update, can the last commit title be changed to something like "move environment declaration to init phase & remove token", in order to describe more what is the change about.

I've renamed my commit title

PR #2591, opened 2020-02-08
Homepage Tests

ioedeveloper:


test-browser/tests/homepage.js
```diff
@@ -0,0 +1,79 @@
...
+    .waitForElementVisible('#icon-panel #fileExplorerIcons > div:nth-child(1)')
+    .waitForElementVisible('#icon-panel #settingsIcons > div:nth-child(1)')
+    .waitForElementVisible('#icon-panel #settingsIcons > div:nth-child(2)')
+  },
+
+  'Loads Side Panel': function (browser) {
+    browser.waitForElementVisible('#side-panel')
+    .assert.containsText('#side-panel h6[class^="swapitTitle"]', 'FILE EXPLORERS')
+    .waitForElementVisible('#side-panel div[class^="treeview"]')
+    .waitForElementVisible('#side-panel ul[key="browser"] > li:nth-child(4)')
```
yann300: we had some discussions a while ago about how we should target elements from the e2e tests. The outcome was that we should target with `id` as much as possible so we don't rely so much in the html structure of the page. Not that you should do this now, but i'd like to open this discussion again
What do you think?
cc @LianaHus

LianaHus: Yes, using id is the right way to target element as long as id is unique, this is why we prepend a 'location of the element' to the element id. for example "remixHomePageOpenFile'". We still change the UI a lot, and if we use the element type we should change the test each time we change an element type, which will never be the case with id.

yann300: the drawback of this is that with all the tests we'll have, if you have to put an unique `id` each time for all the tested elements, the html might be spammed.
@ioedeveloper are there some best practices we can refer too?

ioedeveloper: Making your browser tests less fragile and more resilient is considered a best practice, and this can be achieved by using a unique identifier (not necessarily an ID due to spamming as stated by @yann300 ). A `data-attribute` can be used as a better alternative. Here are a few articles to support my claim [selecting elements best practices](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements), [making-your-ui-tests-resilient-to-change](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change) . I'm not so sure how well nightwatch works with data-attributes but it is something i'm will to look into.

yann300: good let's do that then

PR #2592, opened 2020-02-08
Add global web3 provider plugin

yann300:
this PR add a web3 provider as internal plugin.
merge after https://github.com/ethereum/remix-ide/pull/2589

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -48,6 +48,7 @@ import { FramingService } from './framingService'
 import { MainView } from './app/panels/main-view'
 import { ThemeModule } from './app/tabs/theme-module'
 import { NetworkModule } from './app/tabs/network-module'
+import { Web3ProviderModule } from './app/tabs/web3-provider'
 import { SidePanel } from './app/components/side-panel'
 import { HiddenPanel } from './app/components/hidden-panel'
 import { VerticalIcons } from './app/components/vertical-icons'
@@ -267,6 +268,8 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
 
   // ----------------- network service (resolve network id / name) -----
   const networkModule = new NetworkModule(blockchain)
+  // ----------------- represent the current selected web3 provider ----
+  const web3Provider = new Web3ProviderModule(blockchain)
   // ----------------- convert offset to line/column service -----------
   const offsetToLineColumnConverter = new OffsetToLineColumnConverter()
   registry.put({api: offsetToLineColumnConverter, name: 'offsettolinecolumnconverter'})
@@ -300,7 +303,8 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
     networkModule,
     offsetToLineColumnConverter,
     contextualListener,
-    terminal
+    terminal,
+    web3Provider
   ])
 
   // LAYOUT & SYSTEM VIEWS
@@ -383,7 +387,7 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
     console.log('couldn\'t register iframe plugins', e.message)
   }
 
-  await appManager.activatePlugin(['contentImport', 'theme', 'editor', 'fileManager', 'compilerMetadata', 'compilerArtefacts', 'network', 'offsetToLineColumnConverter'])
+  await appManager.activatePlugin(['contentImport', 'theme', 'editor', 'fileManager', 'compilerMetadata', 'compilerArtefacts', 'network', 'web3Provider', 'offsetToLineColumnConverter'])
   await appManager.activatePlugin(['mainPanel', 'menuicons'])
   await appManager.activatePlugin(['home', 'sidePanel', 'hiddenPanel', 'pluginManager', 'fileExplorers', 'settings', 'contextualListener', 'scriptRunner', 'terminal'])
 
--- a/src/remixAppManager.js
+++ b/src/remixAppManager.js
@@ -4,7 +4,7 @@ import { EventEmitter } from 'events'
 import QueryParams from './lib/query-params'
 
 const requiredModules = [ // services + layout views + system views
-  'manager', 'compilerArtefacts', 'compilerMetadata', 'contextualListener', 'editor', 'offsetToLineColumnConverter', 'network', 'theme', 'fileManager', 'contentImport', 'scriptRunner',
+  'manager', 'compilerArtefacts', 'compilerMetadata', 'contextualListener', 'editor', 'offsetToLineColumnConverter', 'network', 'theme', 'fileManager', 'contentImport', 'web3Provider', 'scriptRunner',
   'mainPanel', 'hiddenPanel', 'sidePanel', 'menuicons', 'fileExplorers',
   'terminal', 'settings', 'pluginManager']
 
--- a/src/app/tabs/web3-provider.js
+++ b/src/app/tabs/web3-provider.js
@@ -1,0 +1,31 @@
+import { Plugin } from '@remixproject/engine'
+import * as packageJson from '../../../package.json'
+
+export const profile = {
+  name: 'web3Provider',
+  displayName: 'Global Web3 Provider',
+  description: 'Represent the current web3 provider used by the app at global scope',
+  methods: ['sendAsync'],
+  version: packageJson.version,
+  kind: 'provider'
+}
+
+export class Web3ProviderModule extends Plugin {
+  constructor (blockchain) {
+    super(profile)
+    this.blockchain = blockchain
+  }
+
+  /*
+    that is used by plugins to call the current ethereum provider.
+    Should be taken carefully and probably not be release as it is now.
+  */
+  sendAsync (payload) {
+    return new Promise((resolve, reject) => {
+      this.blockchain.web3().currentProvider.sendAsync(payload, (error, message) => {
+        if (error) return reject(error)
+        resolve(message)
+      })
+    })
+  }
+}
--- a/src/app/ui/sendTxCallbacks.js
+++ b/src/app/ui/sendTxCallbacks.js
@@ -7,13 +7,13 @@ const typeConversion = remixLib.execution.typeConversion
 const Web3 = require('web3')
 
 module.exports = {
-  getCallBacksWithContext: (udappUI, executionContext) => {
+  getCallBacksWithContext: (udappUI, blockchain) => {
     let callbacks = {}
     callbacks.confirmationCb = confirmationCb
     callbacks.continueCb = continueCb
     callbacks.promptCb = promptCb
     callbacks.udappUI = udappUI
-    callbacks.executionContext = executionContext
+    callbacks.blockchain = blockchain
     return callbacks
   }
 }
@@ -67,7 +67,7 @@ const confirmationCb = function (network, tx, gasEstimation, continueTxExecution
       cb(txFeeText, priceStatus)
     },
     (cb) => {
-      self.executionContext.web3().eth.getGasPrice((error, gasPrice) => {
+      self.blockchain.web3().eth.getGasPrice((error, gasPrice) => {
         const warnMessage = ' Please fix this issue before sending any transaction. '
         if (error) {
           return cb('Unable to retrieve the current network gas price.' + warnMessage + error)
--- a/src/app/ui/universal-dapp-ui.js
+++ b/src/app/ui/universal-dapp-ui.js
@@ -233,7 +233,7 @@ UniversalDAppUI.prototype.runTransaction = function (lookupOnly, args, valArr, i
   const functionName = args.funABI.type === 'function' ? args.funABI.name : `(${args.funABI.type})`
   const logMsg = `${lookupOnly ? 'call' : 'transact'} to ${args.contractName}.${functionName}`
 
-  const callbacksInContext = txCallBacks.getCallBacksWithContext(this, this.executionContext)
+  const callbacksInContext = txCallBacks.getCallBacksWithContext(this, this.blockchain)
 
   const outputCb = (returnValue) => {
     if (outputOverride) {


src/app/panels/terminal.js
```diff
@@ -663,17 +668,27 @@ class Terminal extends Plugin {
...
+      var context = domTerminalFeatures(self, scopedCommands, self.blockchain)
+      try {
+        var cmds = vm.createContext(Object.assign(self._jsSandboxContext, context, self._jsSandboxRegistered))
+        var result = vm.runInContext(script, cmds)
+        self._jsSandboxContext = Object.assign(cmds, context)
+        return done(null, result)
+      } catch (error) {
+        return done(error.message)
+      }
+    }
```
GrandSchtroumpf: Consider using `else` to make it more readable.

src/app/panels/terminal.js
```diff
@@ -663,17 +668,27 @@ class Terminal extends Plugin {
...
-    var context = domTerminalFeatures(self, scopedCommands, self.blockchain)
+    if (script.indexOf('remix.') === 0) {
+      // we keep the old feature. This will basically only be called when the command is querying the "remix" object.
+      // for all the other case, we use the Code Executor plugin
+      var context = domTerminalFeatures(self, scopedCommands, self.blockchain)
+      try {
+        var cmds = vm.createContext(Object.assign(self._jsSandboxContext, context, self._jsSandboxRegistered))
+        var result = vm.runInContext(script, cmds)
+        self._jsSandboxContext = Object.assign(cmds, context)
+        return done(null, result)
```
GrandSchtroumpf: Avoid callback structure when using a Promise structure (async/await here) :
Then, you won't need to try catch, as the error will go in the `catch` of the promise.

```javascript
async _shell (script, scopedCommands) {
  ...
  if (script.split('.').shift() === 'remix') {
    ...
    return result
  } else {
    return this.call('remixCodeExecutor', 'execute', script)
  }
}
```

Then I can do : 
```javascript
this._shell(...)
  .then()
  .catch() // <- message error will be handle here.
```
Or 
```javascript
try {
  this._shell(...)
} catch (err) {
  // Handle error here or later.
}
```

src/app/panels/terminal.js
```diff
@@ -663,17 +668,27 @@ class Terminal extends Plugin {
     }
     return self.commands[name]
   }
-  _shell (script, scopedCommands, done) { // default shell
+  async _shell (script, scopedCommands, done) { // default shell
```
GrandSchtroumpf: Use JSDoc comment synthax, like that tools like VSCode can show it in their tooltips
```javascript
/** The default Shell */
async _shell (....)
```
You can even describe each param, which makes it easier to understand : 
```javascript
/**
 * The default Shell 
 * @params script { string } Script given by the user through the Terminal
 */
async _shell (....)
```

src/app/tabs/app-web3-provider.js
```diff
@@ -0,0 +1,31 @@
+import { Plugin } from '@remixproject/engine'
+import * as packageJson from '../../../package.json'
```
GrandSchtroumpf: Consider using the destructuration synthax to improve tree shaking when enabled : 
`import { version } from '...'`

src/app/tabs/app-web3-provider.js
```diff
@@ -0,0 +1,31 @@
...
+  constructor (blockchain) {
+    super(profile)
+    this.blockchain = blockchain
+  }
+
+  /*
+    that is used by plugins to call the current ethereum provider.
+    Should be taken carefully and probably not be release as it is now.
+  */
+  sendAsync (payload) {
```
GrandSchtroumpf: Use JSDoc synthax to describe the method. Also, `payload` doesn't provider any information about the content. Consider precising the type in the `@params` documentation.

src/app/tabs/app-web3-provider.js
```diff
@@ -0,0 +1,31 @@
+import { Plugin } from '@remixproject/engine'
+import * as packageJson from '../../../package.json'
+
+export const profile = {
+  name: 'appWeb3Provider',
```
GrandSchtroumpf: Name it `web3Provider` or `remixProvider` as it's going to be exposed to other plugins.
- Either we make it "generic" like that this work in any IDE. Then call it `web3Provider`. This is what I recommand.
- Or we force plugins developers to use different version depending on the IDE and we call it `remixProvider`.

src/codeExecutionWorker.js
```diff
@@ -0,0 +1,67 @@
+/* global eval */
+'use strict'
+import { connectIframe, PluginClient } from '@remixproject/plugin'
+import ethers from 'ethers' // eslint-disable-line
+import Web3 from 'web3' // eslint-disable-line
+import swarmgw_fn from 'swarmgw' // eslint-disable-line
+window.swarmgw = swarmgw_fn() // eslint-disable-line
+window.ethers = ethers
```
GrandSchtroumpf: I don't understand why you need to set value at the window level. Who/what can/should access this window object (it's in an hidden Iframe) ?

src/codeExecutionWorker.js
```diff
@@ -0,0 +1,67 @@
...
+    this.plugin = plugin
+  }
+  sendAsync (payload, callback) {
+    this.plugin.call('appWeb3Provider', 'sendAsync', payload).then(result => callback(null, result)).catch(e => callback(e.message))
+  }
+}
+
+class CodeExecutor extends PluginClient {
+  execute (script) {
+    if (!this.currentRequest.isFromNative) return
```
GrandSchtroumpf: `isFromNative` will be remove in next version. Consider your plugin as an open API that other plugins should be able to use. Else don't create a plugin.
Let the developer decide if s.he want the plugin to be called.

src/codeExecutionWorker.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+window.codeExec = new CodeExecutor()
+window.appWeb3Provider = new Web3Provider(window.codeExec)
+window.web3 = new Web3(window.appWeb3Provider)
+
+connectIframe(window.codeExec)
+console.log = function () {
+  window.codeExec.emit('log', {
```
GrandSchtroumpf: Personally I would have create an even per type of console event :
```typescript
const client = new CodeExecutor()

for (const logType of ['log', 'info', ... ]) {
  console[logType] = function () {
    client.emit(logType, Array.from(arguments))
  }
}
```

src/codeExecutionWorker.js
```diff
@@ -0,0 +1,67 @@
...
+          cmd: 'error',
+          data: [e.message]
+        })
+      }
+    }
+  }
+}
+
+window.codeExec = new CodeExecutor()
+window.appWeb3Provider = new Web3Provider(window.codeExec)
```
GrandSchtroumpf: Do you really need this ? I would consider interacting with the provider through web3 or ethers.

src/codeExecutionWorker.js
```diff
@@ -0,0 +1,67 @@
...
+'use strict'
+import { connectIframe, PluginClient } from '@remixproject/plugin'
+import ethers from 'ethers' // eslint-disable-line
+import Web3 from 'web3' // eslint-disable-line
+import swarmgw_fn from 'swarmgw' // eslint-disable-line
+window.swarmgw = swarmgw_fn() // eslint-disable-line
+window.ethers = ethers
+window.Web3 = Web3
+
+class Web3Provider {
```
GrandSchtroumpf: I'm not sure why you need a class for that. You're not using inheritance not instanciating it several times.
You can do : 
```javascript
const client = new CodeExecutor()
const provider = {
  sendAsync(payload, callback) {
    client.call('appWeb3Provider', 'sendAsync', payload)
      .then(result => callback(null, result)
      .catch(e => callback(e))
  }
}
```

Also I thought the Web3 provider exposed a Promise interface now instead of a callback one...

src/remixAppManager.js
```diff
@@ -79,7 +79,18 @@ export class RemixAppManager extends PluginEngine {
   async registeredPlugins () {
     const res = await fetch(this.pluginsDirectory)
     const plugins = await res.json()
-    return plugins.map(plugin => new IframePlugin(plugin))
+    const remixCodeExecutor = {
```
GrandSchtroumpf: Looks like opening Pandora box ^^. You cleaned up everything in the last PR to put it in IPFS, and now you're putting back a local profile of a plugin. 
If you want to create a plugin for real, then consider putting it in the remix-plugin-directory instead.

yann300:
> Looks to me that you don't really want to create a plugin :

Well yes and no, the need here is to run script in an another iframe. reusing the remix-plugin is a way to not rewrite everything.

src/app.js
```diff
@@ -44,6 +44,7 @@ import { FramingService } from './framingService'
 import { MainView } from './app/panels/main-view'
 import { ThemeModule } from './app/tabs/theme-module'
 import { NetworkModule } from './app/tabs/network-module'
+import { AppWeb3Provider } from './app/tabs/app-web3-provider'
```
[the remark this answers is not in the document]
yann300: I want to reflect that this provider is app scoped, like global provider

src/app/panels/terminal.js
```diff
@@ -663,17 +668,27 @@ class Terminal extends Plugin {
...
     if (script.indexOf('remix:') === 0) {
       return done(null, 'This type of command has been deprecated and is not functionning anymore. Please run remix.help() to list available commands.')
     }
     var self = this
```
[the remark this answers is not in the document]
yann300: yes but I want to keep the diff as minimal as possible.

src/app/panels/terminal.js
```diff
@@ -663,17 +668,27 @@ class Terminal extends Plugin {
...
-    var context = domTerminalFeatures(self, scopedCommands, self.blockchain)
+    if (script.indexOf('remix.') === 0) {
+      // we keep the old feature. This will basically only be called when the command is querying the "remix" object.
+      // for all the other case, we use the Code Executor plugin
+      var context = domTerminalFeatures(self, scopedCommands, self.blockchain)
```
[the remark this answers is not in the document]
yann300: same as above. I can plan for another PR for cleaning up the old code

src/app/tabs/app-web3-provider.js
yann300: see previous commit about the global aspect of this provider.
as we might have several providers in the future running at the same time, how to reflect that this provider is the global one? perhaps using `default` is more relevant

src/codeExecutionWorker.js
yann300: `isFromNative` is going to be replaced by something else?

yann300: cause I remember we have this for a specific reason

src/codeExecutionWorker.js
yann300: developer can then use instanciate ethers themselves

src/app/tabs/app-web3-provider.js
GrandSchtroumpf: By calling it web3Provider you say that it's the default one. If another plugin wants to expose there own version they'll have to name it differently. 
It's like solidity. If another plugin want to exposed their own version of Solidity Compiler (for any reason), they can, but they cannot call it "solidity".
If we use `app****` to say it's the default one, then we should do that for all of them...

src/codeExecutionWorker.js
GrandSchtroumpf: Which developers ? I don't understand who is supposed to have access to this "window" object.
Also, why web3 is defined by default and not ethers ? I think you want to expose both or none, don't you think ?

yann300: > why web3 is defined by default and not ethers 

that was the case before, but I can change that to expose both

yann300: > Which developers

Remix IDE users
use case is: you put some javascript code in a file in Remix and run `remix.execute()`.

GrandSchtroumpf: So you need to put the values in the window to be able to run this command ? So they don't have to do `const Web3 = await import('https://unpkg.com/web3')` ?

src/remixAppManager.js
yann300: ok good point

src/codeExecutionWorker.js
yann300: > So you need to put the values in the window to be able to run this command ?

when you run `remix.execute()`, the plugin just execute the code from the selected javascript file. that has nothing to do with the values in `window`.
we need anyway to inject the provider, because it is a plugin under the hood.
Then web3.js and ether are directly injected in the plugin scoped, so user can use them from the script.

yann300:
Another reason to make it a plugin is that this is exactly what Waffle needs, So it is very tempting to do so

src/app/tabs/app-web3-provider.js
yann300: ok so let's name it `provider`

src/codeExecutionWorker.js
GrandSchtroumpf: So if you don't need to access the values in the window object, why do you set them ?

yann300: > So you need to put the values in the window to be able to run this command 

you can run the command even if there's nothing in the window object.
having something in the window object is not mandatory for running the script.
If you don't have web3 in the window object then you can use it from your script.

GrandSchtroumpf:
I agree, that's why I think any plugin should be able to call it (with protection if needed).

yann300:
> with protection if needed

what kind of protection?

GrandSchtroumpf:
With the new engine you can do : 
```javascript
const name = this.currentRequest.from  // name of the caller
const canCall = await client.call('manager', 'canCall', { name }, this.profile, 'execute')
if (canCall) {
  // execute javascript here
} else {
  throw new Error(`Plugin ${name} cannot call method "execute" of plugin ${this.profile.name}`)
}
```
The `canCall` method is implemented by remix-ide to manage the permissions at a method level.

yann300:
`client.call('manager', 'canCall', { name }, this.profile, 'execute')`
where is the permission defined ?

GrandSchtroumpf:
On the IDE. Each IDE defines it's permission handler, and `canCall` will trigger it. 
On remix-ide for example there is a checkbox for "remember my choice", so we don't have to ask again. You can put the logic you want inside `canCall` (show permission UI, allow only "native", ...)

yann300:
in the specific case above, if the user defines in the IDE:
`plugin A` cannot call `plugin B`.`execute()`. Then the developer of the `plugin B` needs to do `
```
const canCall = await client.call('manager', 'canCall', { name }, this.profile, 'execute')
if (canCall) {
```
for the the permission to apply?

GrandSchtroumpf:
yes

yann300:
@GrandSchtroumpf could you review this again please?
the plugin is now no more in the remix ide repository but declared in remix-plugins-directory

PR #2596, opened 2020-02-12
WIP Using the new engine

LianaHus:


src/app/ui/txLogger.js
```diff
@@ -125,8 +125,8 @@ class TxLogger {
       }
       return false
     }
-    this.eventsDecoder = eventsDecoder
-    this.txListener = txListener
+    this.eventsDecoder = globalRegistry.get('teventsDecoder').api
```
yann300: look like a typo

src/app/components/plugin-manager-component.js
```diff
@@ -87,43 +87,40 @@ class PluginManagerComponent extends ViewPlugin {
...
     // Beta
-    if (api.profile.version && api.profile.version.match(/\b(\w*beta\w*)\b/g)) {
+    if (profile.version && profile.version.match(/\b(\w*beta\w*)\b/g)) {
       versionWarning = yo`<small title="Version Beta" class="${css.versionWarning} plugin-version">beta</small>`
     }
 
     const activationButton = isActive
       ? yo`
-      <button onclick="${_ => this.appManager.deactivateOne(name)}" class="btn btn-secondary btn-sm">
+      <button onclick="${_ => this.appManager.deactivatePlugin(profile.name)}" class="btn btn-secondary btn-sm">
```
GrandSchtroumpf: If you don't want to repeat `profile.name` & `profile.description` everywhere eyou can do : 
`const { name, description } = profile` at the beginning of the method.

src/app/components/plugin-manager-component.js
```diff
@@ -154,25 +151,26 @@ class PluginManagerComponent extends ViewPlugin {
...
       .filter(isFiltered)
       .filter(isNotRequired)
       .filter(isNotHome)
       .sort(sortByName)
-      .reduce(({actives, inactives}, api) => {
-        return this.appManager.isActive(api.name)
-          ? { actives: [...actives, api.name], inactives }
-          : { inactives: [...inactives, api.name], actives }
+      .reduce(({actives, inactives}, profile) => {
+        return isActive(profile.name) ? { actives: [...actives, profile], inactives }
```
GrandSchtroumpf: I think it's a little bit more readable like that.
```javascript
return isActive(profile.name)
  ? { actives: [...actives, profile], inactives }
  : { inactives: [...inactives, profile], actives }
```

src/app/components/plugin-manager-component.js
```diff
@@ -203,11 +201,11 @@ class PluginManagerComponent extends ViewPlugin {
         <section>
           ${activeTile}
           <div class="list-group list-group-flush plugins-list-group">
-            ${actives.map(name => this.renderItem(name))}
+            ${actives.map(profile => this.renderItem(profile, isActive(profile.name)))}
```
GrandSchtroumpf: If you pass `profile` as parameter of `renderItem`, you don't need to pass the `isActive` params : 
```typescript
renderItem(profile) {
  const pluginIsActive = isActive(profile.name)
}
```
Like that, here you can do:
`actives.map(profile => this.renderItem(profile)`

src/app/panels/tab-proxy.js
```diff
@@ -49,25 +49,24 @@ export class TabProxy {
       })
     })
 
-    appManager.event.on('activate', (name) => {
-      const { profile } = appManager.getOne(name)
+    appManager.event.on('activate', (profile) => {
```
GrandSchtroumpf: If you want you use deconstruction inside the callback parameter like that :
```javascript
appManager.event.on('activate', ({ name, location, displayName }) => { 

}
```

src/app/panels/terminal.js
```diff
@@ -45,7 +45,7 @@ class Terminal extends Plugin {
     self._api = api
     self._opts = opts
     self.data = {
-      lineLength: opts.lineLength || 80,
+      lineLength: opts.lineLength || 80, // ???
```
GrandSchtroumpf: Not sure why you added `????` 
This mean that he lineLength has he default value of 80, and can be override by the opts.

src/app/panels/terminal.js
```diff
@@ -437,10 +432,10 @@ class Terminal extends Plugin {
     self._shell('remix.help()', self.commands, () => {})
     self.commands.html(intro)
 
-    self._components.txLogger = new TxLogger(self._opts.eventsDecoder, self._opts.txListener, this, self.blockchain)
-    self._components.txLogger.event.register('debuggingRequested', (hash) => {
+    self._components.txLogger = new TxLogger(this, self.blockchain)
+    self._components.txLogger.event.register('debuggingRequested', async (hash) => {
       // TODO should probably be in the run module
-      if (!self._opts.appManager.isActive('debugger')) self._opts.appManager.activateOne('debugger')
+      if (!await self._opts.appManager.isActive('debugger')) await self._opts.appManager.activatePlugin('debugger')
```
GrandSchtroumpf: You can do : 
```typescript
await this.call('manager', 'activatePlugin', 'debugger')
```
Like that you don't need to pass the dependency on appManager.

In any case, `activatePlugin` always verify if the plugin is activated first. So you don't need the if statement.

src/app/tabs/test-tab.js
```diff
@@ -30,11 +30,11 @@ module.exports = class TestTab extends ViewPlugin {
     this.appManager = appManager
     this.renderer = renderer
     this.baseurl = 'https://solc-bin.ethereum.org/bin'
-    appManager.event.on('activate', (name) => {
-      if (name === 'solidity') this.updateRunAction(fileManager.currentFile())
+    appManager.event.on('activate', (profile) => {
```
GrandSchtroumpf: Same as above, if you only need `name` you can deconstruct the argument if you want:
```javascript
appManager.event.on('activate', ({ name }) => {
  if (name === 'solidity') this.updateRunAction(fileManager.currentFile())
})
```

src/remixAppManager.js
```diff
@@ -1,75 +1,77 @@
...
-    super(plugins, settings)
+    super()
     this.event = new EventEmitter()
-    this.registered = {}
     this.pluginsDirectory = 'https://raw.githubusercontent.com/ethereum/remix-plugins-directory/master/build/metadata.json'
     this.pluginLoader = new PluginLoader()
   }
 
-  onActivated (plugin) {
+  async canActivate (from, to) {
```
GrandSchtroumpf: Maybe add a comment to explain what it does and a todo to have more precise behavior here.

src/remixAppManager.js
```diff
@@ -1,75 +1,77 @@
...
+  async canDeactivate (from, to) {
+    return from.name === 'manager'
+  }
+
+  async canCall (From, to, method) {
+    // todo This is the dafault behaviour, we could save user choises in session scope
+    return true
+  }
+
+  onPluginActivated (plugin) {
```
GrandSchtroumpf: The argument is the profile, not the plugin itself.

src/remixAppManager.js
```diff
@@ -1,75 +1,77 @@
...
     this.pluginLoader.set(plugin, this.actives)
-    this.event.emit('activate', plugin.name)
+    this.event.emit('activate', plugin)
```
GrandSchtroumpf: Just `this.emit('activate', profiles)`. 
But the engine already emit the `'pluginActivated'` event. So you shouldn't have to worry with that.

src/remixAppManager.js
```diff
@@ -1,75 +1,77 @@
...
   getIds () {
-    return Object.keys(this.registered)
+    return Object.keys(this.profiles)
   }
 
-  onDeactivated (plugin) {
+  onPluginDeactivated (plugin) {
     this.pluginLoader.set(plugin, this.actives)
-    this.event.emit('deactivate', plugin.name)
+    this.event.emit('deactivate', plugin)
```
GrandSchtroumpf: Same as above for 
- `plugin` -> `profile`
- `this.event.emit` -> `this.emit`
- Event `'deactivatePlugin'` already exists

src/remixAppManager.js
```diff
@@ -1,75 +1,77 @@
...
   }
 
   onRegistration (plugin) {
-    if (!this.registered) this.registered = {}
-    this.registered[plugin.name] = plugin
     this.event.emit('added', plugin.name)
```
GrandSchtroumpf: This event is already emitted by the manager by default : `'profileAdded'`

src/remixAppManager.js
```diff
@@ -1,75 +1,77 @@
...
   }
 
-  // TODO check whether this can be removed
-  ensureActivated (apiName) {
-    if (!this.isActive(apiName)) this.activateOne(apiName)
+  async ensureActivated (apiName) {
+    await this.activatePlugin(apiName)
     this.event.emit('ensureActivated', apiName)
```
GrandSchtroumpf: I'm curious where we actually use this event and if we really need it.
Consider using `this.emit('ensureActivated')` so plugins can listen on it though the engine `this.on('manager', 'ensureActivated', (name) => ...)`

src/remixAppManager.js
```diff
@@ -1,75 +1,77 @@
...
   }
 
-  // TODO check whether this can be removed
-  ensureDeactivated (apiName) {
-    if (this.isActive(apiName)) this.deactivateOne(apiName)
+  async ensureDeactivated (apiName) {
+    await this.deactivatePlugin(apiName)
     this.event.emit('ensureDeactivated', apiName)
```
GrandSchtroumpf: Same as above

src/app/panels/terminal.js
LianaHus: @GrandSchtroumpf  because opts has no lineLength. How it should be used?

src/app/panels/terminal.js
```diff
@@ -95,12 +95,7 @@ class Terminal extends Plugin {
     self._jsSandboxContext = {}
     self._jsSandboxRegistered = {}
 
-    // TODO move this to the application start. Put it in mainView.
-    // We should have a HostPlugin which add the terminal.
-    opts.appManager.register(this)
-    opts.appManager.activate('terminal')
-
-    if (opts.shell) self._shell = opts.shell
+    if (opts.shell) self._shell = opts.shell // ???
```
LianaHus: @GrandSchtroumpf  same here opts.shell does not exist

src/app/panels/terminal.js
GrandSchtroumpf: ah ... I don't know XD. Maybe add a comment a little bit more descriptive than "???" then ^^

src/remixAppManager.js
LianaHus: I'll change this in next PR

Issue #2409, opened 2019-10-21
Permission Icons on Dark theme


GrandSchtroumpf:
Icons are black on the dark theme for the permission modal. It should be the same color as in the vertical icon bar : 

![image](https://user-images.githubusercontent.com/8143464/67211782-97d31c00-f41b-11e9-8701-85d122c1aa28.png)

LianaHus:
tab "x"-s as well

PR #2599, opened 2020-02-13
added filter invert to images

LianaHus:
fixing https://github.com/ethereum/remix-ide/issues/2409

### DIFF
--- a/src/app/tabs/compile-tab.js
+++ b/src/app/tabs/compile-tab.js
@@ -168,12 +168,8 @@ class CompileTab extends ViewPlugin {
       }
     })
 
-    globalRegistry.get('themeModule').api.events.on('themeChanged', (theme) => {
-      const invert = theme.quality === 'dark' ? 1 : 0
-      const img = document.getElementById('swarmLogo')
-      if (img) {
-        img.style.filter = `invert(${invert})`
-      }
+    globalRegistry.get('themeModule').api.events.on('themeChanged', () => {
+      globalRegistry.get('themeModule').api.fixInvert(document.getElementById('swarmLogo'))
     })
 
     // Run the compiler instead of trying to save the website
--- a/src/app/tabs/theme-module.js
+++ b/src/app/tabs/theme-module.js
@@ -24,7 +24,7 @@ const themes = [
 const profile = {
   name: 'theme',
   events: ['themeChanged'],
-  methods: ['switchTheme', 'getThemes', 'currentTheme', 'fixInvert'],
+  methods: ['switchTheme', 'getThemes', 'currentTheme'],
   version: packageJson.version,
   kind: 'theme'
 }
--- a/src/app/tabs/theme-module.js
+++ b/src/app/tabs/theme-module.js
@@ -24,7 +24,7 @@ const themes = [
 const profile = {
   name: 'theme',
   events: ['themeChanged'],
-  methods: ['switchTheme', 'getThemes', 'currentTheme'],
+  methods: ['switchTheme', 'getThemes', 'currentTheme', 'fixInvert'],
   version: packageJson.version,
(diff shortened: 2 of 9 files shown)


src/app/tabs/theme-module.js
```diff
@@ -26,7 +26,7 @@ const themes = [
 const profile = {
   name: 'theme',
   events: ['themeChanged'],
-  methods: ['switchTheme', 'getThemes', 'currentTheme'],
+  methods: ['switchTheme', 'getThemes', 'currentTheme', 'fixInvert'],
```
yann300: `fixInvert`  looks like a method internal to Remix and should not be called by other plugins.
btw `switchTheme` neither I think

LianaHus: I made it external in purpose. why not? Sokrates, for example, could use it for their Logo.

yann300:
could you remove the change in package-lock.json?

LianaHus:
> could you remove the change in package-lock.json?

I did

src/app/tabs/theme-module.js
yann300: yeah but that is more an helper that could be made available through the remix plugin than an API function.

LianaHus: what do you mean by "remix plugin" ?

yann300: I meant the remix-plugin library

GrandSchtroumpf: "theme" is not a part of the standard plugin api. It only apply for remix-ide. I don't have any hard feeling about having `switchTheme` as it doesn't look dangerous to me (maybe devs would like to build custom theme for remix in the future).
Concerning the `fixInvert` it wouldn't work for external plugins anyway, as it expects an Img HTMLElement, and large objects like that cannot be exchanges between Iframes.

Also I think it should be more a global css class than a javascript utils : 
```css
.dark-theme img.invert {
  filter: invert(1);
}

.light-theme img.invert {
  filter: invert(0);
}
```

LianaHus: some images are in color and some shouldn't change the invert.

LianaHus: and if I pass id it will work. but does it make sense at all? do you think it worth to have such a function? Is it obvious that `invert` can be used?

LianaHus: @GrandSchtroumpf

GrandSchtroumpf: I think it doesn't make so much sense to have an "invert" function. It would make more sense to provide a list of classes we are using in the application for developers to create their own theme.

LianaHus: ok then i'll remove it

src/app/tabs/compile-tab.js
```diff
@@ -168,12 +168,8 @@ class CompileTab extends ViewPlugin {
...
     })
 
-    globalRegistry.get('themeModule').api.events.on('themeChanged', (theme) => {
-      const invert = theme.quality === 'dark' ? 1 : 0
-      const img = document.getElementById('swarmLogo')
-      if (img) {
-        img.style.filter = `invert(${invert})`
-      }
+    globalRegistry.get('themeModule').api.events.on('themeChanged', () => {
+      globalRegistry.get('themeModule').api.fixInvert(document.getElementById('swarmLogo'))
```
yann300: doesn't need that for the ipfs logo?

LianaHus: I thought about it but it has color.. not sure. do you want me to add?

yann300: > do you want me to add?

not necessarily, just wanted to make sure you were aware of that

PR #2600, opened 2020-02-13
Default to dark theme

yann300:


src/app/tabs/theme-module.js
```diff
@@ -39,7 +39,7 @@ export class ThemeModule extends Plugin {
       config: registry.get('config').api
     }
     this.themes = themes.reduce((acc, theme) => ({ ...acc, [theme.name]: theme }), {})
-    this.active = this._deps.config.get('settings/theme') ? this._deps.config.get('settings/theme') : 'Flatly'
+    this.active = this._deps.config.get('settings/theme') ? this._deps.config.get('settings/theme') : 'Dark'
```
LianaHus: there are some issues still with the theme. Would be nice to merge this after update

yann300: Is this ok to merge now since an update has been pushed recently?
cc @ryestew

ryestew: I think this is fine to merge now.  We should probably put a link on the landing page to docs about changing themes.  Mostly because we are going to a default dark theme and that might surprise some people.

PR #2601, opened 2020-02-14
Plugin Manager Tests

ioedeveloper:


test-browser/tests/plugin.js
```diff
@@ -0,0 +1,93 @@
...
+    browser.waitForElementVisible('div[data-id="pluginManagerComponentPluginManager"]')
+    .click('button[data-id="pluginManagerSettingsButton"]')
+    .waitForElementVisible('form[data-id="pluginManagerSettingsPermissionForm"]')
+    .assert.containsText('form[data-id="pluginManagerSettingsPermissionForm"]', 'No Permission requested yet')
+    .modalFooterOKClick()
+    .click('div[data-id="verticalIconsFileExplorerIcons"]')
+    .switchFile('browser/3_Ballot.sol')
+    .click('div[plugin="ZoKrates"]')
+    .pause(5000)
+    .frame(0)
```
yann300: what is doing `frame` ?

ioedeveloper: `.frame(0)` allows switching into the first `<iframe>` indexed at 0 (plugin manager's iframe) and selecting DOM elements. I added this because i needed to select the compile button in ZoKrates which is inside an iframe.

test-browser/tests/pluginManager.js
```diff
@@ -0,0 +1,129 @@
...
+    .modalFooterOKClick()
+    .click('*[data-id="verticalIconsFileExplorerIcons"]')
+    .switchFile('browser/3_Ballot.sol')
+    .click('*[plugin="ZoKrates"]')
+    .pause(5000)
+    .frame(0)
+    .useXpath().click("//span[text()='Compile']")
+    .frameParent()
+    .useCss().waitForElementVisible('*[data-id="modalDialogContainer"]')
+    .assert.containsText('*[data-id="permissionHandlerMessage"]', 'ZOKRATES" WOULD LIKE TO ACCESS "FILE MANAGER" :')
```
yann300: we should also check is the api call is successful and returns the correct data, but that will be another PR...

Issue #2603, opened 2020-02-14
Remove i to right of Environment in Deploy & Run


ryestew:
Currently the i - icon to the right of the Environment pulldown menu in Deploy & Run goes to an EIP: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

We should remove the i - it is not clear what information it is revealing.

LianaHus:
I'm replacing it to https://remix-ide.readthedocs.io/en/latest/run.html#run-setup

PR #2604, opened 2020-02-14
fixed Environment doc link

LianaHus:
fixes https://github.com/ethereum/remix-ide/issues/2603


ryestew:
I think we had - debated whether to change this link or to eliminate the i icon all together.  I think your choice to link it to the docs here is the correct choice.  The link at the top of the deploy& run goes to the 1st page of the runtab docs.  This link goes to a section in the second page.  So this is more helpful and more direct.

PR #2605, opened 2020-02-15
Splash screen

yann300:
This PR add a splash screen in order to avoid all the glitches when the app starts.


src/app.js
```diff
@@ -98,12 +98,38 @@ var css = csjs`
     background-color   : var(--info);
     opacity            : 0.5;
   }
+  .centered {
+    position           : fixed;
+    top                : 50%;
```
ryestew: When I put:
top: 20%;
left: 30%;
The logo was a bit easier to see.  I'd recommend at least using the top:20%.  The left: 30% is less important.

PR #2612, opened 2020-02-19
update label for given permission

yann300:
- underline the plugin name
 - update the text


LianaHus:
could you please use <u> instead?

yann300:
> could you please use < u > instead?

done

PR #2623, opened 2020-02-24
some random style fixes in UI

LianaHus:


src/app/components/vertical-icons.js
```diff
@@ -66,6 +66,7 @@ export class VerticalIcons extends Plugin {
    */
   addIcon ({kind, name, icon, displayName, tooltip}) {
     let title = (tooltip || displayName || name)
+    title = title.replace(/^\w/, c => c.toUpperCase())
```
GrandSchtroumpf: If this is for style, better use the `text-tranform` css property : 
```css
target {
  text-transform: uppercase;
}
```

LianaHus: as we discussed we can't do it easily

PR #2625, opened 2020-02-24
[Draft] Filemanager plugin api

GrandSchtroumpf:
This PR implements thie File Manager API as it's defined [here](https://github.com/ethereum/remix-plugins-directory/blob/master/common-api/file-system/api.ts#L19). It only forward calls to existing code and add logic that doesn't exist. Updating the previous code should be done in another PR.

- [x] Add API
- [x] Add comments
- [x] Add error handlers
- [x] Forward to existing code

### DIFF
--- a/src/app/files/fileManager.js
+++ b/src/app/files/fileManager.js
@@ -42,6 +42,74 @@ class FileManager extends Plugin {
     this.init()
   }
 
+  /**
+   * Open the content of the file in the context (eg: Editor)
+   * @param {string} path path of the file
+   * @returns {void}
+   */
+  open (path) {
+
+  }
+  
+  /** 
+   * Set the content of a specific file
+   * @param {string} path path of the file
+   * @param {string} data content to write on the file
+   * @returns {void}
+   */
+  writeFile (path, data) {
+
+  }
+  
+  /**
+   * Return the content of a specific file
+   * @param {string} path path of the file
+   * @returns {string} content of the file
+   */
+  readFile(path) {
+
+  }
+
+  /** 
+   * Change the path of a file
+   * @param {string} oldPath current path of the file
+   * @param {string} newPath new path of the file
+   * @returns {void}
+   */
+  rename(oldPath, newPath) {
+
+  }
+
+  /** 
+   * Upsert a file with the content of the source file
+   * @param {string} src path of the source file
+   * @param {string} dest path of the destrination file
+   * @returns {void}
+   */
+  copyFile(src, dest) {
+
+  }
+
+  /**
+   * Create a directory
+   * @param {string} path path of the new directive
+   * @returns {void}
+   */
+  mkdir(path) {
+
+  }
+
+  /** 
+   * Get the list of files in the directory
+   * @param {string} path path of the directory
+   * @returns {string[]} list of the file/directory name in this directory
+   */
+  readdir(path) {
+    
+  }
+
+
+
   init () {
     this._deps = {
       config: this._components.registry.get('config').api,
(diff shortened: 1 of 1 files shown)


src/app/files/fileManager.js
```diff
@@ -23,13 +23,10 @@ const profile = {
   icon: 'data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0idXRmLTgiPz4KPHN2ZyB3aWR0aD0iMTc5MiIgaGVpZ2h0PSIxNzkyIiB2aWV3Qm94PSIwIDAgMTc5MiAxNzkyIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciPjxwYXRoIGQ9Ik0xNjk2IDM4NHE0MCAwIDY4IDI4dDI4IDY4djEyMTZxMCA0MC0yOCA2OHQtNjggMjhoLTk2MHEtNDAgMC02OC0yOHQtMjgtNjh2LTI4OGgtNTQ0cS00MCAwLTY4LTI4dC0yOC02OHYtNjcycTAtNDAgMjAtODh0NDgtNzZsNDA4LTQwOHEyOC0yOCA3Ni00OHQ4OC0yMGg0MTZxNDAgMCA2OCAyOHQyOCA2OHYzMjhxNjgtNDAgMTI4LTQwaDQxNnptLTU0NCAyMTNsLTI5OSAyOTloMjk5di0yOTl6bS02NDAtMzg0bC0yOTkgMjk5aDI5OXYtMjk5em0xOTYgNjQ3bDMxNi0zMTZ2LTQxNmgtMzg0djQxNnEwIDQwLTI4IDY4dC02OCAyOGgtNDE2djY0MGg1MTJ2LTI1NnEwLTQwIDIwLTg4dDQ4LTc2em05NTYgODA0di0xMTUyaC0zODR2NDE2cTAgNDAtMjggNjh0LTY4IDI4aC00MTZ2NjQwaDg5NnoiLz48L3N2Zz4=',
   permission: true,
   version: packageJson.version,
-  methods: ['getFolder', 'getCurrentFile', 'getFile', 'setFile', 'switchFile'],
+  methods: ['file', 'exists', 'open', 'writeFile', 'readFile', 'copyFile', 'unlink', 'rename', 'readdir', 'rmdir'],
```
yann300: we need to keep the old API for the moment.

yann300:
regarding keeping the old API, here are a few solutions:
 1 - we require all plugins devs to update their plugin
 2 - we require platform devs to keep implementing the old API. It won't be deprecated anymore.
 3 - we put a tiny module in remix-plugin for forwarding the old calls to the new standard.

2 or 3 seems to have less overhead

GrandSchtroumpf:
For me the best practice is : 
- Deprecate the old api for one major version.
- Redirect old apis calls to new one.
- Ask developers to witch to new API
- Remoe the old api once & for all after next major version.

yann300:
ok so
`Deprecate the old api for one major version.` looks like already done (see deprecated tag in the `remix-plugins-directory`) for the next to be released version.

Let's manage `Redirect old apis calls to new one.` do we agree that the code for this should be in remix-plugin?

GrandSchtroumpf:
I would have put this in remix-ide, but I can think of a generic method in remix-plugin we could use to redirect calls from one version to another of the standard API yes.

yann300:
this can be merged after https://github.com/ethereum/remix-plugin/issues/180

PR #2633, opened 2020-02-25
Deploy Code To IPFS

ioedeveloper:


src/app/tabs/runTab/contractDropdown.js
```diff
@@ -67,6 +82,11 @@ class ContractDropdownUI {
             ${this.atAddressButtonInput}
           </div>
         </div>
+        <div class="mt-2">
+          ${ipfsCheckbox}
+          <label for="deployAndRunPublishToIPFS" class="text-dark p-0 m-0">PUBLISH TO IPFS</label>
+          <i class="fas fa-info ml-2" aria-hidden="true" title="By publishing to IPFS is allowing to verify the contract code. The bytecode is already stored transparently in the blockchain. Publishing the source code and its ABI will greatly foster contract adoption (auditing, debugging, calling it, etc...)"></i>
```
LianaHus: I guess we should remove "By" in the Title

src/app/tabs/runTab/contractDropdown.js
```diff
@@ -67,6 +82,11 @@ class ContractDropdownUI {
             ${this.atAddressButtonInput}
           </div>
         </div>
+        <div class="mt-2">
+          ${ipfsCheckbox}
+          <label for="deployAndRunPublishToIPFS" class="text-dark p-0 m-0">PUBLISH TO IPFS</label>
+          <i class="fas fa-info ml-2" aria-hidden="true" title="Publishing to IPFS is allowing to verify the contract code. The bytecode is already stored transparently in the blockchain. Publishing the source code and its ABI will greatly foster contract adoption (auditing, debugging, calling it, etc...)"></i>
```
yann300: `The bytecode is already stored transparently in the blockchain`
that should be more explanatory:
`This will not compromise your code or making it unsafer (the bytecode is anyway already stored transparently in the blockchain). On the contrary, publishing the source code and its ABI will greatly foster adoption on your contract (auditing, debugging, calling it, etc...)`

LianaHus: this is too long for tooltip. if you want to add all these we should think of something else

LianaHus: @yann300

yann300: yeah right that is too much, ok let's keep the short version

ioedeveloper: What is the short version?

yann300: `Publishing to IPFS facilitate source code verification. Publishing the source code and its ABI will greatly foster contract adoption (auditing, debugging, calling it, etc...)`
is that ok @ioedeveloper @LianaHus

ioedeveloper: A little change to it IMHO.
`Publishing the source code and ABI to IPFS facilitates source code verification and greatly foster contract adoption (auditing, debugging, calling it, etc...)`
@yann300 @LianaHus

yann300: sounds good

ioedeveloper: Great! I will make the PR.

src/app/tabs/runTab/contractDropdown.js
```diff
@@ -67,6 +125,11 @@ class ContractDropdownUI {
             ${this.atAddressButtonInput}
           </div>
         </div>
+        <div class="mt-2">
```
yann300: the checkbox should be close to the deploy button

src/app/tabs/runTab/settings.js
```diff
@@ -212,6 +212,7 @@ class SettingsUI {
     this.event.trigger('clearInstance', [])
     this.updateNetwork()
     this.updatePlusButton()
+    this.blockchain.event.trigger('newExecutionContext', [])
```
yann300: `this.blockchain` should already trigger the event `contextChanged`, you don't need this one.

src/app/tabs/runTab/contractDropdown.js
```diff
@@ -43,16 +48,69 @@ class ContractDropdownUI {
...
+
+        const savedConfig = window.localStorage.getItem(`ipfs/${this.exEnvironment}/${this.networkName}`)
+
+        // check if an already selected option exist else use default workflow
+        if (savedConfig !== null) {
+          this.setCheckedState(savedConfig)
+        } else {
+          if (this.exEnvironment === 'vm') {
+            this.setCheckedState(false)
+          } else if (this.exEnvironment === 'web3') {
```
yann300: even if `web3` is selected, could be that the network your local node is syncing is `Main`.
`this.setCheckedState(this.networkName === 'Main')` should be fine

src/app/tabs/runTab/settings.js
ioedeveloper: Yes, `this.blockchain.event.on('contextChanged', () => {` should have been a better option for listening to an already existing event, but the EventManager which `this.blockchain` uses from remix-lib does not support the `.on('existingEvent')`.

ioedeveloper: It only has support for `register`, `unregister` and `trigger` event.

yann300: that's because of some inconsistency in the 2 event libs we use.....
could you use `register` ? `this.blockchain.event.register('contextChanged' , () => {})`

ioedeveloper: Will this not override the already existing callback for `'contextChanged'` event? I'll give it a try, i hope it does not break anything.

yann300: is this working  `this.blockchain.executionContext.event.register('contextChanged'`  ?
that's not ideal, but at least we don't add another event

ioedeveloper: Using `this.blockchain.event.register('contextChanged')` or `this.blockchain.executionContext.event.register('contextChanged')` fires the event, but the callback is not called because somehow `this.networkcallid > callid` in **blockchain.js**
```javascript
  updateNetwork (cb) {
    this.networkcallid++
    ((callid) => {
      this.executionContext.detectNetwork((err, { id, name } = {}) => {
        if (this.networkcallid > callid) return //<--- This line
        this.networkcallid++
        if (err) {
          return cb(err)
        }
        cb(null, {id, name})
      })
    })(this.networkcallid)
  }
```

PR #2634, opened 2020-02-26
Update import.md to use release tag when importing from GitHub

abcoathup:
Update GitHub import and console import instructions to use a release tag where available and change the example of importing from OpenZeppelin Contracts to use an official release rather than the master branch.  This is to avoid the community from using code outside of an official release.

I use the following example and note on the OpenZeppelin Community Forum:
https://forum.openzeppelin.com/t/deploy-a-simple-erc20-token-in-remix/1203


ryestew:
Thanks @abcoathup for this PR.  I appreciate you catching this issue and making the update!

PR #2635, opened 2020-02-27
General Settings Browser Tests

ioedeveloper:


yann300:
do you need these changes in `package-lock.json` ?

yann300:
can you rollback `package-lock.json`changes from the pr, looks like it is not needed

Issue #2528, opened 2020-01-14
Copy Github access token not working


ryestew:
On alpha - in settings in github token, clicking the copy icon - does not do anything.

PR #2652, opened 2020-03-12
Copy Github Access Token

ioedeveloper:
fixing #2528

### DIFF
--- a/src/app/ui/copy-to-clipboard.js
+++ b/src/app/ui/copy-to-clipboard.js
@@ -12,7 +12,7 @@ var css = csjs`
   }
 `
 
-module.exports = function copyToClipboard (getContent, tip = 'Copy value to clipboard', icon = 'fa-copy') {
+module.exports = function copyToClipboard (getContent, tip = 'Copied value to clipboard', icon = 'fa-copy') {
   var copyIcon = yo`<i title="${tip}" class="${css.copyIcon} far ${icon}" data-id="copyToClipboardCopyIcon" aria-hidden="true"></i>`
   copyIcon.onclick = (event) => {
     event.stopPropagation()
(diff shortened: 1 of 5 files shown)


src/app/ui/copy-to-clipboard.js
```diff
@@ -12,7 +12,7 @@ var css = csjs`
   }
 `
 
-module.exports = function copyToClipboard (getContent, tip = 'Copy value to clipboard', icon = 'fa-copy') {
+module.exports = function copyToClipboard (getContent, tip = 'Copied value to clipboard', icon = 'fa-copy') {
```
LianaHus: this is the title(tooltip) of the button. IMHO this should stah "Copy"

test-browser/tests/generalSettings.js
```diff
@@ -57,7 +57,7 @@ module.exports = {
     browser.waitForElementVisible('*[data-id="verticalIconsKindsettings"]')
     .click('*[data-id="copyToClipboardCopyIcon"]')
     .waitForElementVisible('*[data-shared="tooltipPopup"]:nth-last-of-type(1)')
-    .assert.containsText('*[data-shared="tooltipPopup"]:nth-last-of-type(1)', 'Copy value to clipboard')
+    .assert.containsText('*[data-shared="tooltipPopup"]:nth-last-of-type(1)', 'Copied value to clipboard')
```
LianaHus: also here

src/app/ui/copy-to-clipboard.js
ioedeveloper: Okay, then we should probably have a separate message for the success pop up.

src/app/ui/copy-to-clipboard.js
```diff
@@ -12,7 +12,7 @@ var css = csjs`
...
   var copyIcon = yo`<i title="${tip}" class="${css.copyIcon} far ${icon}" data-id="copyToClipboardCopyIcon" aria-hidden="true"></i>`
```
LianaHus: since you are on this PR could you please add p-2 to the class list of copyToClipboardCopyIcon here

PR #2655, opened 2020-03-15
Reject with a proper Error object

yann300:
If promises reject without using `Error` (e.g also without the `message` attribute), remix plugin is not able to identify this has an exception and thus to rethrow the error to the client ( see https://github.com/ethereum/remix-plugin/blob/v0.2.0/projects/engine/src/plugin/iframe.ts#L92 ).

In this PR, error can be stringified JSON... this is discutable...


yann300:
thanks to @bunsenstraat for helping finding the cause! ;)

yann300:
Will add some changes to this PR: having to stringify the result error is not really good and unit test should just return an error of type CompilationFailedError (which derive from Error).
I will create a PR in unit test module.

Issue #2630, opened 2020-02-25
Autoscroll in terminal to the bottom is not working


ryestew:

yann300:
that's either a bug or an enhancement

PR #2664, opened 2020-03-23
Fix Auto-scroll To Bottom In Terminal

ioedeveloper:
Fixed https://github.com/ethereum/remix-ide/issues/2630

### DIFF
--- a/src/app/panels/terminal.js
+++ b/src/app/panels/terminal.js
@@ -302,11 +302,24 @@ class Terminal extends Plugin {
     })
 
     function focusinput (event) {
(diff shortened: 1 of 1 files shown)


ioedeveloper:
Scroll to bottom only works for alphanumeric keys and not special keys like Ctrl, Alt, Arrow keys. These keys can be included if needed.

LianaHus:
@ioedeveloper  this is an idea for another PR maybe but would be nice to focus(not scroll) on input line after ctr+c to perform ctrl+v without clicking in the input line and scroll after ctrl+v

PR #2665, opened 2020-03-23
Update udapp.md: example nested struct in function

KevinSmall:
I recently had to pass a nested struct to a function in the remix ide, and found it difficult to find the information about the syntax to use. The docs would benefit from an example to show the correct way to type in a nested struct.


ryestew:
Thanks so much for this PR!  I adapted your example in another PR - https://github.com/ethereum/remix-ide/pull/2774.  This has just been merged and will be visible on Readthedocs in the very near future.  Thanks again and we really appreciate your contribution!

Issue #2663, opened 2020-03-23
plugin permission typo


yann300:
`allow` is written as the plugin name

```
solidity permissions:

Allow allow to call quorum

Allow allow to call mythx

Allow allow to call cfg

```

LianaHus:
but then users will loose their saved data. is it fine?

yann300:
they will loose current stored permission, yes I think that's ok

yann300:
they won't loose all the saved data. we need to only clear the stored permission

PR #2670, opened 2020-03-24
remove the old permissionss saved in localstorage

LianaHus:
since the structure has been changed

fixing https://github.com/ethereum/remix-ide/issues/2663

### DIFF
--- a/src/app/ui/persmission-handler.js
+++ b/src/app/ui/persmission-handler.js
@@ -41,6 +41,11 @@ export class PermissionHandler {
 
   constructor () {
     this.permissions = this._getFromLocal()
+    // remove the old permissions saved in versions < 0.9.4 since the structior has been changed
+    if (!localStorage.getItem('isNewEngine')) {
+      localStorage.setItem('plugins/permissions', '')
+      localStorage.setItem('isNewEngine', 'true')
+    }
   }
 
   _getFromLocal () {
--- a/src/app/ui/persmission-handler.js
+++ b/src/app/ui/persmission-handler.js
@@ -43,10 +43,10 @@ export class PermissionHandler {
     this.permissions = this._getFromLocal()
     this.currentVersion = 1
     // here we remove the old permissions saved before adding 'permissionVersion'
-    // since with v1 the structior has been changed because of new engine ^0.2.0-alpha.6 changes
+    // since with v1 the structure has been changed because of new engine ^0.2.0-alpha.6 changes
     if (!localStorage.getItem('permissionVersion')) {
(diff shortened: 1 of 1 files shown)


src/app/ui/persmission-handler.js
```diff
@@ -41,6 +41,11 @@ export class PermissionHandler {
 
   constructor () {
     this.permissions = this._getFromLocal()
+    // remove the old permissions saved in versions < 0.9.4 since the structior has been changed
+    if (!localStorage.getItem(['isNewEngine'])) {
```
yann300: why is this an array? `['isNewEngine']`

LianaHus: ah, a typo. will fix it

src/app/ui/persmission-handler.js
```diff
@@ -41,6 +41,11 @@ export class PermissionHandler {
 
   constructor () {
     this.permissions = this._getFromLocal()
+    // remove the old permissions saved in versions < 0.9.4 since the structior has been changed
+    if (!localStorage.getItem('isNewEngine')) {
```
yann300: perhaps having a more precise flag like `isNewEngine-<version_number>` would be better, the structure might change again in the future

LianaHus: well i had that before but then remove it. I can add it if you want, i didn't want to overthink

yann300: ok let's add it then

src/app/ui/persmission-handler.js
```diff
@@ -41,6 +41,13 @@ export class PermissionHandler {
 
   constructor () {
     this.permissions = this._getFromLocal()
+    this.currentVersion = 1
+    // here we remove the old permissions saved before adding 'permissionVersion'
+    // since with v1 the structior has been changed because of new engine ^0.2.0-alpha.6 changes
+    if (!localStorage.getItem('permissionVersion')) {
```
yann300: that still won't work like this, you have to do 
`if (localStorage.getItem('permissionVersion') !== this.currentVersion) {`

LianaHus: if this.currentVersion is 2. or 4. the user loads and there is no 'permissionVersion' this would be undefined. ok i can leave the check for undefined. 
now lets say permissionVersion is 1 and we have currentVersion=2 now so 1 != 2 we gonna rm the old one(correct) and set it to 1(incorrect).

with my logic, if the 'permissionVersion' was not introduced it will rm the old permissions and set 1 to permissionVersion. 
if it is already there will not do anything

PR #2676, opened 2020-03-25
Solidity Unit Testing Plugin E2E Tests

ioedeveloper:


Aniket-Engg:
Its good if we can test if various error in running tests are appeared on remix-ide UI side. We can try to test them for compilation error, deployment error and test run error at least.

ioedeveloper:
> We can try to test them for compilation error, deployment error and test run error at least.

These tests will come in a separate PR.

yann300:
> We can try to test them for compilation error, deployment error and test run error at least.

Is that still something to do/plan? @ioedeveloper @Aniket-Engg

ioedeveloper:
> Is that still something to do/plan? @ioedeveloper @Aniket-Engg

No, i think it is all good.

ioedeveloper:
> > Is that still something to do/plan? @ioedeveloper @Aniket-Engg
> 
> No, i think it is all good.

test-browser/tests/solidityUnittests.js
```diff
@@ -59,34 +62,62 @@ module.exports = {
...
+    .addFile('deployError_test.sol', sources[0]['browser/deployError_test.sol'])
+    .clickLaunchIcon('fileExplorers')
+    .switchFile('browser/deployError_test.sol')
+    .clickLaunchIcon('solidityUnitTesting')
+    .click('*[data-id="testTabCheckAllTests"]')
+    .click('.singleTestLabel:nth-of-type(5)')
+    .scrollAndClick('*[data-id="testTabRunTestsTabRunAction"]')
+    .pause(10000)
+    .waitForElementPresent('*[data-id="testTabSolidityUnitTestsSummary"]')
+    .assert.containsText('*[data-id="testTabSolidityUnitTestsSummary"]', 'contract deployment failed after trying twice')
```
Aniket-Engg, line 96: I am not sure how this is passing. `deployError_test ` contract will throw revert on execution. Text to check can be `Transaction has been reverted by the EVM`.

`contract deployment failed after trying twice` error is expected in a case where first time contract deployment is failed with OOG error and it also failed when we try it by doubling the gas in next execution.

test-browser/tests/solidityUnittests.js
```diff
@@ -33,3 +117,171 @@ function runTests (browser) {
...
+
+      import "remix_tests.sol"; // this import is automatically injected by Remix.
+      import "remix_accounts.sol";
+      import "./ks2a.sol";
+
+      contract kickstarterTest {
+          enum State { Started, Completed }
+
+          Kickstarter kickstarter;
+          /// #sender: account-0
```
Aniket-Engg, line 240: custom sender doesn't work with special method like `beforeEach`, `beforeAll` etc. for now.

test-browser/tests/solidityUnittests.js
```diff
@@ -33,3 +117,171 @@ function runTests (browser) {
...
+          function beforeAll () public {
+            kickstarter = new Kickstarter();
+            kickstarter.createProject("ProjectA", 123000);
+            kickstarter.createProject("ProjectB", 100);
+          }
+      
+          /// #sender: account-0
+          /// #value: 10000000
+          function checkProjectExists () public payable {
```
Aniket-Engg, line 249: Its good to check expected value of sender and value inside test.  Use:

```
Assert.equal(msg.sender, TestsAccounts.getAccount(0), "wrong sender");
Assert.equal(msg.value, 10000000, "wrong value");
```

Also better to use any other account except account-0 as that is a default one.

Further, its good if we can have a test failing because of `wrong sender` or `wrong value`

Aniket-Engg:
@ioedeveloper Please add one more test for a method with parameters inside test contract. Sample can be:
```
contract methodfailure {
    function add(uint a, uint b) public {
       uint c = a+b;
       Assert.equal(a+b, c, "wrong value");
    }
}
```
Expected error will be: `Method 'add' can not have parameters inside a test contract`

test-browser/tests/solidityUnittests.js
ioedeveloper: This should be an issue with the plugin for incorrect error thrown, because `contract deployment failed after trying twice` is the error received and i can only make assertions with the displayed output.
Screenshot from 2020-04-10 10-57-45

test-browser/tests/solidityUnittests.js
ioedeveloper: I'm having troubles testing multiple failing cases in a test because the test exits and skips to next test after encountering a single failing case.

ioedeveloper:
> @ioedeveloper Please add one more test for a method with parameters inside test contract. 
> Expected error will be: `Method 'add' can not have parameters inside a test contract`

Running this test does not display the desired output/error message. It just displays `Result: Running tests...` and never exits.
Screenshot from 2020-04-10 16-35-13

ioedeveloper:
`methodFailure` test now works with latest Remix Libs.

test-browser/tests/solidityUnittests.js
Aniket-Engg: ok, I see

test-browser/tests/solidityUnittests.js
Aniket-Engg: This is tracked further here: https://github.com/ethereum/remix-ide/issues/2745

test-browser/tests/solidityUnittests.js
Aniket-Engg: This will be further investigated here: https://github.com/ethereum/remix/issues/1453

test-browser/tests/solidityUnittests.js
ioedeveloper: I created separate tests in the contract to handle the failing cases.

Issue #2460, opened 2019-11-27
copy files to other instance doesn't work


LianaHus:

LianaHus:
@yann300 so should we remove this?

yann300:
yes i think we should

PR #2678, opened 2020-03-25
Removed Copy Files To Another RemixIDE Instance Icon/Button

ioedeveloper:
Fixed https://github.com/ethereum/remix-ide/issues/2460


LianaHus:
I think you should also remove from several other places

LianaHus:
file-Panel-Style.js
file-Panel.js
file-Explorer.js the function itself

LianaHus:
@ryestew we would need to remove this from docs as well
https://github.com/ethereum/remix-ide/blob/74ed4b6321e6d618170fbd32fa7cce20d96b92a6/docs/file_explorer.md

Issue #2672, opened 2020-03-24
Toaster - setFile


yann300:
For the moment, the toaster is displayed only if `canCall` is `false` which doesn't really make sense.
https://github.com/ethereum/remix-ide/blob/master/src/app/files/fileManager.js#L181

either
 - completely removing the toaster.
 - continue displaying the toaster even if the user check "remember"

PR #2679, opened 2020-03-25
transform toaster to inform message

LianaHus:
fixing https://github.com/ethereum/remix-ide/issues/2672


LianaHus:
currently, it will never ask but just inform the user on any set File. I would suggest leaving this only for not native plugins.

PR #2688, opened 2020-03-30
Web3 Provider modal for local devmode

ryestew:
Don't merge this PR yet.  There is an unresolved problem with an angle bracket

### DIFF
--- a/src/app/tabs/runTab/settings.js
+++ b/src/app/tabs/runTab/settings.js
@@ -193,15 +193,23 @@ class SettingsUI {
   }
 
   web3ProviderDialogBody () {
+    const thePath = '<path/to/local/folder/for/test/chain>'
+
     return yo`
       <div class="">
-        <span>Note: If you are using </span>
-        <a href="https://geth.ethereum.org/docs/rpc/server" target="blank">Geth</a>
-        <span>, please configure it to allow requests from Remix:</span>
+        Note: To use Geth & https://remix.ethereum.org, configure it to allow requests from Remix:(see <a href="https://geth.ethereum.org/docs/rpc/server" target="_blank">Geth Docs on rpc server</a>)
+        <div class="border p-1">geth --rpc --rpccorsdomain https://remix.ethereum.org</div>
         <br>
-        <b>geth --rpc --rpccorsdomain https://remix.ethereum.org</b>
-        <br><br>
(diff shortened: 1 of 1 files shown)


ryestew:
Can be merged - pending approval ( of course)

src/app/tabs/runTab/settings.js
```diff
@@ -193,15 +193,23 @@ class SettingsUI {
...
     return yo`
       <div class="">
-        <span>Note: If you are using </span>
-        <a href="https://geth.ethereum.org/docs/rpc/server" target="blank">Geth</a>
-        <span>, please configure it to allow requests from Remix:</span>
+        Note: To use Geth & https://remix.ethereum.org, configure it to allow requests from Remix:(see <a href="https://geth.ethereum.org/docs/rpc/server" target="_blank">Geth Docs on rpc server</a>)
+        <div class="border p-1">geth --rpc --rpccorsdomain https://remix.ethereum.org</div>
+        <br>
+        To run Remix locally & a local Geth test node, use this command: (see <a href="https://geth.ethereum.org/getting-started/dev-mode" target="_blank">Geth Docs on Dev mode</a>)
+        <div class="border p-1">geth --rpc --rpccorsdomain="package://a7df6d3c223593f3550b35e90d7b0b1f.mod" --rpcapi web3,eth,debug,personal,net --vmdebug --datadir ${thePath} --dev console</div>
```
yann300, line 204: instead of `package://a7df6d3c223593f3550b35e90d7b0b1f.mod` you should print `window.origin`

yann300: that will use the current origin of the page

ryestew: When running locally this will become `http://localhost:8080` instead of `package://a7df6d3c223593f3550b35e90d7b0b1f.mod`  - it does work successfully like that.  The problem with this approach is that when this is on remix-alpha, the command will have that URL in it - so you won't be running it locally.  If that is OK - then we should change the line above so it is not specifying running a local version of Remix.  Thus making it:
To run Remix & a local Geth test node ...

yann300: ah right, so we don't need to specify `locally` for remix

ryestew: &#x1f44d;

PR #2702, opened 2020-04-02
refactor provider and use web3 to send txs

iurimatias:


yann300:
@iurimatias closing this here cause it needs to be reopened in remix-project

Issue #2686, opened 2020-03-27
Incorrect data display for contract method 'call'


Aniket-Engg:
Steps to Reproduce:
* Compile & deploy `2_Owner.sol` with `Injected web3`
* call `getOwner` method

Issues:
1. On terminal, it shows call details as:
```
transaction hash | call0xaE55E6A96769C7E472f56358B74C16372021890A0x048672bDb4d8A4e001c667330272324ba3c31aA60x893d20e8
-- | --
from | 0xaE55E6A96769C7E472f56358B74C16372021890A
to | Owner.getOwner() 0x048672bDb4d8A4e001c667330272324ba3c31aA6
hash | call0xaE55E6A96769C7E472f56358B74C16372021890A0x048672bDb4d8A4e001c667330272324ba3c31aA60x893d20e8
input | 0x893...d20e8
decoded input | {}
decoded output | { 	"0": "address: 0xaE55E6A96769C7E472f56358B74C16372021890A" }
logs | []
```
Check `transaction hash` & `hash` details.

2. It shows `0x0000000000000000000000000000000000000000 `  as a result of `getOwner` call which should be an address same as shown in the `decoded output ` of call.

See this for above two issues:
<img width="1198" alt="calldisplay" src="https://user-images.githubusercontent.com/30843294/77753589-d23a8800-704f-11ea-936a-cb124888f0c7.png">

3. I call `getOwner` once again. Now if details are previous call is already expanded and I click on new call to expand details, it does nothing. On browser console, an error is displayed as:

`Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at txDetails (http://remix-alpha.ethereum.org/build/app.js:275834:8)
    at HTMLDivElement._8.onclick (http://remix-alpha.ethereum.org/build/app.js:275733:12)`

I tested it on remix-alpha with ropsten testnet.

Aniket-Engg:
Also on contract deployment, terminal shows tx details as:
<img width="237" alt="constructor" src="https://user-images.githubusercontent.com/30843294/77890992-ebc91300-728d-11ea-8f3f-ab8cd794e6c6.png">
 
Notice the text inside brackets.

yann300:
For the first issue `Check transaction hash & hash details.`
calls doesn't have hash, this is a made up hash, that I think is used for debugging calls when being connected to the `Javascript VM`

yann300:
[...]

PR #2708, opened 2020-04-06
Return result for non VM calls

LianaHus:
partially_fixed https://github.com/ethereum/remix-ide/issues/2686


Aniket-Engg:
I think this only solves number 2 of the issue. Not all issues are fixed. @LianaHus

LianaHus:
Have you tried? I think I can't reproduce others anymore

Aniket-Engg:
@LianaHus I checked it. I can still reproduce no. 3 of the issue.

Issue #2697, opened 2020-04-01
Improve the font of 'Open Files'


Aniket-Engg:
<img width="185" alt="openFile" src="https://user-images.githubusercontent.com/30843294/78138923-4d7fad80-7445-11ea-92e7-98a3fac97f07.png">

LianaHus:
the theme will be updated soon

Issue #2620, opened 2020-02-21
Low level functions UI


ryestew:
The Low level functions section should follow the same design rules as the rest of the Udapp - or at least it should basically follow the same design.  Here's a screenshot with some suggestions.

![low-level-interactions](https://user-images.githubusercontent.com/6050001/75071193-2accb180-54f5-11ea-93a6-6eb74749e8ff.png)

ryestew:
I'll attach a png ( below).
1. the btn is not the same height as the text box -it should be 32px high.
2. the btn could be btn-danger - because it is payable - but it is a debatable point.
3. I think it looks better when the text CALLDATA is 10px so it is noticeably smaller than Low Level Interactions above it.  But this involves adding a new class.
4. The icon "i" looks a bit better when its closer to the text "Low level interactions" but it also is legible where you have it.

![low-level-interactions](https://user-images.githubusercontent.com/6050001/78625284-abebc800-7859-11ea-9272-f883b21e3c25.png)

PR #2709, opened 2020-04-07
Low level ui

LianaHus:
fixes https://github.com/ethereum/remix-ide/issues/2620
fixes https://github.com/ethereum/remix-ide/issues/2697


ryestew:
I'll attach a png ( below).

    the btn is not the same height as the text box -it should be 32px high.
    the btn could be btn-danger - because it is payable - but it is a debatable point.
    I think it looks better when the text CALLDATA is 10px so it is noticeably smaller than Low Level Interactions above it. But this involves adding a new class.
    The icon "i" looks a bit better when its closer to the text "Low level interactions" but it also is legible where you have it.

low-level-interactions

ryestew:
low-level-interactions

ryestew:
low-level-interactions

ryestew:
The only essential point is #1 above.  The rest are to consider.

LianaHus:
Calldata is the same size as other labels, no? I'll check.
Others - good points! Will fix.
Thanks for fast review!

src/app/tabs/runTab/settings.js
```diff
@@ -98,7 +98,7 @@ class SettingsUI {
         </label>
         <div class="${css.account}">
           <select data-id="runTabSelectAccount" name="txorigin" class="form-control ${css.select} custom-select" id="txorigin"></select>
-          ${copyToClipboard(() => document.querySelector('#runTabView #txorigin').value)}
+          <div style="margin-left: 0px;">${copyToClipboard(() => document.querySelector('#runTabView #txorigin').value)}</div>
```
ryestew, line 101: Its probably a bad idea to put in the inline style.  But the css definition .account_3NyvTG i {
    margin-left: 12px;
}
is winning anyway.

ryestew:
Homepage changes:
Check the figma - the homepage is a bit too busy.   The font sizes in the buttons with the logos on the homepage ( like the one for Solidity) - is too big.  And the Gist button and the rest in that row are too tall.  I think also that the alert -jumbotron - shouldn't be a different color (unfortunately) - with Remi there - it doesn't need the background to highlight that part of the screen.  Resolver Engine should be on one line.

Deploy & Run
in the white theme the transact text is not centered in the button and the input box and the button have different color borders next to each other.  We can fix this later if you want to merge this now.  It looks fine in the dark theme.  Oh but there are changes in the setting.js file - I think to address the icons that are widely spaced.  But I don't think your changes fixed that.  The other issue on the Deploy module is that the up/down arrow in the account select box is not that visible.  This is another problem - probably should be addressed in another PR.


Terminal
additional padding looks good.

Solidity Compiler
I need to check later

I'm not done with the review.  I'll finish on my next gander at it.





.

LianaHus:
thanks for review!
text size can be fixed.
jumbatron background - you told me to leave it :)
resolver engine - should be fixed

D&R I'll take a look. if you have all those things documented somewhere please tag me.

src/app/tabs/runTab/settings.js
LianaHus: it's a hack, this is a wraper for copy to clipboard which we want to do only here and we have to do this since we 'generate' it.

Issue #2588, opened 2020-02-07
[Solidity Unit Testing] 'Run Tests' button improvement


Aniket-Engg:
- Button remains enabled even if no test file is selected, it can be disabled.
- Clicking on button, when no test file is selected does nothing, it can show a message like 'No test file selected'

PR #2715, opened 2020-04-09
Disabled 'Run Tests' Button When No Test Is Selected.

ioedeveloper:
Fixed https://github.com/ethereum/remix-ide/issues/2588

### DIFF
--- a/src/app/tabs/test-tab.js
+++ b/src/app/tabs/test-tab.js
@@ -87,9 +87,19 @@ module.exports = class TestTab extends ViewPlugin {
 
     if (eChecked) {
       checkAll.checked = true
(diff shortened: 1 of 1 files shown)


Aniket-Engg:
Button disabling is fine but currently, It shows `No test file selected` text in the `Results` section. So, if tests for a file are run and their result is getting shown in `Results` section, in that case, if someone clicks on `Uncheck all`, result will be wiped out and it will show the text `No test file selected`.

Not sure if it is very user friendly, may be we can display the text in some different way. What you think @ioedeveloper @LianaHus @ryestew ?

yann300:
@Aniket-Engg  is it good to merge this?

PR #2733, opened 2020-04-15
Move filemanager api to new api

ioedeveloper:
Originally https://github.com/ethereum/remix-ide/pull/2625


src/app/files/fileManager.js
```diff
@@ -23,7 +23,7 @@ const profile = {
   icon: 'data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0idXRmLTgiPz4KPHN2ZyB3aWR0aD0iMTc5MiIgaGVpZ2h0PSIxNzkyIiB2aWV3Qm94PSIwIDAgMTc5MiAxNzkyIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciPjxwYXRoIGQ9Ik0xNjk2IDM4NHE0MCAwIDY4IDI4dDI4IDY4djEyMTZxMCA0MC0yOCA2OHQtNjggMjhoLTk2MHEtNDAgMC02OC0yOHQtMjgtNjh2LTI4OGgtNTQ0cS00MCAwLTY4LTI4dC0yOC02OHYtNjcycTAtNDAgMjAtODh0NDgtNzZsNDA4LTQwOHEyOC0yOCA3Ni00OHQ4OC0yMGg0MTZxNDAgMCA2OCAyOHQyOCA2OHYzMjhxNjgtNDAgMTI4LTQwaDQxNnptLTU0NCAyMTNsLTI5OSAyOTloMjk5di0yOTl6bS02NDAtMzg0bC0yOTkgMjk5aDI5OXYtMjk5em0xOTYgNjQ3bDMxNi0zMTZ2LTQxNmgtMzg0djQxNnEwIDQwLTI4IDY4dC02OCAyOGgtNDE2djY0MGg1MTJ2LTI1NnEwLTQwIDIwLTg4dDQ4LTc2em05NTYgODA0di0xMTUyaC0zODR2NDE2cTAgNDAtMjggNjh0LTY4IDI4aC00MTZ2NjQwaDg5NnoiLz48L3N2Zz4=',
   permission: true,
   version: packageJson.version,
-  methods: ['getFolder', 'getCurrentFile', 'getFile', 'setFile', 'switchFile'],
+  methods: ['file', 'exists', 'open', 'writeFile', 'readFile', 'copyFile', 'unlink', 'rename', 'readdir', 'rmdir'],
```
LianaHus, line 26: readdir -> readDir
rmdir -> rmDir

GrandSchtroumpf: Actually the node api on filesystem use the lowercase here : 
https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_fs_readdir_path_options_callback

src/app/files/fileManager.js
```diff
@@ -42,6 +42,212 @@ class FileManager extends Plugin {
...
+    if (this.isFile(path)) {
+      throw new Error({ code: 'ENOTDIR', message })
+    }
+  }
+
+  /**
+   * Emits error based on error code
+   * @param {object} error error { code, message }
+   */
+  _handleError (error) {
```
LianaHus, line 82: what;s the reason of having this?

src/app/files/fileManager.js
```diff
@@ -42,6 +42,212 @@ class FileManager extends Plugin {
...
+
+  /**
+   * Verify if the path provided is a file
+   * @param {string} path path of the directory or file
+   * @returns {boolean} true if path is a file.
+   */
+  isFile (path) {
+    const extension = path.split('/').pop()
+
+    return extension && extension.indexOf('.') > -1
```
LianaHus, line 134: you can have browser/myfolder.is.cool folder

LianaHus: i think you should ask provider for this

src/app/files/fileManager.js
```diff
@@ -42,6 +42,212 @@ class FileManager extends Plugin {
...
+    return file
+  }
+
+  /**
+   * Verify if the path exists (directory or file)
+   * @param {string} path path of the directory or file
+   * @returns {boolean} true if the path exists
+   */
+  exists (path) {
+    const provider = this.fileProviderOf(path)
```
LianaHus, line 122: this will not work like that, it will return localHost or current provider without checking for path

src/app/files/fileManager.js
ioedeveloper: Okay

src/app/files/fileManager.js
ioedeveloper: So that errors have a uniform format and it is not just thrown everywhere in our codebase.

src/app/files/fileManager.js
ioedeveloper: Alright.

src/app/files/fileManager.js
ioedeveloper: We are using tailoring our api like node filesystem api.

src/app/files/fileManager.js
```diff
@@ -42,6 +42,212 @@ class FileManager extends Plugin {
...
+  /**
+   * Removes a directory recursively
+   * @param {string} path path of the directory to remove
+   * @note will not work on a file, use `unlink` instead
+   * @returns {void}
+   */
+  rmdir (path) {
+    this._handleExists(path, `Cannot remove directory ${path}`)
+    this._handleIsDir(path, `Cannot remove directory ${path}`)
+    // To implement
```
yann300, line 248: that will help https://github.com/ethereum/remix-ide/blob/master/src/app/files/fileProvider.js#L144

src/app/files/fileManager.js
```diff
@@ -42,6 +42,212 @@ class FileManager extends Plugin {
...
+  /**
+   * Create a directory
+   * @param {string} path path of the new directory
+   * @returns {void}
+   */
+  mkdir (path) {
+    if (this.exists(path)) {
+      this._handleError({ code: 'EEXIST', message: `Cannot create directory ${path}` })
+    }
+    // To implement
```
yann300, line 217: that should do the trick `window.remixFileSystem.mkdirSync`
or look at https://github.com/ethereum/remix-ide/blob/master/src/app/files/fileProvider.js#L92 for recursively creating folder.

src/app/files/fileManager.js
yann300, line 26: is `file` a replacement for `getCurrentFile`

yann300:
from what i see, there's no issue with parameters ordering right?

ioedeveloper:
@LianaHus left a comment about `_handle_error`, What do you think about it? @yann300 @GrandSchtroumpf

ioedeveloper:
> from what i see, there's no issue with parameters ordering right?

Yes, i think parameters ordering is okay.

src/app/files/fileManager.js
```diff
@@ -42,6 +42,212 @@ class FileManager extends Plugin {
...
+
+  /**
+   * Removes a file
+   * @param {string} path path of the file to remove
+   * @note will not work on a directory, use `rmdir` instead
+   * @returns {void}
+   */
+  unlink (path) {
+    this._handleExists(path, `Cannot remove file ${path}`)
+    this._handleIsDir(path, `Cannot remove file ${path}`)
```
yann300, line 192: and `window.remixFileSystem.unlinkSync(curPath, console.log)` should work for this case

yann300:
btw I've put some comment about using `window.remixFileSystem.*` . Not saying that we should put `window.remixFileSystem.*` call in the fileManager. These call should stay in the `fileProvider.js` . 
But that will probably require some changes in `fileProvider.js`

src/app/files/fileManager.js
ioedeveloper: Okay.

src/app/files/fileManager.js
yann300: sounds ok to me, what are the concern about that?

src/app/files/fileManager.js
LianaHus: shame

LianaHus: why they did half low case half uppercase???

src/app/files/fileManager.js
LianaHus: but it adds nothing more to the throw

GrandSchtroumpf: You're doing three time the job.
For example if the `mkdir` method throw, the message will be: 
"File already exists Cannot create directory ${path}" with the first part beeing handle by the code & the second by the error. But still you have a method `_handleExists`.

Either you have one method to handle all error with a mapping on the error code : 
```typescript
const errorMsg = {
  ENOENT: 'No such file or directory',
  ...
}
const createError = (err) => new Error(`${errorMsg[err]} ${err.message || ''}`)
...
mkdir (path) {
  if (this.exists(path)) {
    throw createError({ code: 'EEXIST', message: `Cannot create directory ${path}` })
  }
}
```
Or you have a named methods that take a string as argument : 
```typescript
_handleEnoent(msg) {
  throw new Error('No such file or directory ' + msg);
}
```
Or you can use assert methods by merging check & error : 
```typescript
_assertExist(path, msg) {
  if (....) {
    throw new Error('No such file or directory ' + msg);
  }
}
```

src/app/files/file-explorer.js
```diff
@@ -275,9 +280,17 @@ function fileExplorer (localRegistry, files, menuItems) {
...
           `Delete file`, `Are you sure you want to delete ${currentFilename} file?`,
-          () => {
-            files.remove(key)
-            self.updatePath('browser')
+          async () => {
+            const fileManager = self._deps.fileManager
+            const removeFile = await fileManager.remove(key)
+
+            if (!removeFile) {
+              tooltip(`failed to remove file ${key}.`)
```
LianaHus, line 288: I guess this should be uppercase

src/app/files/fileManager.js
```diff
@@ -42,6 +52,202 @@ class FileManager extends Plugin {
...
+   * @param {string} src path of the source file
+   * @param {string} dest path of the destrination file
+   * @returns {void}
+   */
+  async copyFile (src, dest) {
+    await this._handleExists(src, `Cannot copy from ${src}`)
+    await this._handleIsFile(src, `Cannot copy from ${src}`)
+    await this._handleIsFile(dest, `Cannot paste content into ${dest}`)
+    const content = await this.readFile(src)
+
```
LianaHus, line 189: how do you readFile or switchFile if it is not exist or is not a file?

package.json
```diff
@@ -62,7 +62,7 @@
     "remix-solidity": "0.3.30",
     "remix-tabs": "1.0.48",
     "remix-tests": "0.1.33",
-    "remixd": "0.1.8-alpha.10",
+    "remixd": "0.1.8-alpha.16",
```
LianaHus, line 65: I think lock file also needs to be checked in

src/app/panels/tab-proxy.js
```diff
@@ -34,7 +34,7 @@ export class TabProxy {
         return
       }
       this.addTab(file, '', () => {
-        this.fileManager.switchFile(file)
+        this.fileManager.open(file)
         this.event.emit('switchFile', file)
```
LianaHus, line 38: should we also change the name of event to openFile? 'switchFile' makes no sense anymore

src/app/files/fileManager.js
ioedeveloper: Should there be provision for that? Because if the file does not exist, there is no point for copy or switchToFile. On the other hand, `writeFile` api handles files that do not exist by creating a new one.

src/app/panels/tab-proxy.js
ioedeveloper: Okay, that can be done.

src/app/files/fileManager.js
```diff
@@ -23,9 +23,19 @@ const profile = {
...
+  ENOTDIR: 'Path is not on a directory',
+  EEXIST: 'File already exists',
+  EPERM: 'Permission denied'
+}
+const createError = (err) => {
+  return new Error(`${errorMsg[err.code]} ${err.message || ''}`)
+}
 
 // File System profile
 // - methods: ['getFolder', 'getCurrentFile', 'getFile', 'setFile', 'switchFile']
```
LianaHus, line 41: could you please remove this comment?

src/app/panels/tab-proxy.js
ioedeveloper: @yann300 Liana suggested we also rename `switchFile` command for nightwatch to `openFile` to avoid confusion. What do you think?

src/app/files/fileManager.js
ioedeveloper: Yeah sure

src/app/panels/tab-proxy.js
yann300: Only for nightwatch or for all the codebase?  (I mean for the event name too?)
If the changes are well scoped I think that makes sense. If there's a lot of internal rewording changes, I would suggest to split that for another PR.

ioedeveloper: It's not so much a change. It can go in this PR.

yann300:
@LianaHus @ioedeveloper pretty sure it's because the old api has been removed https://github.com/ethereum/remix-ide/pull/2733/files#diff-f28e4232b4dbf7d53c7b5543bdc9ad4fR26

ioedeveloper:
Should be merged after remix-plugin release https://github.com/ethereum/remix-plugin/pull/208 before merging.

Issue #2711, opened 2020-04-07
Improve UI of unit testing plugin


Aniket-Engg:
- [x] Add 'Stop' button to stop running tests
- [x] Add icons to 'Generate test file', 'Run Tests' & 'Stop' buttons
- [x] Remove parentheses from testcase title

LianaHus:
fix checkbox spaces

Aniket-Engg:
May be we can also show/hide all the explanatory text there as revisiting users may like it clean next time.

Aniket-Engg:
fixed by https://github.com/ethereum/remix-ide/pull/2746

PR #2746, opened 2020-04-16
redesigning layout of unit test plugin

LianaHus:
fixing https://github.com/ethereum/remix-ide/issues/2711


src/app/tabs/test-tab.js
```diff
@@ -239,12 +245,18 @@ module.exports = class TestTab extends ViewPlugin {
...
+      </div>
+    `
     const isSolidityActive = this.appManager.actives.includes('solidity')
     if (!currentFile || !isSolidityActive) {
       el.setAttribute('disabled', 'disabled')
-      if (!currentFile) el.setAttribute('title', 'No file selected')
-      if (!isSolidityActive) el.setAttribute('title', 'The solidity module should be activated')
+      if (!currentFile) {
+        el.setAttribute('title', 'No file selected')
+      } else if (!isSolidityActive) el.setAttribute('title', 'The solidity module should be activated')
```
Aniket-Engg, line 259: Better message can be: `'Solidity Compiler' plugin should be activated`

src/app/tabs/test-tab.js
```diff
@@ -255,50 +267,70 @@ module.exports = class TestTab extends ViewPlugin {
...
+          checked="true"
+        >
+        <label class="text-nowrap pl-2 mb-0" for="checkAllTests"> Select all </label>
+      </div>
+    `
+  }
+
+  infoButton () {
+    return yo`
+      <a class="btn border d-flex w-50 ml-2" title="Check out documentation." target="__blank" href="https://remix-ide.readthedocs.io/en/latest/unittesting.html">
```
Aniket-Engg, line 305: `How to use` button currently shows an underline on text. I think that should be removed. 

Also, better link to redirect can be https://remix-ide.readthedocs.io/en/latest/unittesting.html#generate-test-file as plugin is already activated.

src/app/tabs/test-tab.js
LianaHus: it is like that in my local changes. I'll push them

src/app/tabs/test-tab.js
```diff
@@ -239,12 +257,19 @@ module.exports = class TestTab extends ViewPlugin {
...
+      </button>
+    `
     const isSolidityActive = this.appManager.actives.includes('solidity')
     if (!currentFile || !isSolidityActive) {
       el.setAttribute('disabled', 'disabled')
-      if (!currentFile) el.setAttribute('title', 'No file selected')
-      if (!isSolidityActive) el.setAttribute('title', 'The solidity module should be activated')
+      if (!currentFile) {
+        el.setAttribute('title', 'No file selected')
+      } else if (!isSolidityActive) el.setAttribute('title', 'The "Solidity Plugin" should be activated')
```
Aniket-Engg, line 271: title can better updated to `'Solidity Compiler' plugin should be activated`

src/app/tabs/test-tab.js
```diff
@@ -255,50 +280,75 @@ module.exports = class TestTab extends ViewPlugin {
...
+          checked="true"
+        >
+        <label class="text-nowrap pl-2 mb-0" for="checkAllTests"> Select all </label>
+      </div>
+    `
+  }
+
+  infoButton () {
+    return yo`
+      <a class="btn border text-decoration-none pr-0 d-flex w-50 ml-2" title="Check out documentation." target="__blank" href="https://remix-ide.readthedocs.io/en/latest/unittesting.html">
```
Aniket-Engg, line 323: better link to redirect can be https://remix-ide.readthedocs.io/en/latest/unittesting.html#generate-test-file

LianaHus: will do

ryestew:
The icons on the buttons are not changing color when the theme goes to the light theme to the dark theme - but the text is changing.  A number of times I thought that the stop icon was an unchecked checkbox.  So maybe the icon on stop is not necessary.  With play it adds some focus.

Also I don't think the generate and the how to use are worth being buttons.  They could both be links in the paragraph above.
Also in the paragraph above - it could say that there needs to be a file compiled and a test file to have the run button enabled.

ryestew:
The link from the button How to use... doesn't provide much info - it is just the section about the generate a sample file.  So I think it would be better to link to the top of the docs - https://remix-ide.readthedocs.io/en/latest/unittesting.html
While it is a duplication of the link at the top of the page - it will get them to read about how to make unit tests.

Aniket-Engg:
@ryestew top of the docs starts with plugin activation. As `How to use...` will appear only after plugin is activated, so IMHO it is fine to avoid that plugin activation section and link after it.

PR #2755, opened 2020-04-20
Fetch and compile before debugging

### LINKED ISSUE #2712: Improve Compiler plugin API
currently the API has 2 functions:
 - `compile(name)`
 - `getCompilationResult()`

This is rather limited and we add more function like:

 - setConfiguration({compilerVersion, compilerName, evmVersion, optimization, ...})

This has probably to be guarded by an alert/warning because that will change behavior of the module.

### LINKED ISSUE #2732: Toaster upon setFile
Look like there's a `setTimeout` when an API call is made to the `setFile`, we don't need this anymore because the toaster is no longer proposing user actions.

### LINKED ISSUE #2742: ContractVerifier integration with the debugger
The contract verifier plugin allows to fetch metadata and source code from a given contract address.
We can use this in the debugger for fetching the contract, compile it and provide source highlighting.

yann300:
needs https://github.com/ethereum/remix/pull/1464

fix #2742


yann300:
For testing the code, 
- activate `debugger` and `run`
- switch to ropsten
- in the hash input, put `0x5db1b4212e4f83e36bf5bc306888df50f01a73708a71322bdc6f39a76a7ebdaa`
- click start debugging

This unfortunately will require this PR  from the remix repo https://github.com/ethereum/remix/pull/1463

src/app/tabs/debugger-tab.js
```diff
@@ -34,7 +34,38 @@ class DebuggerTab extends ViewPlugin {
       <div class="${css.debuggerTabView}" id="debugView">
         <div id="debugger" class="${css.debugger}"></div>
       </div>`
-    this.debuggerUI = new DebuggerUI(this.el.querySelector('#debugger'), this.blockchain)
+
+    this.on('fetchAndCompile', 'compiling', (settings) => {
+      toaster(yo`<div><b>Recompiling and debugging with params</b><pre>${JSON.stringify(settings, null, '\t')}</pre></div>`)
```
LianaHus, line 39: could you please add class="text-left" to this pre to align text to the left

src/app/compiler/compiler-sourceVerifier-fetchAndCompile.js
```diff
@@ -0,0 +1,125 @@
...
+        setTimeout(_ => this.emit('usingLocalCompilation', contractAddress), 0)
+        return compilation
+      }
+    }
+
+    let name = network.name.toLowerCase()
+    name === 'main' ? 'mainnet' : name // source-verifier api expect "mainnet" and not "main"
+    let data
+    try {
+      data = await this.call('source-verification', 'fetch', contractAddress, name.toLowerCase())
```
LianaHus, line 76: here is the format  {id: 3, name: "Ropsten"}

edisinovcic: @LianaHus you don't need to do toLowerCase it will be handled on the API level.

PR #2773, opened 2020-04-24
Accordion for static analysis modules

LianaHus:


ryestew:
If the accordion expansion and collapse could have the animation of https://getbootstrap.com/docs/4.3/components/collapse/#accordion-example - it would be immediately understandable what is happening.  And maybe even have the links be underlined links like in the bootstrap docs - as headings they look like headings not probably not clickable - and when I did click I was surprised by the lack of animation and had to do it a couple of times before I understood.

LianaHus:
I havn't use bootstrap example because I wanted to avoid jquery and complexity. I did much simpler version of it tricking radio buttons. I can try to add animation later.
about the headings. this is why I made an examples of UI with icons. btw two themes look different.

LianaHus:
added double arrow for closed title lines.. For animation, we should fix the height of the sections which I would prefer to void since those are not equal.

PR #2776, opened 2020-04-27
static analysis doc updated

### LINKED ISSUE #2744: Improve the documentation for solidity static analysis
https://remix-ide.readthedocs.io/en/latest/static_analysis.html

Aniket-Engg:
fixes #2744


docs/static_analysis.md
```diff
@@ -1,33 +1,296 @@
...
+Solidity Static Analysis
+========================
 
-This section gives information about the last compilation. By default, a
-new analysis is run at each compilation.
+Static code analysis is a process to debug the code by examining it and without actually executing the code. 
 
-The analysis tab gives detailed information about the contract code. It
-can help you avoid code mistakes and to enforce best practices.
+`Solidity Static Analysis` plugin performs static analysis on Solidity smart contracts once they are compiled. It helps to check security vulnerabilities and bad development practices. This plugin comes with `Solidity` environment of Remix IDE. It can also be activated individually from `Plugin Manager`.
```
ryestew, line 6: edit: - the sentence that begins with "It helps to check..." I think could be: "It checks for check security vulnerabilities and bad development practices, among other issues.

docs/static_analysis.md
```diff
@@ -1,33 +1,296 @@
...
+_Example:_
+```
+// using now for date comparison
+if(startDate > now)
+    isStarted = true;
+
+// using block.timestamp
+uint c = block.timestamp;
+```
+-   **Low level calls: Semantics maybe unclear**
```
ryestew, line 74: I'm not sure why this section title has "Semantics may be unclear" - the problems don't seem to be about Semantics.

docs/static_analysis.md
```diff
@@ -1,33 +1,296 @@
...
+
+Use of low level `call`, `callcode` or `delegatecall` should be avoided whenever possible. `send` does not throw an exception when not successful, make sure you deal with the failure case accordingly. Use `transfer` whenever failure of the ether transfer should rollback the whole transaction.
+
+_Example:_
+```
+x.call('something');
+x.send(1 wei);
+```
+-   **Blockhash usage: Semantics maybe unclear**
```
ryestew, line 83: Also I'm not sure this is a Semantics problems.

ryestew: Of course - changing the warning text would not happen in this PR - so this request and the one above - should probably go into an Issue and then the docs could be changed after the warning text is changed.

docs/static_analysis.md
```diff
@@ -1,33 +1,296 @@
...
+    uint[] users;
+    function resetState() public{
+        delete users;
+    }
+}
+```
+
+-   **For loop over dynamic array: Iterations depend on dynamic array's size**
+
+Loops that do not have a fixed number of iterations, for example, loops that depend on storage values, have to be used carefully: Due to the block gas limit, transactions can only consume a certain amount of gas. The number of iterations in a loop can grow beyond the block gas limit which can cause the complete contract to be stalled at a certain point. Additionally, using unbounded loops incurs in a lot of avoidable gas costs. Carefully test how many items at maximum you can pass to such functions to make it successful.
```
ryestew, line 148: I think these sentences should be edited:
The number of iterations in a loop can grow beyond the block gas limit which can cause the complete contract to be stalled at a certain point. Additionally, using unbounded loops incurs in a lot of avoidable gas costs.

The number of iterations in a loop can grow beyond the block gas limit which can stall the contract. 
(uhhh - does that have the same meaning? - my edit just above)

Additionally, using unbounded loops can incur avoidable gas costs.

docs/static_analysis.md
```diff
@@ -1,33 +1,296 @@
...
+
+_Example:_
+```
+function length(string memory a) public pure returns(uint) {
+    bytes memory x = bytes(a);
+
+    return x.length;
+}
+```
+-   **Delete from dynamic qrray: 'delete' on an array leaves a gap**
```
ryestew, line 265: typo in qrray ->array

docs/static_analysis.md
Aniket-Engg: done

docs/static_analysis.md
Aniket-Engg: This comes from libs. Lets discuss and improve that

docs/static_analysis.md
Aniket-Engg: done

docs/static_analysis.md
Aniket-Engg: done

ryestew: I'm sorry, I put a typo in my edit - this is the correct sentence:
Additionally, using unbounded loops incurs a lot of avoidable gas costs.

docs/static_analysis.md
Aniket-Engg: created https://github.com/ethereum/remix/issues/1470 to track this

Issue #2793, opened 2020-04-30
Allow highlighting multiple parts of the code


fvictorio:
See https://github.com/ethereum/remix-plugin/issues/165

tl;dr we are working on a solhint plugin for remix and when we call `client.editor.highlight` multiple times, only the last one is applied.

fvictorio:
I'm not familiar with `remix-ide` code, but it looks like each highlight is keyed with a `from` property:

https://github.com/ethereum/remix-ide/blob/9bf4f721f8f32b23475d6ababe90d46b92e37e4d/src/app/editor/SourceHighlighters.js#L13

and that `from` is the `currentRequest`:

https://github.com/ethereum/remix-ide/blob/9bf4f721f8f32b23475d6ababe90d46b92e37e4d/src/app/editor/editor.js#L196-L197

Maybe one solution would be to have an optional key to use there, and default to the current request?

yann300:
this is added to the next release and go to remix-alpha soon

yann300:
that's deployed in https://remix-alpha.ethereum.org
the previous highlight is now not discarded and `discardHighlightAt(line, filePath)` has been added
https://github.com/ethereum/remix-plugins-directory/blob/master/common-api/editor/api.ts#L12

fvictorio:
Awesome, thank you!

PR #2815, opened 2020-05-04
Allow multiple highlight per plugin

yann300:
fix https://github.com/ethereum/remix-ide/issues/2793


test-browser/commands/elementIsNotPresent.js
```diff
@@ -0,0 +1,15 @@
+const EventEmitter = require('events')
+
+class ElementIsPresent extends EventEmitter {
```
ioedeveloper, line 3: Nightwatch api methods contains commands that check for element present and element not present [Nightwatch API](https://nightwatchjs.org/js/app/api/method.html), so there may be no need to implement custom commands for those assertions. Except you have other reasons for wanting to use a custom command.

PR #2819, opened 2020-05-05
Update runtab to support new geth dev mode recommendations

ChrisChinchilla:
The recommended dev mode steps for geth are changing, and it made sense to get the run tab dialog in remix in line with those. Also changing from `localhost` to `127.0.0.1` seems to be more reliable for browser support.

### DIFF
--- a/src/app/tabs/runTab/settings.js
+++ b/src/app/tabs/runTab/settings.js
@@ -178,7 +178,7 @@ class SettingsUI {
     selectExEnv.addEventListener('change', (event) => {
       let context = selectExEnv.options[selectExEnv.selectedIndex].value
       this.blockchain.changeExecutionContext(context, () => {
-        modalDialogCustom.prompt('External node request', this.web3ProviderDialogBody(), 'http://localhost:8545', (target) => {
+        modalDialogCustom.prompt('External node request', this.web3ProviderDialogBody(), 'http://127.0.0.1:8545', (target) => {
           this.blockchain.setProviderFromEndpoint(target, context, (alertMsg) => {
             if (alertMsg) addTooltip(alertMsg)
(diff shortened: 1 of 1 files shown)


yann300:
ok thanks, it is no longer required to set the `rpcapi``?

fjl:
You still need the `--rpcapi`.

ryestew:
@fjl @ChrisChinchilla - there are 2 other flag differences between Chris' update and the current command suggestion in Remix's modal.  
1. The current command has `--dev console` & the suggested update has just `--dev` without the word **console**.
2. The suggested update does not include `--vmdebug`.  

Both versions worked for me.

So should we update the modal to the suggested command ( with the addition of the rpcapi flag) to look like this:
```
geth --datadir <path/to/local/folder/for/test/chain> --rpc --rpccorsdomain="http://localhost:8080" --rpcapi web3,eth,debug,personal,net  --dev
```
or should the command contain the  --dev console (probably not - as it looks like the flag now is just --dev) and --vmdebug (probably) like in the original:
```
geth --datadir <path/to/local/folder/for/test/chain> --rpc --rpccorsdomain="http://localhost:8080" --rpcapi web3,eth,debug,personal,net  --dev console --vmdebug
```

ChrisChinchilla:
@fjl @ryestew I guess it's a bit case by case. I understand the `rpcapi` flag as "use to enable APIs you need, but really we recommend you don't enable them", and the guide I was working on that related to this PR doesn't use the flag. But maybe Remix needs them enabled to be usable?

The most important change was the URl really, as it seems to specifically need to be `127.0.0.1` to work in all browsers.

yann300:
cc @ryestew  https://github.com/ethereum/remix-ide/pull/2848

PR #2833, opened 2020-05-12
Check if ast present

### LINKED ISSUE #2649: Yul debugging does not work with latest compiler
It is not possible to debug transactions compiled from yul code with the latest nightly solidity compiler. Note that we recently also added source maps for yul code.

yann300:
fix #2649
need https://github.com/ethereum/remix/pull/1485

### DIFF
--- a/src/lib/offsetToLineColumnConverter.js
+++ b/src/lib/offsetToLineColumnConverter.js
@@ -19,11 +19,17 @@ export class OffsetToLineColumnConverter extends Plugin {
 
   offsetToLineColumn (rawLocation, file, sources, asts) {
     if (!this.lineBreakPositionsByContent[file]) {
-      for (var filename in asts) {
-        var source = asts[filename]
-        if (source.id === file) {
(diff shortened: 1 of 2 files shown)


src/lib/offsetToLineColumnConverter.js
```diff
@@ -19,11 +19,17 @@ export class OffsetToLineColumnConverter extends Plugin {
...
-        if (source.id === file) {
-          this.lineBreakPositionsByContent[file] = this.sourceMappingDecoder.getLinebreakPositions(sources[filename].content)
-          break
+      const sourcesArray = Object.keys(sources)
+      if (!asts && file === 0 && sourcesArray.length === 1) {
+        // if we don't have ast, we process the only one available content
+        this.lineBreakPositionsByContent[file] = this.sourceMappingDecoder.getLinebreakPositions(sources[sourcesArray[0]].content)
+      } else {
+        for (var filename in asts) {
+          var source = asts[filename]
```
LianaHus, line 28: could you use const instead?

PR #2840, opened 2020-05-18
Fix and activate e2e test - source verification

yann300:


yann300:
can we merged that one?

yann300:
thanks ;)

Issue #2834, opened 2020-05-13
Source highlighter should keep one copy of highlight


LianaHus:
if we click a couple of times on the same warning from staticAnalysis for example, SourceHighlighter is adding new highlighting each time for the same segment of code.

PR #2858, opened 2020-05-27
use SourceHighlighters in StaticAnalisys and remove multiple highlight

LianaHus:
for the same position
partly_fix https://github.com/ethereum/remix-ide/issues/2834


yann300:
wasn't it possible to use the plugin API in the case of the debugger too?

PR #2859, opened 2020-05-27
new Ace editor dark theme

LianaHus:


LianaHus:
Svg is also fine.
I asked the same question to @yann300 today
I was planning to put it in assets. Not sure about external plugins.

PR #2878, opened 2020-06-03
Framing with parameters url

yann300:
add url parameters: 
 - minimizeterminal, 
 - minimizesidepanel, 
 - focus the last loaded plugin

This is part of https://github.com/ethereum/remix-ide/issues/2738


yann300:
This PR improve the way Remix can be embedded, by choosing which part should be or not minimized and by focusing to the last loaded plugin (if loaded from the URL parameter)

Issue #2795, opened 2020-04-30
Use Solidity as default (but only if data supports it)


marsrobertson:
If took me a trip to help the documentation manual.

I was missing some buttons.

Most of the software, most of the products have some sort of defaults.

You can skip configuration, use express settings and intuitively figure out the rest.

In my parallel universe Solidity has 90% (or more) market dominance, Vyper being an experimental thing.

If data supports it, please select Solidity as default.

LianaHus:
@yann300 how this should work with preloaded plugins from URL? should we load all together or only if we don't have the list in URL?

yann300:
The list in the URL as predominance.
yeah so, I think we could by default (and first time the user loads remix) activate the solidity plugin (and perhaps also the run&deploy). Then the list of plugins to activate is the list of previously loaded plugin (same as it is now). If there's something in the URL, then we'd take the list from there (same as it is now).

PR #2879, opened 2020-06-03
preactivate solidity

LianaHus:
fix https://github.com/ethereum/remix-ide/issues/2795


src/app.js
```diff
@@ -406,6 +406,10 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
   const framingService = new FramingService(sidePanel, menuicons, mainview, this._components.resizeFeature)
   framingService.start()
 
+  // preactivate solidity plugin
```
yann300, line 409: I think we should preactivate solidity and udapp, only if the `workspace` is empty, not always

LianaHus: your are right.  I'll do the change.

src/app.js
```diff
@@ -400,7 +400,13 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
   await appManager.activatePlugin(['home', 'sidePanel', 'hiddenPanel', 'pluginManager', 'fileExplorers', 'settings', 'contextualListener', 'scriptRunner', 'terminal', 'fetchAndCompile'])
 
   // Set workspace after initial activation
-  if (Array.isArray(workspace)) await appManager.activatePlugin(workspace)
+  if (Array.isArray(workspace)) {
```
yann300, line 403: you should check for the length here

PR #2891, opened 2020-06-08
Remixd Websocket Plugin

ioedeveloper:
Fix https://github.com/ethereum/remix-ide/issues/2811


src/app/files/file-explorer.js
```diff
@@ -240,15 +244,6 @@ function fileExplorer (localRegistry, files, menuItems) {
 
             if (!removeFolder) {
               tooltip(`failed to remove ${key}. Make sure the directory is empty before removing it.`)
-            } else {
-              const provider = fileManager.currentFileProvider()
-
-              if (provider) {
-                const { type } = provider
-
-                return self.updatePath(type)
```
yann300, line 249: is it not needed anymore? I think this was changed to : `self.updatePath(self.files.type)`

ioedeveloper: Yes, it is no longer needed. I moved it to https://github.com/ethereum/remix-ide/pull/2891/files#diff-4b3872e86385f2fb931b0375fe866ef5R151

src/app/files/remixDProvider.js
```diff
@@ -1,52 +1,45 @@
...
-              this.event.trigger('fileExternallyChanged', [path, content])
-            }
-          })
-        } else if (data.name === 'rootFolderChanged') {
-          // new path has been set, we should reset
-          this.event.trigger('folderAdded', [this.type + '/'])
-        }
-      }
+    this._appManager.on('remixd', 'folderAdded', (path) => {
+      this.event.trigger('folderAdded', [this.addPrefix(path)])
```
LianaHus, line 26: Don't you need folder removed?

ioedeveloper: file and folder removed share the same event `fileRemoved`

PR #2899, opened 2020-06-11
Optimizing docker build image

edisinovcic:
Docker image optimization.

```
remixproject/remix-ide                          latest              d3412ff0f646        4 minutes ago       46.2MB
remixproject/remix-ide                          old                 0a3ca03b3fd8        3 days ago          1.32GB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,6 @@
-FROM node:10
-# Create Remix user, don't use root!
-# RUN yes | adduser --disabled-password remix && mkdir /app
-# USER remix
+FROM nginx:alpine
+WORKDIR /
 
-# #Now do remix stuff
-# USER remix
-WORKDIR /home/remix
+COPY ./temp_publish_docker/ /usr/share/nginx/html/
 
-RUN git clone https://github.com/ethereum/remix-ide.git
-RUN git checkout origin remix_live
-
-WORKDIR /home/remix/remix
-RUN npm install
-RUN npm run build
-
-EXPOSE 8080 65520
-
-CMD ["npm", "run", "serve"]
+EXPOSE 80
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -10,11 +10,19 @@ WORKDIR /home/remix
 
 COPY ./ ./
 
-WORKDIR /home/remix/remix
-# npm ci would probably be better
-RUN npm install
+RUN npm ci
 RUN npm run build
 
-EXPOSE 8080 65520
+FROM nginx:alpine
+WORKDIR /
 
-CMD ["npm", "run", "serve"]
+COPY --from=0 /home/remix/build/ /usr/share/nginx/html/build/
+COPY --from=0 /home/remix/index.html /usr/share/nginx/html/index.html
+COPY --from=0 /home/remix/nginx.conf /etc/nginx/nginx.conf
+COPY --from=0 /home/remix/assets/ /usr/share/nginx/html/assets/
+COPY --from=0 /home/remix/icon.png /usr/share/nginx/html/icon.png
+COPY --from=0 /home/remix/background.js /usr/share/nginx/html/background.js
+COPY --from=0 /home/remix/soljson.js /usr/share/nginx/html/soljson.js
+COPY --from=0 /home/remix/package.json /usr/share/nginx/html/package.json
+
+EXPOSE 80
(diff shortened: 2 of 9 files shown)


docker-compose.yaml
```diff
@@ -16,3 +16,5 @@ services:
     build: 
       context: .
       dockerfile: Dockerfile.dev
+    ports: 
+      - 8080:80
```
yann300, line 20: maybe `8080` can be removed, is this not necessary to provide the port `65520` here?

Dockerfile.dev
```diff
@@ -10,11 +10,16 @@ WORKDIR /home/remix
 
 COPY ./ ./
 
-WORKDIR /home/remix/remix
-# npm ci would probably be better
-RUN npm install
+RUN npm ci
 RUN npm run build
```
yann300, line 14: do you need to `build` here?

Dockerfile.dev
```diff
@@ -10,11 +10,16 @@ WORKDIR /home/remix
...
 
-EXPOSE 8080 65520
+FROM nginx:alpine
+WORKDIR /
 
-CMD ["npm", "run", "serve"]
+COPY --from=0 /home/remix/build/ /usr/share/nginx/html/build/
```
yann300, line 19: this is all the files that need to be packaged in the docker https://github.com/ethereum/remix-ide/blob/master/.circleci/config.yml#L20

docker-compose.yaml
edisinovcic: This basically forwards Remix IDE from port 80 inside of the container to port 8080 to localhost, usually, port 80 is busy and it's not recommended to expose these ports.

Sure I can add 65520 for remixd.

Dockerfile.dev
edisinovcic: But also index.html?

Dockerfile
```diff
@@ -1,19 +1,13 @@
...
+COPY ./build/ /usr/share/nginx/html/build/
+COPY ./index.html /usr/share/nginx/html/index.html
+COPY ./nginx.conf /etc/nginx/nginx.conf
+COPY ./assets/ /usr/share/nginx/html/assets/
+COPY ./icon.png /usr/share/nginx/html/icon.png
+COPY ./background.js /usr/share/nginx/html/background.js
+COPY ./soljson.js /usr/share/nginx/html/soljson.js
+COPY ./package.json /usr/share/nginx/html/package.json
+
+EXPOSE 80
```
edisinovcic, line 13: @yann300 this is a variant of dockerfile which just copies all necessary files from host to the docker container without building anything.

Dockerfile.dev
edisinovcic: With the changes to Dockerfile (which is just a production variant) this can be used as a development environment (so you build then serve that content). Naming can be changed. Dockerfile -> Dockerfile.dev and Dockerfile.dev to Dockerfile if that's clearer :)

docker-compose.yaml
```diff
@@ -15,4 +15,7 @@ services:
     container_name: remixide-${TAG}
     build: 
       context: .
-      dockerfile: Dockerfile.dev
+      dockerfile: Dockerfile
+    # ports: 
+    #   - 80:80
```
edisinovcic, line 20: Commented this so it can be definined by user if they need only one or both

Dockerfile.dev
yann300: yes also index.html

ci/build_and_publish_docker_images.sh
```diff
@@ -1,9 +1,12 @@
 #!/bin/bash
 set -e
 
-# If not staging and master branch are existing
 export TAG="$CIRCLE_BRANCH"
 
+if [ "$CIRCLE_BRANCH" == "master" ]; then
+    export TAG="latest";
```
edisinovcic, line 7: @yann300 always master branch builds are tagged as latest (as that is default tag with docker) and we want users to always aim for latest stable build. 

Specific build releases can also be added and that is not covered here.

docker-compose.yaml
edisinovcic: Added both 8080 and 65520.

ci/build_and_publish_docker_images.sh
```diff
@@ -1,12 +1,16 @@
...
+fi
+
+rm -rf temp_publish_docker
```
edisinovcic, line 10: This is to ensure that we don't copy old files @yann300

PR #2907, opened 2020-06-15
Ci fix

ioedeveloper:
Bug: Circleci passes tests that fails in `before` of test suite.


ioedeveloper:
Nightwatch's exit code is 0 when a suite's `before` times out which makes circleci build to pass.
Circleci expects an exit code of 1 to indicate that tests was unsuccessful. This seems to be an issue with nightwatch and has probably been fixed in a recent version.
https://github.com/nightwatchjs/nightwatch/issues/1293 
https://stackoverflow.com/questions/44932865/nightwatchs-exit-code-is-0-when-a-suites-beforeeach-times-out

PR #2913, opened 2020-06-19
Fix Docker run instructions in Readme

boxhock:
Easy fix for a small problem

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,13 +57,13 @@ If you want to run latest changes that are merged into master branch then run:
 
 ```
 docker pull remixproject/remix-ide:latest
-docker run -p 8080:80 remixproject-remix-ide:latest
+docker run -p 8080:80 remixproject/remix-ide:latest
 ```
 
 If you want to run latest remix-live release run.
 ```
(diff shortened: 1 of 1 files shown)


edisinovcic:
@boxhock did you manage to get this running?

As your proposed format is not working for me

```
docker run -p 8080:80 remixproject-remix-ide:latest
Unable to find image 'remixproject-remix-ide:latest' locally
docker: Error response from daemon: pull access denied for remixproject-remix-ide, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
```

And usually format of images when pulling from docker hub is repository/image:tag. (For example solc: https://hub.docker.com/repository/docker/ethereum/solc/tags?page=1)

boxhock:
Hi @edisinovcic, I think you got that wrong.

The current readme has the format you just tried that fails.
My PR fixes this.

edisinovcic:
Ah sorry, good catch :)

yann300:
thanks @boxhock ;)

Issue #2911, opened 2020-06-18
File Manager external functiones should be e2e tested


LianaHus:
mkdir for example had a bug

PR #2916, opened 2020-06-24
Filemanager external api e2e

ioedeveloper:
Fix https://github.com/ethereum/remix-ide/issues/2911


src/app/files/fileManager.js
```diff
@@ -194,10 +194,24 @@ class FileManager extends Plugin {
...
    */
   async rename (oldPath, newPath) {
-    await this.__handleExists(oldPath, `Cannot rename ${oldPath}`)
+    await this._handleExists(oldPath, `Cannot rename ${oldPath}`)
     const isFile = await this.isFile(oldPath)
+    const newPathExists = await this.exists(newPath)
+    const provider = this.fileProviderOf(oldPath)
 
-    this.fileRenamedEvent(oldPath, newPath, !isFile)
+    if (isFile) {
```
yann300, line 202: looks like `fileRenameEvent` is not called and that the event located at https://github.com/ethereum/remix-ide/pull/2916/files#diff-f28e4232b4dbf7d53c7b5543bdc9ad4fR314 will never be trigerred

ioedeveloper: `fileRenamedEvent` is triggered after rename from provider. https://github.com/ethereum/remix-ide/blob/1b2bac5f1b84254731e17bebecc90ab2076e6f8a/src/app/files/fileManager.js#L270
So the changes are in order.

ioedeveloper: Adding a call to the `fileRenamedEvent` event in rename function triggers it twice. You can try it out locally.

yann300: does that mean it was actually a bug?

ioedeveloper: Yes, it was.

PR #2918, opened 2020-06-29
Use 0.2.4 remix-plugin & set default timeout

yann300:


src/app.js
```diff
@@ -239,6 +239,13 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
   const pluginLoader = appManager.pluginLoader
   const workspace = pluginLoader.get()
   const engine = new Engine(appManager)
+  engine.setPluginOption = (profile) => {
```
GrandSchtroumpf, line 242: I think it's ok to do it like that, but if you need to overide more stuff you might want to extends the Engine at some point.

src/app.js
```diff
@@ -239,6 +239,13 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
...
+    if (profile.kind === 'provider' ||
```
GrandSchtroumpf, line 243: It's a little bit weird to have this kind of indentation ^^.
If you have an automatic corrector you can do : 
```typescript
({ kind, name }) => {
  if (kind === 'provider' || name === 'LearnEth') {
    ...
 }
 ...
}
```

yann300: ok sure

Issue #2799, opened 2020-05-04
[debugger] Function stack trace


yann300:
This PR from remix libs add an API for retrieving the function stack trace from a given vm trace index: https://github.com/ethereum/remix/pull/1467
This issue is about putting this in the current UI

PR #2921, opened 2020-06-29
add panel to show function calls

iurimatias:
fix https://github.com/ethereum/remix-ide/issues/2799


yann300:
fyi `remix-ide-run-deploy ` fails because of the plugin name which has been reverted.

PR #3000, opened 2021-01-14
update to debugger doc

ryestew:
fixes https://github.com/ethereum/remix-project/issues/36
fixes fixes https://github.com/ethereum/remix-project/issues/606


ryestew:
do not merge yet

ryestew:
Please check the description of the stack.  Maybe the paragraph on line 74 is not necessary.  
Also I need a better explanation of the middle button in the debugger's nav.

yann300:
@ryestew you could perhaps just link to https://en.wikipedia.org/wiki/Stack_(abstract_data_type)

yann300:
> Also I need a better explanation of the middle button in the debugger's nav.

when you are in a call and click on that button it should directly move the slider to the end of that call

yann300:
but seems there's an issue with it

yann300 (approved):

PR #3009, opened 2021-03-02
Update remixd.md

LianaHus:


ryestew (commented):
You section about Warning for old users is better and more specific than what I just merged.  Of course I'm not sure they know that they are old users.  But line 11 has already been fixed.

docs/remixd.md
```diff
@@ -17,6 +17,14 @@ The code of `remixd` is
 Or just install it in the directory of your choice by removing the -g flag:
 `npm install remixd`
```
ryestew, line 18: should be:
npm install -g @remix-project/remixd

ryestew (changes requested):

docs/remixd.md
ryestew: and please check the live docs:
https://remix-ide.readthedocs.io/en/latest/remixd.html

PR #3011, opened 2021-04-06
rewrite of imports.md

ryestew:


docs/import.md
```diff
@@ -1,74 +1,101 @@
...
-For a detailed explanation of the `import` keyword see the
-[Solidity documentation](https://solidity.readthedocs.io/en/develop/layout-of-source-files.html?highlight=import#importing-other-source-files)
+When importing from NPM, or a URL (like github, a IPFS gateway, or a Swarm gateway) you do not need to do anything more than use the `import` statement in your contract. The dependencies do not need to be "preloaded" into the File Explorer's current Workspace before the contract is compiled.
 
-Here are a some of the main methods of importing a file:
+Files loaded from the import statement are placed in the **Files Explorer's** current Workspace's `.deps` folder. 
 
-Importing a file from the browser's local storage
--------------------------------------------------
+Under the hood, Remix checks to see if the files are already loaded in the **.deps** directory.  If not, it goes out and gets them from a url - if provided or via unpkg (if it is an NPM lib).
```
Aniket-Engg, line 14: `get them from the url`

Aniket-Engg: `via unpkg (if it is an NPM lib)` : Rather than putting `if it is an NPM lib`  in brackets, we should make it a statement for explanation to users

docs/import.md
```diff
@@ -1,74 +1,101 @@
...
-It is possible to import files directly from GitHub.  You should specify the release tag (where available), otherwise you will get the latest code in the master branch.  For OpenZeppelin Contracts you should only use code published in an official release, the example below imports from OpenZeppelin Contracts v2.5.0.
+### Import from Swarm
 
 ```
-pragma solidity >=0.4.22 <0.6.0;
+import 'bzz-raw://5766400e5d6d822f2029b827331b354c41e0b61f73440851dd0d06f603dd91e5';
+```
 
-import "https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v2.5.0/contracts/math/SafeMath.sol";
+### Importing from IPFS
```
Aniket-Engg, line 41: should be `Import from IPFS` for consistency

Aniket-Engg (approved):

docs/import.md
LianaHus: don't we need an example here?

PR #3044, opened 2021-09-04
Replace "step number" with "program counter"

paulrberg:
"Step number" is not as specific as "program counter". It leads to confusions:

https://ethereum.stackexchange.com/questions/109354/is-the-vm-trace-step-in-the-remix-debugger-the-program-counter


yann300:
I think "Step number" refers to the index in the vm trace, I need to double check that

paulrberg:
Yeah, it's [the index not he program counter](https://ethereum.stackexchange.com/a/109357/24693). Sorry!

PR #3051, opened 2021-11-14
Fixed bug in bash code to install remix-tests globally

robriks:
npm -g install @remix-project/remix-tests doesn't work; global option should go after install like so:
npm install -g @remix-project/remix-tests


Aniket-Engg (approved):

Aniket-Engg:
Thanks @robriks !!!

PR #3055, opened 2022-01-19
Removed duplicate text - 'in the browser'

raj-pranav:
Removed duplicate text 'in the browser' , it was being repeated two times continuously on line number 10


raj-pranav:
@yann300 : Can you please accept this pull request (It is a minor though)

yann300 (approved):

PR #3059, opened 2022-03-07
docs for js unit testing

Aniket-Engg:


docs/testing_using_Chai_&_Mocha.md
```diff
@@ -0,0 +1,161 @@
...
+## Test a contract
+
+Similarly unit tests can be written to test the functionality of a smart contract. An example to test default `1_Storage.sol` contract can be as:
+
+```
+const { expect } = require("chai");
+
+describe("Storage", function () {
+  it("test initial value", async function () {
+    const metadata = JSON.parse(await remix.call('fileManager', 'getFile', 'browser/contracts/artifacts/Storage.json'))
```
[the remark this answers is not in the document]
Aniket-Engg: fixed

docs/testing_using_Chai_&_Mocha.md
```diff
@@ -0,0 +1,161 @@
...
+
+![](images/run_with_mocha_option.png)
+
+Click on `Run with Mocha` option, tests will be executed and result will be shown on Terminal.
+
+![](images/run_with_mocha_result.png)
+
+## Test a contract
+
+Similarly unit tests can be written to test the functionality of a smart contract. An example to test default `1_Storage.sol` contract can be as:
```
ryestew, line 45: Maybe after the 1st sentence, add a sentence like:
The tests are run on the artifact file (the .json file) that is generated when the .sol file has been compiled.

docs/testing_using_Chai_&_Mocha.md
```diff
@@ -0,0 +1,161 @@
...
+});
+```
+
+Result will be as:
+
+![](images/run_with_mocha_storage_test.png)
+
+## Debugging a test transaction
+
+To debug a transaction in one of the tests can be done by printing the tranasation hash and using the [Remix Debugger plugin](https://remix-ide.readthedocs.io/en/latest/tutorial_debug.html).
```
ryestew, line 91: Some small edits:
To debug a transaction in one of the tests,  print the transaction hash and input that in the [Remix Debugger plugin](https://remix-ide.readthedocs.io/en/latest/tutorial_debug.html).

docs/testing_using_Chai_&_Mocha.md
```diff
@@ -0,0 +1,161 @@
...
+
+![](images/run_with_mocha_print_hash.png)
+
+## Hardhat Ethers Plugin
+
+Remix script runner also supports Hardhat plugin for integration with [ethers.js](https://github.com/ethers-io/ethers.js/) named as `hardhat-ethers`. Available methods under this plugin are:
```
ryestew, line 97: I am a little confused by this sentence.  Are you saying to the user that they need to 1st activate the hardhat-ethers plugin.  But this plugin does not have to be activated (right? ) so calling it a plugin is confusing if there is not activating involved - unless you say that it is always activated.  Also, I assume that no one would know what **Remix script runner** is.  So either, it needs to be explained or not called by its name and just by its description.

ryestew: Also this phrase: **supports Hardhat plugin** should be **supports the Hardhat plugin**
But which hardhat plugin?  The Hardhat Provider plugin?

ryestew (changes requested):

docs/testing_using_Chai_&_Mocha.md
Aniket-Engg: added something similar in the code itself

docs/testing_using_Chai_&_Mocha.md
Aniket-Engg: updated

docs/testing_using_Chai_&_Mocha.md
Aniket-Engg: updated

yann300 (approved):

ryestew (approved):

PR #3073, opened 2022-05-18
use https

yann300:


docs/remixd.md
```diff
@@ -84,9 +84,9 @@ application** that can access the `TCP port 65520` on your local host.
   https://remix.ethereum.org
   https://remix-alpha.ethereum.org
   https://remix-beta.ethereum.org
-  http://remix.ethereum.org
-  http://remix-alpha.ethereum.org
-  http://remix-beta.ethereum.org
+  https://remix.ethereum.org
```
ryestew, line 87: Do we want to leave these as http - if not then these need to be erased because they are duplicated

ryestew (changes requested):

ryestew (approved):

PR #3079, opened 2022-05-25
update to remixd with short command

ryestew:
fix https://github.com/ethereum/remix-project/issues/2413


Aniket-Engg (commented):
I think rather than presenting short or full command, we can say this is the command and these are the options.

yann300 (approved):

PR #3109, opened 2022-07-30
Updated wording to avoid confusion

robriks:
I usually see transaction identifiers referred to as a 'transaction hash' and not, as this file reads, a 'transaction address'

The current wording may confuse newcomers who think it refers to the function caller address or even their EOA address, so I propose changing it to 'hash' similar to the way it is referred to on block explorers


Aniket-Engg (approved):

Aniket-Engg:
Thanks @robriks !!

PR #3115, opened 2022-09-07
fix two errors

drafish:
1. When inputting parameters in the collapsed view, strings do not need to be wrapped in double quotes. I have tested.
2. a-recorder.png in run.md is wrong.

image


drafish:
Review required @Aniket-Engg @ryestew

ryestew (approved):

PR #3117, opened 2022-09-18
fix markdown parse issue

drafish:
image
I don't know if `recommonmark` was working properly before, but now there is a problem. All source strings parsed by recommonmark lose their markdown signs. 

For example, `**[Remixd](remixd.html)**` will be parsed into `Remixd`, emphasis and links are gone, and the same will be when it is converted into html, which means all translated docs will be effected.

English Version
image

Chinese Version
image

It even affect the english docs.
image

The string "npm version" should be the alt of this image. But now it became an element next to the image.

Since readthedocs officially recommends using `MyST` as an alternative to `recommonmark`, I tested it, and `MyST` does work as I expected.

Also, there is one more problem. The `docs/_build` folder was added to the `.gitignore` file. So when you commit `docs/_build/gettext`, the new pot files will not be committed, nor will they be in crowding.

I'm a little confuse about the `docs/_build/gettext` folder, because in `translation.md`, it says "The `_build/gettext` folder doesn't need to be pushed to master". But if crowding get the pot files from this repository, then you should commit this folder.

My suggestion is to ignore all files in the `docs/_build` folder except the pot files, which means the`docs/_build/gettext/.doctrees` folder need to be deleted, and pot files used to be ignored need to be commited, and also `.gitignore` and `translation.md` should be updated. Besides, don't forget to update the pot files before you commit it, because of the markdown parse issue.

I'd be happy to push another commit to get this done. But I'm not sure you will accept pot files update from outside the team. So I leave you guys to decide. @yann300 @ryestew @Aniket-Engg

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Remix libraries work as a core of native plugins of Remix IDE. Read more about l
 Steps to build this project as as:
 ```
 pip3 install sphinx sphinx_rtd_theme
-pip3 install recommonmark
+pip3 install myst-parser
 git clone https://github.com/ethereum/remix-ide.git
 cd docs/
 make html
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -30,14 +30,19 @@ import os
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['recommonmark'] # required for sphinx v3.0.0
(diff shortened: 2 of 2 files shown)


ryestew (approved):

ryestew:
Thanks @drafish for another great PR!!  I'll check into the .gitignore files.  Thanks also for the recommendations regarding that.  If I need to have you make a PR for this - I'll let you know shortly.

PR #3118, opened 2022-09-28
update .gitignore & committing .pot files

ryestew:
Updated .gitignore - but I did include doctrees - so I may be committing too much to the repo. 
see dragfish's comments: https://github.com/ethereum/remix-ide/pull/3117

The updated .pot files had only been saved locally - but with the update to .gitignore, they are being committed.


drafish:
Rob, you need to use this command `sphinx-build -b gettext . _build/gettext` to update the pot files. Now, it' just the old files parsed by `recommonmark`.  @ryestew

ryestew:
updated - but do not merge yet - there is an issue with images

drafish:
If you mean the issue that image string like this `` should not appear in pot file. I'v already take a lot of time trying to fix it, even debug the source code of `sphinx` and `myst` and `recommonmark`. But, still no clue. 

The ideal way is to exclude images and only extract alt if it have one. But it's just the case that the image is a block element. If it is inline, then it should be included in the source string.

So my thought is to find a configuration that can exclude block image. Sadly, I didn't find one. I only found this `myst_disable_syntax = ['image']`, which can disable alt extraction. Because I can't exclude block image, the alt string will be duplicated with the image string, so I add this configuration to `conf.py`.

@ryestew If you find a way to exclude block image, remember to remove `myst_disable_syntax = ['image']`.

I will also take some time to see if I can find a perfect solution.

drafish:
I found the code that cause this image issue. I'v already submited an issue in `myst` repository, and also submited a pr to fix it. 

But I'm start to think that maybe it's not an issue.  Most of the images are screenshot of remix-ide, which is also translable. If we want to replace these images into other language version, I think we should keep it there. @ryestew

drafish:
image
If anyone wondering why there is this metion. It's because I misunderstood about what @ryestew said `there is an issue with images`.  I submit a issue in `MyST` repository, and metioned.

Then I found out that images didn't render correctly. I realize that this should be the issue. I already submit a PR to fix it. 

And I tried to remove this mistaken metion.  But it seems like unremovable.

yann300 (approved):

PR #3121, opened 2022-10-10
Fix internal link and image render

drafish:
The images not render correctly.
image

I added `myst_disable_syntax = ['image']` into conf.py in pr #3117  , because I wanted to remove the alt string which was duplicated with the image string. But it also caused the image render issue. I think this is the image issue @ryestew metioned in pr #3118 .

Now I removed `myst_disable_syntax = ['image']`, and tested it again, it rendered correctly.

There is another issue I found when I trying to fix the image issue, which also caused by pr #3117 . 

link text
image
will become like this
image
 
`myst` will ignore all the internal links by default. Add this `myst_all_links_external = True` into conf.py will fix it.

Sorry for my careless work, I should test it more carefully.


ryestew (approved):

ryestew:
Thanks @dragfish!

PR #3157, opened 2023-07-25
specify sphinx ver

ryestew:


drafish:
@ryestew I don't think this kind of commits requires a PR. I suppose you can commit on master directly if you have the authority.

PR #3193, opened 2024-04-17
update to setting and community

ryestew:


vermouth22 (changes requested):
On new Line 4, first sentence should end in a question mark: "Do you have questions about Remix?"

on new line 10, words repeat. It should read: "If you'd like to develop a plugin for Remix, check this guide..."

Also, in new line 33, under "Temperature" Are you sure this is accurate?:
"... As the setting is increased, the range of possible answers will INCREASE. A higher number is more likely to return a relevant answer."

Or, should it be: 
"... As the setting is increased, the range of possible answers will DECREASE. A higher number is THEREFORE more likely to return a relevant answer."

It would seem decreasing the range of possible answers would give a more relevant (tighter) range of relevant answers?

ryestew:
@vermouth22  For AI settings - this is what Stephane had told me.  He had read through the AI docs which this is taken from.

vermouth22 (approved):

PR #3198, opened 2024-05-08
Remix RTD Docs design revamp and unification

wackerow:
## Description
Brings new menu navigation and color themes in tandem with unifying [updates to the Remix Project homepage](https://github.com/ethereum/remix-website/pull/71).

image

## Preview URL
- https://remix-ide-dev.readthedocs.io/


pettinarip (commented):
@wackerow minor UI issue

The base font seems to be Helvetica, but in this case, its using Arial
image

pettinarip (commented):
1. Small problem with the sidebar. Long titles cause horizontal scrolling, which I'm not sure we want. This example is in the Spanish version
image

This scrolling is also going to break a bit the mobile experience.

2. Probably not something related to this revamp but just noted that we have this copy where we mention that we have more translations than the ones we have on the menu
image

3. Nice to have: capital "E" on "espanol"
image

Last observation, the docs are translated into 3 languages while the homepage is in 2. Spanish is the difference. Probably not a problem, just wanted to mention it in case we need consistency there.

wackerow:
Thanks @pettinarip! Few responses and follow-up questions:

> The base font seems to be Helvetica, but in this case, its using Arial

I'm not getting this... are you still seeing this? If so, let me know what browser you're using.

> 1. Small problem with the sidebar. Long titles cause horizontal scrolling, which I'm not sure we want. This example is in the Spanish version

Thanks, I'll look into and try to get a patch up.

> 2. Probably not something related to this revamp but just noted that we have this copy where we mention that we have more translations than the ones we have on the menu

Yeah this would be out of scope here... would suggest the team edits that once the upgrade is complete with the latest languages.

> 3. Nice to have: capital "E" on "espanol"

I'm grabbing the language codes from the "flyover" menu that generates automatically by RTD (the names don't populate, just the codes).. from there I'm feeding it to `Intl.DisplayNames` to generate that display name... it's browser specific, so some will produce `español`, others will show `Español`.... the only way to avoid this would be to also store a mapping of lang codes to their display names, which I avoided to make maintenance easier.

wackerow:
Just pushed a patch for the text-overflow... @pettinarip let me know if you're still getting the Arial bug, otherwise I think this is good for another review.

pettinarip:
> Just pushed a patch for the text-overflow... @pettinarip let me know if you're still getting the Arial bug, otherwise I think this is good for another review.

I see this just in Brave (Version 1.66.110 Chromium: 125.0.6422.60 (Official Build) (64-bit)). Chrome and FF are working fine. 

Minor issue I'd say. Anyway, for some reason, the styles for this button are overridden by the browser's styles.

image

pettinarip:
> Thanks @pettinarip! Few responses and follow-up questions:
> 
> > The base font seems to be Helvetica, but in this case, its using Arial
> 
> I'm not getting this... are you still seeing this? If so, let me know what browser you're using.
> 
> > 1. Small problem with the sidebar. Long titles cause horizontal scrolling, which I'm not sure we want. This example is in the Spanish version
> 
> Thanks, I'll look into and try to get a patch up.
> 
> > 2. Probably not something related to this revamp but just noted that we have this copy where we mention that we have more translations than the ones we have on the menu
> 
> Yeah this would be out of scope here... would suggest the team edits that once the upgrade is complete with the latest languages.
> 
> > 3. Nice to have: capital "E" on "espanol"
> 
> I'm grabbing the language codes from the "flyover" menu that generates automatically by RTD (the names don't populate, just the codes).. from there I'm feeding it to `Intl.DisplayNames` to generate that display name... it's browser specific, so some will produce `español`, others will show `Español`.... the only way to avoid this would be to also store a mapping of lang codes to their display names, which I avoided to make maintenance easier.

Got it. All working fine from my side now.

pettinarip (approved):

ryestew (approved):

PR #3202, opened 2024-05-21
Edit of Update run.md

vermouth22:
Edits to run.md

### DIFF
--- a/docs/run.md
+++ b/docs/run.md
@@ -153,22 +153,22 @@ The **Value** field is **NOT** for gas.
 
  ![](images/a-Runtab-deploy-atAddress.png)
 
-Deploy & AtAddress
+Deploy and AtAddress
 ------------------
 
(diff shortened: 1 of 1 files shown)


ryestew (changes requested):
AtAddress is not AtAddress.  The button is called AtAddress - nowhere else is it At Address.  We could consider renaming the button - but not to At Address.  It should be a single word.

vermouth22:
> AtAddress is not AtAddress. The button is called AtAddress - nowhere else is it At Address. We could consider renaming the button - but not to At Address. It should be a single word. But I think we should leave the naming discussion for another time (delay it) and bring it up out side of this PR.

Agreed. It should be "AtAddress" everywhere... including on the button

vermouth22 (commented):
Now edited

ryestew (approved):

PR #3214, opened 2024-11-20
Fix: typo and add link for reference

HiroyukiNaito:
Typo: repeating "will also will also"
No reference link: More about the Editor updates in this article: Major Updates to Remix Editor (the last line of this page)


123123nope (approved):

ryestew:
Thanks for noticing - but we need to find a better URL than the one you updated.  I'll take it from here.  Thanks so much for submitting!

PR #3215, opened 2024-11-30
Update recieve() to receive() name function

intls:


ryestew:
I'm sorry only the md files need to be updated.

PR #3225, opened 2025-05-16
Update 4337 text

ryestew:


vermouth22 (changes requested):
It should probably read:

The Simple EIP-7702 template includes a README file that may be helpful to review as an introduction to this EIP's features.

The Account Abstraction template has many examples and resources for both EIP-4337 and EIP-7702, including a very informative README file.

vermouth22 (approved):

PR #3231, opened 2025-07-26
Updated custom.css to accomodate switch-menus

Nishant891:
Related Issue - #6076 (In remix-project repository)

Version and Language Dropdowns in Sidebar Are Nearly Invisible Due to Low Contrast

The _switch-menu_ element is dynamically generated but, the _wy-side-nav-search_ is supposed to be flex row which causes both the elements to be rendered side-by-side and causes space distortion.

**FIX** :  Changed the _wy-side-nav-search_ flex-direction to column and alignment to stretch.

Should fix this issue.

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -355,12 +355,11 @@ nav.wy-nav-shift {
 /* Remove background from the search container */
 .wy-side-nav-search {
   display: flex;
-  flex-direction: row-reverse;
-  align-items: end;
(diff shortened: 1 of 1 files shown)


Nishant891:
Hey is this project deserted?

ryestew:
Thanks so much for contributing.  I am sorry this has taken so long to get to.  We needed to address this internally - so we on this problem we couldn't use your work.  Again - thanks so much for contributing your time and effort.

PR #3235, opened 2025-08-28
New workspaces page

davidekete:
This PR improves the file explorer page by updating the images and improving text and structure for readability. It also separates the workspaces and Git guide from the page.


davidekete:
Hi @ryestew, when you try to run this locally, you will get two warnings about files not being in the toctree. That's intentional and will be fixed in my next two PRs (one for each).

ryestew (changes requested):

PR #3238, opened 2025-09-02
tooling: add vale for spell checking, create custom wordlist, configu…

davidekete:
…re vale, add new make commands


.custom_wordlist.txt
```diff
@@ -0,0 +1,101 @@
...
+Pectra
+preloaded
+Pragma
+Remappings
+remappings
+Remix's
+remixd
+Remixd
+Rinkeby
+Ropsten
```
ryestew, line 73: Please remove

.custom_wordlist.txt
```diff
@@ -0,0 +1,101 @@
...
+param
+Pectra
+preloaded
+Pragma
+Remappings
+remappings
+Remix's
+remixd
+Remixd
+Rinkeby
```
ryestew, line 72: Please remove - we need to make sure that the Old Testnets are removed from Remix.

davidekete: Done

PR #3292, opened 2026-01-15
remix 1.4

davidekete:


docs/settings.md
```diff
@@ -41,11 +41,16 @@ When performing Git operations on Github and when creating GISTs, it may be nece
 
 ### Etherscan Access Token
 
-When debugging verified contracts with the Remix Debugger, it is necessary to input your Etherscan access token in the Settings panel. Whereas, when verifying a contract with the Contract Verification - Etherscan plugin, the API key is input there and not in the Settings panel.
+When debugging verified contracts with the Remix Debugger, it is necessary to input your Etherscan access token in the Settings panel. Whereas, when verifying a contract with the Contract Verification plugin (on Etherscan), you have to add it in the Settings section of the Contract Verification plugin.
```
ryestew: I think we need to rewrite this paragraph.  
It should say something more like:
**The Etherscan Access token is used in three operations in Remix:**
1. When using the Deploy & Verify button, the token can be saved either in the Settings panel or in the Contract Verification plugin's Settings tab.
2. When using the Contract Verification plugin, the Etherscan Access token must be placed in the Contract Verification plugin's Settings tab.
3. When debugging verified contracts with the Remix Debugger,  the token must be input in the Settings panel.

See the Contract Verification plugin Docs for more info about that plugin.

ryestew: @davidekete - the paragraph doesn't look updated.

PR #3310, opened 2026-03-29
update order

davidekete:


docs/run.md
```diff
@@ -81,6 +81,19 @@ The ABI is a JSON array that describes a contract's functions, inputs, and outpu
 To get the ABI of a contract you've compiled, open the Solidity Compiler plugin, compile the contract, then click **Compilation Details**. The ABI is listed in the modal that appears.
 ```
 
+## Forking chains in Remix
+
+Forking a chain will bring that chain to the Remix VM. Once it is forked, you'll have access to the 10 accounts loaded with 100 ETH.
+
+You can also fork the Remix VM in its current state. Click the version control icon to the right of the Environment title to create a fork and then to name it. Switching between forks is done in the Environment dropdown menu. You can also reset the state of the VM by clicking the refresh icon to the right of the version control icon. Forking the VM lets you snapshot a working state, try changes, and return to the original if needed.
```
ryestew: There is forking the current Remix VM - which is described in this paragraph.  But we also need to add how to fork another chain - which is done through the environment select box.  So images of that selection are needed and then the custom chain forking modal are needed.

ryestew (changes requested):

ryestew (approved):

PR #3354, opened 2026-07-26
docs: add subgraph docs

davidekete:


ryestew (commented):
Subgraph should always be captialized

ryestew (approved):

